PR #218, opened 2025-04-08
chore: add codecov project check

jpmunz:
For future reference various other things can be configured in this yaml as documented in https://docs.codecov.com/docs/commit-status


jpmunz:
I think this will only take affect once it is merged to main

gregorio-gerardi (approved):

PR #222, opened 2025-04-09
EMBR-7161 refactor initSDK to match README instructions

jpmunz:
Still need to write tests but wanted to put up the initial pass for feedback:
* Moved from "@opentelemetry/auto-instrumentations-web" to setting up the otel instrumentations we want directly
* Introduced `getDefaultInstrumentations` to allow configuring our instrumentations
* Changed from accepting the 'embrace' / 'otlp' export flags to accepting optional `spanExporters` and `logExporters` directly. We assume sending to Embrace if there's a valid app ID


src/sdk/defaultInstrumentations.ts
```diff
@@ -0,0 +1,136 @@
...
+    'enabled'
+  >;
+}
+
+export const getDefaultInstrumentations = (
+  config: DefaultInstrumenationConfig = { omit: new Set() }
+): Instrumentation[] => {
+  const instrumentations = [];
+
+  if (!config.omit.has('session-on-load')) {
```
jpmunz, line 75: not sure if there's a more automated way to do this that still respects the typing

src/sdk/initSDK.ts
```diff
@@ -47,171 +37,232 @@ import {
...
+import type { DiagLogger } from '@opentelemetry/api/build/src/diag/types';
+import { getDefaultInstrumentations } from './defaultInstrumentations.js';
+import { createSessionSpanProcessor } from '@opentelemetry/web-common';
 
 interface SDKInitConfig {
   /**
-   * appID is a unique identifier for your application. It is used to identify your application in Embrace, and it is only required when
-   * the Embrace exporter is enabled. You can find your appID in the Embrace dashboard. If embrace exporter is disabled this value will be ignored.
+   * appID is a unique identifier for your application. It is used to identify your application in Embrace, and can be
+   * found in the Embrace dashboard. This can only be omitted if at least one spanExporter or logExporter is setup to
```
gregorio-gerardi, line 48: related to "This can only be omitted if at least one spanExporter or logExporter is setup", this is a good use case for typescript union types, to validate we either a) have at least one exporter or b) have appID = true. Just a nice to have, nothing wrong with the runtime if else validation, but ts is always a nice feedback that validates in the IDE and at build time before running nothing

src/sdk/defaultInstrumentations.ts
```diff
@@ -0,0 +1,136 @@
...
+    instrumentations.push(
+      new SpanSessionOnLoadInstrumentation(config['session-on-load'])
+    );
+  }
+
+  if (!config.omit.has('session-visibility')) {
```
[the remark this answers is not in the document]
jpmunz: ya it's a good point, I'm rethinking this interface altogether as I hit some other issues during testing where if you call `getDefaultInstrumentations` yourself the session managers aren't setup correctly because the instrumentation constructors run too early (`registerInstrumentations` handles setting up tracer/logger providers, but not our custom Embrace managers)

maybe it's best for this just to take the config object and ya as you say some of the instrumentations we just don't allow to be turned off

src/sdk/initSDK.ts
```diff
@@ -47,171 +37,232 @@ import {
 } from '../processors/index.js';
 import { getWebSDKResource } from '../resources/index.js';
 import { isValidAppID } from './utils.js';
-
-type Exporter = 'otlp' | 'embrace';
+import type { SpanExporter } from '@opentelemetry/sdk-trace-base';
+import type { DiagLogger } from '@opentelemetry/api/build/src/diag/types';
```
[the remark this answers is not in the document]
jpmunz: yep not sure why the auto-complete ended up with the other import, fixed locally

src/sdk/initSDK.ts
jpmunz: oh nice, I'll take a look

jpmunz:
added more of an integration test around exporting to Embrace: https://github.com/embrace-io/embrace-web-sdk/pull/222/commits/3450a99345717b6a951d7da7d36db077c82bc0b1

src/sdk/setupDefaultInstrumentations.ts
```diff
@@ -0,0 +1,66 @@
...
+    instrumentations.push(
+      new GlobalExceptionInstrumentation(config['exception'])
+    );
+  }
+
+  if (!config.omit?.has('click')) {
+    instrumentations.push(new ClicksInstrumentation(config['click']));
+  }
+
+  if (!config.omit?.has('web-vital')) {
```
gregorio-gerardi, line 39: just double checking: if we enable web vitals but not document load, nothing will break in the UI with the new grouping you added, right? they can also be shown as standalone if they are not both found?

gregorio-gerardi (approved):

src/sdk/types.ts
```diff
@@ -0,0 +1,209 @@
...
+};
+
+type SpanExportSDKInitConfig = {
+  /**
+   * spanExporters can be set to export span to a collector other than Embrace. If `appID` is omitted at lease one
+   * exporter needs to be set here, or in `logExporters`.
+   *
+   * **default**: []
+   */
+  spanExporters: SpanExporter[];
```
gregorio-gerardi, line 136: this should be
`spanExporters: [SpanExporter,...SpanExporter[]];` 

that will tell typescript that an empty array is invalid and you need at least 1 element

the same for the Log export config

src/sdk/initSDK.test.ts
```diff
@@ -0,0 +1,261 @@
...
+  afterEach(() => {
+    spanExporter.reset();
+    logExporter.reset();
+    trace.disable();
+    logs.disable();
+  });
+
+  it('should require an app ID when not setting custom exporters', () => {
+    const diagLogger = new InMemoryDiagLogger();
+    // @ts-expect-error need to bypass type checking to test this invalid configuration
```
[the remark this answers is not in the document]
jpmunz: someone not using typescript could still use our sdk and then not have the benefit of the build time check correct?

src/sdk/setupDefaultInstrumentations.ts
jpmunz: yep exactly, it's an optional grouping and they will just be shown individually otherwise

src/sdk/types.ts
jpmunz: ah cool, will update

src/sdk/initSDK.test.ts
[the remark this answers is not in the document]
gregorio-gerardi: oh, that is true. 🤔 ... We should probably decide on supporting either all typescript validations or fully checking everything assuming javascript. Every single parameter we support that is not marked as "optional" will throw a runtime error if the typescript compiler didn't validate it for a user, the same if you pass a parameter of the wrong type (like an instrumentation without a `enable` method). I believe what most libraries do is just use typescript, and then document the expected and supported use case in the readme or docs too. If then you choose to, for example, send an array of `exporters` that are not of type exporters, but something like strings, well... we won't be validating every single interface at runtime, we would be rewriting the typescript compiler at that point 😅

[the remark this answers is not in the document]
jpmunz: ya I agree that would be overkill, I think for this particular case the runtime check for appID or custom exporter doesn't hurt to leave in even if a bit redundant, the check itself I think is clear in the code and serves as a bit of extra documentation of what we're expecting

gregorio-gerardi (approved):

pablomatiasgomez (approved):

PR #227, opened 2025-04-10
EMBR-7259 - Create high level API that allows calling methods without getting the manager

pablomatiasgomez:
This should allow us to do things like:

```
import { session } from '@embrace-io/web-sdk';

session.addBreadcrumb("something happened");
```

### DIFF
--- a/src/api-sessions/api/SessionAPI/SessionAPI.ts
+++ b/src/api-sessions/api/SessionAPI/SessionAPI.ts
@@ -3,8 +3,10 @@ import {
   type SpanSessionManager,
 } from '../../manager/index.js';
 import type { SessionAPIArgs } from './types.js';
+import type { ReasonSessionEnded } from '../../manager/types.js';
+import type { HrTime } from '@opentelemetry/api';
 
-export class SessionAPI {
+export class SessionAPI implements SpanSessionManager {
   private static _instance?: SessionAPI;
   private readonly _proxySpanSessionManager;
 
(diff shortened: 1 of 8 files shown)


src/api-sessions/api/SessionAPI/SessionAPI.ts
```diff
@@ -29,4 +31,32 @@ export class SessionAPI {
   public setGlobalSessionManager(sessionManager: SpanSessionManager): void {
     this._proxySpanSessionManager.setDelegate(sessionManager);
   }
+
+  public endSessionSpan() {
+    this.getSpanSessionManager().endSessionSpan();
+  }
+
+  public endSessionSpanInternal(reason: ReasonSessionEnded) {
```
jpmunz, line 39: maybe not this one since we're flagging it as Internal, @gregorio-gerardi was the intention to hide this part of the interface somehow?

pablomatiasgomez: ah, right. I implemented it as I wanted to make use of the interface, and I see it has a TODO there to be moved to somewhere else.. https://github.com/embrace-io/embrace-web-sdk/blob/ed895e82c1c6c94ca72a5777407a41dbe0ada4ce/src/api-sessions/manager/types.ts#L19 maybe to be treated in a separate ticket? 


I really wanted to implement the interface so that we don't miss other methods in the future, but maybe I can avoid doing that and remove this method.

jpmunz (approved):

PR #229, opened 2025-04-10
EMBR-6970 support configuring SDK log levels

jpmunz:
Making sure we're using DiagLogger over directly sending to console and then exposing a way in `initSDK` to control logging. Prior to this we were using a no-op logger and turning this on revealed a couple existing errors that were being logged by the SDK


src/sdk/initSDK.ts
```diff
@@ -126,6 +131,17 @@ export const initSDK = (
   }
 };
 
+const sdkLogLevelToDiagLogLevel = (logLevel: SDKLogLevel): DiagLogLevel => {
```
jpmunz, line 134: decided to add this custom type to avoid users of the SDK needing to depend on the OTEL API for `DiagLogLevel`, also we're exposing less levels (found that below 'info' is not helpful since there's a ton of debug logging coming from some of the instrumentations)

src/sdk/initSDK.ts
```diff
@@ -179,7 +195,6 @@ const setupTraces = ({
     contextManager,
     propagator,
   });
-  trace.setGlobalTracerProvider(tracerProvider);
```
jpmunz, line 182: this was an existing error we were throwing, `register` above already does this so it was complaining that we were setting a duplicate global provider

src/sdk/initSDK.ts
pablomatiasgomez: should we also map Debug level?

pablomatiasgomez (approved):

src/sdk/initSDK.ts
jpmunz: that was part of my comment above I was seeing that 'debug' is really verbose for some of the otel instrumentations so I was hesitant to have it as an option though maybe it does make sense for completeness

pablomatiasgomez: I agree on the first part of the comment but I still believe we should allow configuring all of them.. I mean, it's up to he user at some point, I don't see any harm by allowing that, do you?

src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.test.ts
```diff
@@ -495,4 +495,55 @@ describe('WebVitalsInstrumentation', () => {
...
+      rating: 'good',
+      delta: 0,
+      id: 'm1',
+      entries: [],
+      navigationType: 'navigate',
+      attribution: {},
+    } as MetricWithAttribution;
+
+    // @ts-expect-error invalid attribution for testing
+    metricValue.attribution['largestShiftTarget'] = metricValue;
```
gregorio-gerardi, line 522: same comment as in the other PR. How would we end up in this scenario? is there a place in the logic when we iterate over attributions and cast them without validating them first? o how would someone end up bypassing typescript to make this test necessary?

(It doesn't hurt, it is just that disabling ts for a test is a weird thing to see in a project)

src/sdk/initSDK.ts
gregorio-gerardi: > avoid users of the SDK needing to depend on the OTEL API for DiagLogLevel

you can achieve the same by just re exporting the enum, if you end up supporting all the values OTel exposes like Pablo suggested. Any time we want to just avoid a user depending on otel we can reexport any interface or helper method they may need from our package

gregorio-gerardi (approved):

src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.test.ts
jpmunz: ya I think the error handler we have is probably fairly unlikely to ever be hit, I couldn't see a way to trigger it with the current available attribution keys though we iterate over them generically so it could be possible that some new one is added that could actually trigger the error

gregorio-gerardi (approved):

README.md
```diff
@@ -310,6 +310,22 @@ TODO
 
 ## Troubleshooting
 
+### Turning on verbose logging in the SDK
+
+The log level of the SDK can be increased when initializing as follows:
```
pablomatiasgomez, line 315: Maybe add a comment stating that the default is error?

pablomatiasgomez (approved):

README.md
jpmunz: 👍 will include on my next PR w/ the the updated documentation

PR #232, opened 2025-04-10
chore: remove unused code

jpmunz:
Mainly wanted to remove the dependency we have on 'shimmer' since we weren't using it, however after removing some unused blocks it seems like we could have `EmbraceInstrumentationBase` inherit directly from `InstrumentationAbstract` and not need the `InstrumentationBase` as an intermediate


src/instrumentations/InstrumentationAbstract/InstrumentationAbstract.ts
```diff
@@ -76,16 +55,13 @@ export abstract class InstrumentationAbstract<
     this._updateMetricInstruments();
   }
 
-  // TBH, I agree with typescript here, but keeping it disabled for consistency with the base repo
```
jpmunz, line 79: saw this comment but there didn't seem to be a typescript rule ignored here

src/instrumentations/InstrumentationAbstract/InstrumentationAbstract.ts
```diff
@@ -76,16 +55,13 @@ export abstract class InstrumentationAbstract<
...
   private _tracer: Tracer;
 
   /* Returns tracer */
   protected get tracer(): Tracer {
     return this._tracer;
   }
 
-  // NOTE: disabling typescript check, as this class was copied from OTel repo.
```
jpmunz, line 87: saw this comment but there didn't seem to be a typescript rule ignored here

pablomatiasgomez (approved):

src/instrumentations/InstrumentationAbstract/InstrumentationAbstract.ts
gregorio-gerardi: oh, my bad, I ended up reverting the rule because it was just too many places where it was complaining

gregorio-gerardi (approved):
While I like this change, my only concern is: how can we be sure none of the removed mehotd was being used by OTel internally while interacting with our instrumentation? specially the ones marked as "public".

gregorio-gerardi:
> > While I like this change, my only concern is: how can we be sure none of the removed mehotd was being used by OTel internally while interacting with our instrumentation? specially the ones marked as "public".
> 
> ya it's a good point, `init` in particular is hard to search for on its own to see it's usage so I restored that. `getModuleDefinitions` and `_runSpanCustomizationHook` don't seem to be used searching through otel-js and js-contrib repos however it doesn't really hurt to keep them in case they are used by some helper in the future that we want to make use of
> 
> I kept the removal of massWrap/Unwrap to avoid our dependency on shimmer, it seems these are only used as conveniences in a couple instrumentations. I also kept the removal of `InstrumentationBase` as it seems we didn't really need that, let me know what you think

Yes, all good now. The massWrap and Unwrap are protected methods anyway, so it is unlikely (impossible?) for anyone other than our own extension to use them

PR #245, opened 2025-04-11
Fix doc to reflect startPerformanceSpan method

pablomatiasgomez:
## Goal

Doc is wrong, method should be `startPerformanceSpan`

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ For the most basic usage simply start a span and end it after some operation com
 ```typescript
 import { trace } from '@embrace-io/web-sdk';
 
-const span = trace.startSpan("span-name");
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -175,7 +175,7 @@ For the most basic usage simply start a span and end it after some operation com
 ```typescript
 import { trace } from '@embrace-io/web-sdk';
 
-const span = trace.startSpan("span-name");
+const span = trace.startPerformanceSpan("span-name");
 
 someAsyncOperation().then(() => span.end());
```
jpmunz, line 180: should also do a null check here actually since `startPerformanceSpan` could return null

jpmunz (approved):

PR #244, opened 2025-04-11
fix: typescript build broken

gregorio-gerardi:
## Goal

Simplify imports from the OpenTelemetry API by importing both `Span` and `SpanOptions` types directly from the main package instead of importing `SpanOptions` from a build path.

## Testing

This change only affects type imports and doesn't modify any runtime behavior, so existing tests should continue to pass without modification.

PR #246, opened 2025-04-11
style: prevent nested imports using the linter

gregorio-gerardi:
(AI generated)## Goal

Add ESLint rules to enforce better import patterns across the codebase. This PR introduces three new rules:

1. Prevent third-party imports from referencing internal packages
(this will prevent issues like https://github.com/embrace-io/embrace-web-sdk/pull/244 which I also faced twice already)
2. Limit relative imports to reference only one folder deep
(this will force us to remind our convention of reexporting from the intermediate folder when the scope requires it)
3. Ensure relative imports reference index.js files
(also a convention we were following already, this enforces it)

Additionally, this PR reorganizes imports and exports to follow these new patterns by:
- Exporting types from index.js files instead of directly from type files
- Standardizing import paths to use index.js references
- Reordering class methods alphabetically in several files
- Updating imports to use constants from index files rather than direct attribute files

## Testing

The code has been updated to pass the new ESLint rules while maintaining the same functionality. All existing tests continue to pass with the updated import patterns.

## NOTE 
You only need to review the file `eslint.config.mjs`. Everything else is just applying those rules (but you should also spot check a couple changes to validate my rules are working 😅 )

Also, some methods for some classes got orderer alphabetically by my IDE, but there are no functional changes


gregorio-gerardi:
* **#246** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-web-sdk/246?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-web-sdk/246?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

eslint.config.mjs
```diff
@@ -102,6 +102,37 @@ export default tseslint.config({
...
+        // make sure all relative imports are reference the index.js file
+        // e.g. import type { something } from './folder/other.js'; is invalid,
+        // e.g. import type { something } from './folder/index.js'; is valid,
+        // import type { something } from './someOther.js'; is invalid
+        // import type { something } from './index.js'; is valid
+        // import { something } from './something.js'; is valid because it references the same folder as the imported file
+        // import { something } from '../something.js'; is invalid
+        {
+          regex: "import .* from '\\.{2}.*(?<!\\/index)\\.js'",
+          message: `Relative imports should always reference the index.ts files`,
```
jpmunz, line 134:
```suggestion
          message: `Relative imports should always reference the index.js files`,
```

eslint.config.mjs
```diff
@@ -102,6 +102,37 @@ export default tseslint.config({
           regex: `CLI_VERSION = '(?!${sdkPackageInfo.version}).*'`,
           message: `CLI_VERSION version mismatch. It should always match the one listed in both package.json (${sdkPackageInfo.version}) and cli/package.json (${cliPackageInfo.version}) and those 2 should be in sync.`,
         },
+        // make sure all third party imports are not referencing internal packages
+        // e.g. import type { something } from '@openteleme/something'; is valid, but
+        // import type { something } from '@openteleme/something/internal'; is not
+        // import type { something } from 'openteleme/something/internal'; is not
+        {
+          regex: "import .* from '@?(\\w)+\\/(\\w)+\\/.+'",
```
jpmunz, line 110: is this regex preventing internal packages specifically, or is it just enforcing imports from third-parties aren't more than 2 levels deep? E.g. would it flag something like `from '@opentelemetry/semantic-conventions/incubating';` as invalid?

src/exporters/EmbraceLogExporter/OTLPFetchLogExporter.ts
```diff
@@ -4,8 +4,8 @@ import type {
   ReadableLogRecord,
 } from '@opentelemetry/sdk-logs';
 import { BaseFetchExporter } from '../BaseFetchExporter/index.js';
-import { createOtlpBrowserFetchExportDelegate } from '../otlpBrowserFetchExportDelegate.js';
-import type { OtlpFetchExporterConfig } from '../types.js';
+import type { OtlpFetchExporterConfig } from '../index.js';
+import { createOtlpBrowserFetchExportDelegate } from '../index.js';
```
jpmunz, line 8: not sure if possible to enforce with the linter but for something like this would it be preferred to have it like this?

```suggestion
import { createOtlpBrowserFetchExportDelegate, type OtlpFetchExporterConfig } from '../index.js';
```

eslint.config.mjs
gregorio-gerardi: It will be reported as invalid 😞 I didn't see it because I forgot to include the `-` char for the `semantic-conventions` section. I will allow it if it ends in "incubating", that seems to be the only error reported by the linter after adding the `-`

src/exporters/EmbraceLogExporter/OTLPFetchLogExporter.ts
gregorio-gerardi: I like having typescript-specific lines as isolated statements. It makes it a lot easier for bundlers, node, and other tools to just ignore it or tree shake it when not dealing with types. Also, now that we are no longer importing from a "types" file, it is clear in its intent if the whole line is prefixes with "type".  Does it makes sense? 

Changing it is just a linter rule, yes, but I made it this way because I liked it more 😅

gregorio-gerardi: it is also less verbose when you have mora than one type imported, you don't need to repeat the `type`

jpmunz: np, yep let's go with this way then

jpmunz (approved):

gregorio-gerardi:
### Merge activity

* **Apr 11, 5:38 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-web-sdk/246).

PR #270, opened 2025-04-14
fix: set global trace manager

pablomatiasgomez:
## Goal

I missed to add that here https://github.com/embrace-io/embrace-web-sdk/pull/226
so currently the Noop is being set and traces are not being sent :(


gregorio-gerardi (approved):

jpmunz (approved):
lgtm, probably a good idea for us to test custom log + traces in `initSDK.test.ts` I can take a look at adding that as a follow-up I think it should be doable with the test setup that we have in there now

pablomatiasgomez:
> lgtm, probably a good idea for us to test custom log + traces in `initSDK.test.ts` I can take a look at adding that as a follow-up I think it should be doable with the test setup that we have in there now

added!

PR #271, opened 2025-04-14
chore: change logException signature to accept optional timestamp

pablomatiasgomez:
## Goal

Change logException to take an optional timestamp parameter. If not received, we use now()

### DIFF
--- a/src/api-logs/api/LogAPI/LogAPI.ts
+++ b/src/api-logs/api/LogAPI/LogAPI.ts
@@ -30,12 +30,12 @@ export class LogAPI implements LogManager {
   }
 
   public logException(
-    timestamp: number,
     error: Error,
     handled: boolean,
-    attributes?: Record<string, AttributeValue | undefined>
+    attributes?: Record<string, AttributeValue | undefined>,
+    timestamp?: number
   ) {
-    this.getLogManager().logException(timestamp, error, handled, attributes);
+    this.getLogManager().logException(error, handled, attributes, timestamp);
   }
 
   public message(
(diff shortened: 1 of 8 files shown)


src/api-logs/api/LogAPI/LogAPI.ts
```diff
@@ -30,12 +30,12 @@ export class LogAPI implements LogManager {
   }
 
   public logException(
-    timestamp: number,
     error: Error,
     handled: boolean,
-    attributes?: Record<string, AttributeValue | undefined>
+    attributes?: Record<string, AttributeValue | undefined>,
```
gregorio-gerardi, line 35: if you have more than one optional parameter, it is usually better to make the whole list an object so the order doesn't matter anymore. Otherwise, you will need to pass explicit `undefined` values for the optional parameters. 

It is also easier to add thing that way in the future without requiring a breaking change

jpmunz (approved):

src/managers/EmbraceLogManager/EmbraceLogManager.ts
```diff
@@ -42,18 +42,19 @@ export class EmbraceLogManager implements LogManager {
   }
 
   public logException(
-    timestamp: number,
     error: Error,
     handled: boolean,
-    attributes?: Record<string, AttributeValue | undefined>
+    attributes?: Record<string, AttributeValue | undefined>,
+    timestamp?: number
```
gregorio-gerardi, line 48: suggestion, just a bit more javascript-ish 😅 

`timestamp: number = this._perf.getNowMillis()`

(

gregorio-gerardi (approved):

PR #277, opened 2025-04-14
chore: protect against the SDK being initialized multiple times

jpmunz:
## Goal

I think a few different ways to handle this let me know what you think about this approach, with this we could potentially expose the `registry` to the user for grabbing the setup SDK elsewhere in their app

### DIFF
--- a/src/sdk/initSDK.test.ts
+++ b/src/sdk/initSDK.test.ts
@@ -33,6 +33,7 @@ import {
   EmbraceUserManager,
 } from '../managers/index.js';
 import { ProxyUserManager, user } from '../api-users/index.js';
+import { registry } from './registry.js';
 
 chai.use(sinonChai);
 const { expect } = chai;
@@ -52,6 +53,7 @@ describe('initSDK', () => {
     trace.disable();
     logs.disable();
     diag.disable();
+    registry.clear();
   });
 
   it('should require an app ID when not setting custom exporters', () => {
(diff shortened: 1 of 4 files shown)


src/sdk/initSDK.test.ts
```diff
@@ -359,4 +366,83 @@ describe('initSDK', () => {
...
+        logLevel: DiagLogLevel.WARN,
+        defaultInstrumentationConfig: {
+          omit: new Set(['web-vital']),
+          'web-vital': { listeners: testWebVitalListeners.listeners },
+        },
+      });
+      void expect(result1).not.to.be.false;
+      void expect(testWebVitalListeners.clsStub).not.to.have.been.called;
+
+      // 2nd invocation does not omit the web vital instrumentation, this should be ignored since only the first
```
gregorio-gerardi, line 396: not really following where you are making any assertion based on the web vitals disabled state. Both checks are for `void expect(testWebVitalListeners.clsStub).not.to.have.been.called`

src/sdk/registry.ts
```diff
@@ -0,0 +1,19 @@
...
+
+class Registry implements SDKRegistryManager {
+  private _sdk: SDKControl | null = null;
+
+  public register: (sdk: SDKControl) => void = sdk => {
+    this._sdk = sdk;
+  };
+
+  public clear: () => void = () => {
+    this._sdk = null;
```
gregorio-gerardi, line 11: would it be useful to first check if there is an SDK registered and print a log warning about "there is no sdk registered, this will be a noop" like we do in other similar managers? not a required change, I just thought it may be useful for debugging in the future. The same applies to "register" if you call it and there is something already registered

gregorio-gerardi (approved):

src/sdk/registry.ts
jpmunz: yep makes sense, updated

PR #286, opened 2025-04-15
chore: document how data is exported

jpmunz:
## Goal

Documenting how we export data from the SDK. In reviewing the code I think the only issue I noticed was that we aren't currently ever returning an error as 'retryable' so the retry logic isn't being used: https://github.com/embrace-io/embrace-web-sdk/blob/7276b54b2ca914fbd80bd863457ca06721eaeb1a/src/transport/FetchTransport/FetchTransport.ts#L87

Will file a follow-up for that specifically

### DIFF
--- a/DATA_EXPORT.md
+++ b/DATA_EXPORT.md
@@ -1,0 +1,29 @@
+# SDK Data Export
+
+## When is data sent?
+
+Telemetry data for Spans and Logs are exported from the SDK independently. Spans are sent as a batch whenever a session
+ends, this is managed by the [EmbraceSessionBatchedSpanProcessor](./src/processors/EmbraceSessionBatchedSpanProcessor/EmbraceSessionBatchedSpanProcessor.ts).
+Logs are also batched but can be sent throughout a session, this is managed by a [BatchLogRecordProcessor](https://github.com/open-telemetry/opentelemetry-js/blob/experimental/v0.57.0/experimental/packages/sdk-logs/src/platform/browser/export/BatchLogRecordProcessor.ts).
+
+## How is data sent?
+
+Data from the SDK is sent to Embrace using a CORS HTTP POST request. The data is gzip compressed and encoding as
+[OTLP JSON Protobuf](https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding). We make use of the
+[keepalive property](https://developer.mozilla.org/en-US/docs/Web/API/Request/keepalive) of the browser's Fetch API to
+help ensure the transmission of data even if the page has been closed. If possible we will attempt to retry request
+failures following an exponential backoff. See [FetchTransport](./src/transport/FetchTransport/FetchTransport.ts) and
+[RetryingTransport](src/transport/RetryingTransport/RetryingTransport.ts) for more details. 
+
+> [!NOTE]
+> The Fetch API's `keepalive` property is supported by all modern browsers, however it was a relatively recent addition
+> to Firefox and may not exist on some older versions of that browser in particular. Refer to this [Browser compatibility chart](https://developer.mozilla.org/en-US/docs/Web/API/Request/keepalive#browser_compatibility)
+> for more details.
+
+## What happens with custom export?
+
+If you [provide your own custom exporter](./README.md#custom-exporters) then batching of spans and logs are managed by a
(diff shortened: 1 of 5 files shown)


src/transport/RetryingTransport/constants.ts
```diff
@@ -2,4 +2,4 @@ export const MAX_ATTEMPTS = 5;
 export const INITIAL_BACKOFF = 1000;
 export const MAX_BACKOFF = 5000;
 export const BACKOFF_MULTIPLIER = 1.5;
-export const JITTER = 0.2;
+export const JITTER = 20;
```
jpmunz, line 5: this is in milliseconds so the `0.2` value seemed to small to have an effect

src/processors/EmbraceSessionBatchedSpanProcessor/EmbraceSessionBatchedSpanProcessor.ts
```diff
@@ -34,17 +34,27 @@ export class EmbraceSessionBatchedSpanProcessor extends EmbraceProcessor {
...
-      this.diag.debug('Non-session span ended. Adding to pending spans queue.');
+      this.diag.debug('non-session span ended. Adding to pending spans queue.');
       this._pendingSpans.push(span);
     } else {
-      // TODO: handle errors
-      this.diag.debug('Session span ended. Exporting all pending spans.');
-      void internal._export(this._exporter, [span, ...this._pendingSpans]);
+      this.diag.debug('session span ended. Exporting all pending spans.');
+      internal
+        ._export(this._exporter, [span, ...this._pendingSpans])
```
jpmunz, line 47: as far as I could tell this never returns a rejected promise: https://github.com/open-telemetry/opentelemetry-js/blob/cf6dffeebcf72c42b2cb4d2bf2db720369b53081/packages/opentelemetry-core/src/internal/exporter.ts#L37

added this for completeness though given that I wasn't sure how to test, alternatively I could just update the comment to reflect that this will always return a resolved promise?

DATA_EXPORT.md
```diff
@@ -0,0 +1,29 @@
...
+
+## When is data sent?
+
+Telemetry data for Spans and Logs are exported from the SDK independently. Spans are sent as a batch whenever a session
+ends, this is managed by the [EmbraceSessionBatchedSpanProcessor](./src/processors/EmbraceSessionBatchedSpanProcessor/EmbraceSessionBatchedSpanProcessor.ts).
+Logs are also batched but can be sent throughout a session, this is managed by a [BatchLogRecordProcessor](https://github.com/open-telemetry/opentelemetry-js/blob/experimental/v0.57.0/experimental/packages/sdk-logs/src/platform/browser/export/BatchLogRecordProcessor.ts).
+
+## How is data sent?
+
+Data from the SDK is sent to Embrace using a CORS HTTP POST request. The data is gzip compressed and encoding as
```
gregorio-gerardi, line 11: encoding -> encoded?

gregorio-gerardi, line 11: CORS HTTP POST request -> HTTP POST request ? Not sure what do you mean with  CORS here

src/processors/EmbraceSessionBatchedSpanProcessor/EmbraceSessionBatchedSpanProcessor.ts
gregorio-gerardi: 🤔 weird. I assume if ANYTHING throws an exception during the export, that would be caught here, but looking at the different classes involved everything seems to be catching its errors and return a `status: 'failure'` to the exporter, which then gets ignored 😅 . 

You probably want to add a `then` handler that checks for the status and do the same there as in you current catch block if it is an error.

The returned promise will include https://github.com/open-telemetry/opentelemetry-js/blob/cf6dffeebcf72c42b2cb4d2bf2db720369b53081/packages/opentelemetry-core/src/ExportResult.ts#L17-L19

gregorio-gerardi (approved):

src/processors/EmbraceSessionBatchedSpanProcessor/EmbraceSessionBatchedSpanProcessor.ts
jpmunz: ah good call, ok I updated to include the `then` handler, I did keep the `catch` handler just in case but added a comment to emphasize it probably wouldn't be hit

DATA_EXPORT.md
jpmunz: updated

PR #310, opened 2025-04-17
chore: additional API for indicating a span failed

jpmunz:
## Goal

Adding an API to simplify ending a span as a failure: `trace.performanceSpanFailed(span)`

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,9 @@ import { trace } from '@embrace-io/web-sdk';
 
 const span = trace.startPerformanceSpan("span-name");
 
-someAsyncOperation().then(() => span?.end());
+someAsyncOperation()
+  .then(() => span?.end())
+  .catch(() => trace.performanceSpanFailed(span));
 ```
 
 Attributes and events can also be added to the span either on start or later during its lifespan. Our API wraps that of
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ const span = trace.startPerformanceSpan("span-name");
 
 someAsyncOperation()
   .then(() => span?.end())
-  .catch(() => trace.performanceSpanFailed(span));
+  .catch(() => span?.fail());
 ```
 
 Attributes and events can also be added to the span either on start or later during its lifespan. Our API wraps that of
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,9 @@
 import { log } from './api-logs/index.js';
 import { session } from './api-sessions/index.js';
 import { trace } from './api-traces/index.js';
+import { type PerformanceSpan } from './api-traces/index.js';
 import { user } from './api-users/index.js';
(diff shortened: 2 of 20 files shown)


README.md
```diff
@@ -89,7 +89,9 @@ import { trace } from '@embrace-io/web-sdk';
 
 const span = trace.startPerformanceSpan("span-name");
 
-someAsyncOperation().then(() => span?.end());
+someAsyncOperation()
+  .then(() => span?.end())
+  .catch(() => trace.performanceSpanFailed(span));
```
gregorio-gerardi, line 94: isn't it weird that one ending api is on the span itself and the other one is on the global trace api? should we also expose a `trace.performanceSpanSucceded(span)` for consistency? (or alternately, extend the Span type and add a method to it to end it like `span?.fail()`? 

Are the only 2 states for a span success and failure? should we include others if Embrace supports them?

src/managers/EmbraceTraceManager/EmbraceTraceManager.ts
```diff
@@ -10,4 +17,21 @@ export class EmbraceTraceManager implements TraceManager {
...
   }
+
+  public performanceSpanFailed(
+    span: Span | null,
+    options: PerformanceSpanFailedOptions = {
+      code: 'failure',
+    }
+  ) {
+    if (!span) {
+      return;
```
gregorio-gerardi, line 28: should we log a debug entry here?

README.md
jpmunz: there is also user_abandon which can be set through the optional parameter on `performanceSpanFailed`

I originally was looking at offering something like `span?.fail()` I think that is a bit cleaner though I wasn't able to get it working, let's discuss in standup I may have missed something obvious there. If that approach doesn't work I didn't find it too weird that we have `span?.end()` vs `trace.performanceSpanFailed(span)` but let's discuss that as well

gregorio-gerardi (approved):
pending the optional span?.fail() interface. LGTM

src/index.ts
```diff
@@ -1,7 +1,9 @@
 import { log } from './api-logs/index.js';
 import { session } from './api-sessions/index.js';
 import { trace } from './api-traces/index.js';
+import { type PerformanceSpan } from './api-traces/index.js';
 import { user } from './api-users/index.js';
 import * as sdk from './sdk/index.js';
 
+export type { PerformanceSpan };
```
jpmunz, line 8: @embrace-io/opensource-web are we ok with this style going forward for exposing types that might be useful for consumers of the SDK? Another options would be to collect them all under a separate `types` export

gregorio-gerardi: they will end up in a separate types export (a .d.ts file) after tsc compile the code anyways, so this looks good to me

gregorio-gerardi (approved):
LGTM!

PR #317, opened 2025-04-22
feat: expose support for sampling reusing OTel existing sampling

gregorio-gerardi:
(AI generated)## Goal

Add support for sampling traces in the SDK. This allows users to control the volume of traces generated by:

1. Passing a custom `spanSampler` parameter when initializing the SDK for complete control over sampling strategy
2. Passing a `spanSamplingRatio` value between 0 and 1 to use the default sampling strategy targeting a specific percentage

## Testing

The implementation has been thoroughly tested with unit tests that verify:
- Default behavior generates all spans when no sampling is configured
- Custom samplers can block or allow spans as expected
- The `spanSamplingRatio` parameter correctly limits trace generation to the specified percentage
- Edge cases like invalid ratio values (too high or too low) are handled properly
- Sampling works correctly with both session spans and custom spans


gregorio-gerardi:
* **#317** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-web-sdk/317?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-web-sdk/317?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

src/sdk/initSDK.test.ts
```diff
@@ -318,6 +320,314 @@ describe('initSDK', () => {
       ]);
     });
   });
+  describe('sampling', () => {
```
jpmunz, line 323: I think ok for this PR but I wonder if we should start splitting this file up as we add more tests, not sure the best convention maybe into things like `initSDK.sampling.test.ts`?

src/sdk/initSDK.ts
```diff
@@ -194,7 +201,13 @@ const setupTraces = ({
     );
   }
 
+  let sampler = spanSampler;
+  if (typeof spanSamplingRatio === 'number') {
```
jpmunz, line 205: won't this always be the case?

src/testUtils/FakeSessionManager/FakeSessionManager.ts
```diff
@@ -0,0 +1,47 @@
+import type { HrTime, Span } from '@opentelemetry/api';
+import type {
+  ReasonSessionEnded,
+  SpanSessionManager,
+} from '../../api-sessions/index.js';
+
+export class FakeSessionManager implements SpanSessionManager {
```
jpmunz, line 7: maybe worth extending from `NoOpSpanSessionManager`, only saves a couple lines atm but would avoid having to keep this helper up-to-date as we add to the interface

src/processors/EmbraceSessionBatchedSpanProcessor/EmbraceSessionBatchedSpanProcessor.ts
```diff
@@ -44,6 +44,13 @@ export class EmbraceSessionBatchedSpanProcessor extends EmbraceProcessor {
     }
 
     if (!isSessionSpan(span)) {
+      const currentSession = this.sessionManager.getSessionSpan();
+      if (currentSession === null || !currentSession.isRecording()) {
```
jpmunz, line 48: based on my reading of https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#isrecording it seems like this isn't quite the same as the span being sampled and instead refers to whether it is currently in a readonly state, e.g.:

> After a Span is ended, it SHOULD become non-recording and IsRecording SHOULD always return false

there is a separate `SampledFlag` though not sure if it is accessible here:

> Users of the API should only access the IsRecording property when instrumenting code and never access SampledFlag unless used in context propagators.

README.md
```diff
@@ -274,6 +276,25 @@ sdk.initSDK({
 });
 ```
 
+## Sampling
+
+If you wish to sample the amount of traces generated by the sdk, you can do so by
+a) Passing a `spanSampler` parameter implementing your custom sampling when initializing the SDK. This allows
+for complete control on the sampling strategy.
+a) passing a `spanSamplingRatio` with a value between 0 and 1. This will use our default sampling strategy targeting
+a percentage of sampling equal to the value of `spanSamplingRatio`. This is a simple way to control the amount of traces
```
jpmunz, line 285: not sure how to phrase it exactly but I think worth noting that we treat the SessionSpan as special, when sampling they should probably either always include the session spans or if one is omitted then also omitting ALL spans that are part of that same session to avoid sending incomplete data?

README.md
```diff
@@ -274,6 +276,25 @@ sdk.initSDK({
 });
 ```
 
+## Sampling
+
+If you wish to sample the amount of traces generated by the sdk, you can do so by
+a) Passing a `spanSampler` parameter implementing your custom sampling when initializing the SDK. This allows
```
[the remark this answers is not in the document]
gregorio-gerardi: ah, yes. Good catch

src/testUtils/FakeSessionManager/FakeSessionManager.ts
gregorio-gerardi: Good idea!

src/sdk/initSDK.ts
gregorio-gerardi: It could also be undefined. It is optional

README.md
gregorio-gerardi: I added a Note: what do you think? Even if the user just ignores it, I added tests validating that spans without a session don't crash or make the sdk break. They will work just fine for exporters other than the embrace one

src/processors/EmbraceSessionBatchedSpanProcessor/EmbraceSessionBatchedSpanProcessor.ts
gregorio-gerardi: yeah, I saw that too. I ended up with isRecording as that seems to be the flag that existing sampling classes turn off. The examples I referenced are

https://github.com/open-telemetry/opentelemetry-js/blob/fb2fdd293fcd2a4479d0ee10f1dfef12283225d7/packages/opentelemetry-sdk-trace-base/src/sampler/TraceIdRatioBasedSampler.ts#L34-L35 

and 

https://github.com/open-telemetry/opentelemetry-js/blob/fb2fdd293fcd2a4479d0ee10f1dfef12283225d7/api/src/trace/SamplingResult.ts#L27-L28

gregorio-gerardi:
After discussion with Hanson, we don't want to expose custom sampling for now. We will instead work on an approach similar to mobile sdks, where we configure sampling server side and send it to the sdk through our config server

PR #335, opened 2025-04-30
EMBR-7455 chore: add configuration for creating background sessions

jpmunz:
## Goal

Adding a configuration parameter to `SpanSessionVisibilityInstrumentation` to have it create sessions when the document becomes hidden, defaulting to false for now

### DIFF
--- a/src/api-sessions/manager/types.ts
+++ b/src/api-sessions/manager/types.ts
@@ -25,4 +25,4 @@ export type ReasonSessionEnded =
   | 'timer' // max_time_reached limit
   | 'manual' // using the public api
   | 'max_size_reached'
-  | 'bkgnd_state'; // visibility change to hidden
+  | 'state_changed'; // visibility change
--- a/src/instrumentations/session/SpanSessionVisibilityInstrumentation/types.ts
+++ b/src/instrumentations/session/SpanSessionVisibilityInstrumentation/types.ts
@@ -1,9 +1,5 @@
 import type { EmbraceInstrumentationBaseArgs } from '../../EmbraceInstrumentationBase/index.js';
-
-// Useful for testing so that we can pass in a document-like object and change its visibilityState
-export interface VisibilityStateDocument {
-  visibilityState: DocumentVisibilityState;
-}
+import type { VisibilityStateDocument } from '../../../common/index.js';
 
 export type SpanSessionVisibilityInstrumentationArgs = Pick<
   EmbraceInstrumentationBaseArgs,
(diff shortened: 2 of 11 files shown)


src/api-sessions/manager/types.ts
```diff
@@ -25,4 +25,4 @@ export type ReasonSessionEnded =
   | 'timer' // max_time_reached limit
   | 'manual' // using the public api
   | 'max_size_reached'
-  | 'bkgnd_state'; // visibility change to hidden
+  | 'state_changed'; // visibility change
```
jpmunz, line 28: I believe this is interchangeable on the backend and is a little more generic to handle both visible -> hidden and hidden -> visible

src/instrumentations/session/SpanSessionVisibilityInstrumentation/types.ts
```diff
@@ -1,6 +1,14 @@
 import type { EmbraceInstrumentationBaseArgs } from '../../EmbraceInstrumentationBase/index.js';
 
+// Useful for testing so that we can pass in a document-like object and change its visibilityState
```
jpmunz, line 3: let me know if this seems like a reasonable way to make the visibility state more testable since the `document.visibilityState` property is readonly

I also needed this in a 2nd spot and ended up adding separately to `src/managers/EmbraceSpanSessionManager/types.ts` since they're both only used internally by these classes but maybe it makes sense to have some top-level `types/` folder to define it just once?

src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -106,15 +112,18 @@ export class EmbraceSpanSessionManager implements SpanSessionManager {
   public startSessionSpan() {
     //if there was a session in progress already, finish it first.
     if (this._sessionSpan) {
-      this.endSessionSpanInternal('manual');
+      this.endSessionSpanInternal('unknown');
```
jpmunz, line 115: I think this is slightly more correct since we call `startSessionSpan` from instrumentations so the previous one ending it is not necessarily because of a manual API call, generally if we're doing that we should end the previous session with a particular reason before calling this method

src/instrumentations/session/SpanSessionVisibilityInstrumentation/types.ts
joaquin-diaz: Yeah. It makes sense to me to keep this interface for now, in the future if we need to test other read only document props we could consolidate it into one interface that picks what we need from the document. 

Maybe we can have `src/types/common.ts` or something? `src/types.ts` feels weird since everything is inside a folder. And then having `src/types/types.ts` may be too generic?

joaquin-diaz (approved):

src/instrumentations/session/SpanSessionVisibilityInstrumentation/types.ts
jpmunz: yep I like `src/types/common.ts` I can add that

PR #347, opened 2025-05-06
breaking: consistently group optional parameters in exported API

jpmunz:
## Goal

Make the log API consistent with the other ones we expose where the optional parameters are all grouped together in a single `options` argument after the required ones. Labelling the commit as "feat" so that we bump the minor version since this is a breaking change for the interface (though not the major version since we're still on 0.x)

## Testing

Added a few more tests where different options are specified / omitted to confirm the defaults are being applied correctly

### DIFF
--- a/src/managers/EmbraceLogManager/EmbraceLogManager.ts
+++ b/src/managers/EmbraceLogManager/EmbraceLogManager.ts
@@ -48,7 +48,7 @@ export class EmbraceLogManager implements LogManager {
   public logException(
     error: Error,
     {
-      handled = false,
+      handled = true,
       attributes = {},
       timestamp = this._perf.getNowMillis(),
     }: LogExceptionOptions = {}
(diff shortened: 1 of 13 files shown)


src/managers/EmbraceLogManager/EmbraceLogManager.ts
```diff
@@ -43,9 +47,11 @@ export class EmbraceLogManager implements LogManager {
 
   public logException(
     error: Error,
-    handled: boolean,
-    attributes: Record<string, AttributeValue | undefined> = {},
-    timestamp: number = this._perf.getNowMillis()
+    {
+      handled = false,
```
pablomatiasgomez, line 51: I wonder if handled should be one of the required ones, to make people think what they are actually calling...
But, having said that, I was looking at android https://embrace.io/docs/android/integration/log-message-api/?android-language=kotlin#log-handled-exception and looks like it's not even able to tell that? 🤔

pablomatiasgomez (approved):

joaquin-diaz (approved):

PR #386, opened 2025-05-20
chore: [BREAKING CHANGE] rename startPerformanceSpan method to startSpan

joaquin-diaz:
## Goal

Rename startPerformanceSpan method to startSpan to avoid the name being related to an internal attribute  

## Testing

n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ For the most basic usage simply start a span and end it after some operation com
 ```typescript
 import { trace } from '@embrace-io/web-sdk';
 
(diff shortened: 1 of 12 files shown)


pablomatiasgomez (approved):

jpmunz:
@joaquin-diaz as part of this rename it probably makes sense to rename the type that is returned from `PerformanceSpan` to maybe `EmbraceSpan` (`Span` potentially works too though I'd worry about it being confusing that it has the same name as the base otel type

PR #387, opened 2025-05-21
chore: rename EmbracePerformanceSpan and PerformanceSpan to remove references to internal implementation

joaquin-diaz:
## Goal

chore: rename EmbracePerformanceSpan and PerformanceSpan to remove references to internal implementation

## Testing

n/a

### DIFF
--- a/src/managers/EmbraceTraceManager/EmbraceSpan.ts
+++ b/src/managers/EmbraceTraceManager/EmbraceSpan.ts
@@ -9,17 +9,17 @@ import type {
 } from '@opentelemetry/api';
 import { type Span } from '@opentelemetry/api';
 import type {
-  PerformanceSpan,
-  PerformanceSpanFailedOptions,
+  EmbraceExtendedSpan,
+  EmbraceExtendedSpanFailedOptions,
 } from '../../api-traces/index.js';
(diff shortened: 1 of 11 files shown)


src/managers/EmbraceTraceManager/EmbraceSpan.ts
```diff
@@ -9,17 +9,17 @@ import type {
...
 
 /**
- * EmbracePerformanceSpan for the most part simply delegates to the underlying Span it receives on initialization so
+ * EmbraceSpan for the most part simply delegates to the underlying Span it receives on initialization so
  * that it satisfies the Span interface. In addition, it gives us a spot where we can implement helpers that are part
- * of the PerformanceSpan interface.
+ * of the EmbraceSpan interface.
  */
-export class EmbracePerformanceSpan implements PerformanceSpan {
+export class EmbraceSpan implements EmbraceExtendedSpan {
```
jpmunz, line 22: minor but following the naming from before what if the interface was `ExtendedSpan` and the class implementation was `EmbraceExtendedSpan`? Do we have any best practices around this from the dashboard codebase where we need to name both an interface and a class similarly?

jpmunz (approved):

PR #403, opened 2025-05-29
chore: support sourcemap uploads in the CDN case

jpmunz:
## Goal

This was close to working as is, just needed to provide a way to to specify a bundle ID placeholder in the app's code in the `initSDK` call when the default placeholder isn't available since our SDK code isn't part of the bundle

### DIFF
--- a/src/sdk/types.ts
+++ b/src/sdk/types.ts
@@ -100,6 +100,15 @@ type BaseSDKInitConfig = {
    */
   logLevel?: DiagLogLevel;
 
+  /**
+   * bundleID should only be provided when loading the SDK from CDN through a script tag. It is used to specify a 32
+   * character placeholder string which will then be substituted by our CLI tool when uploading source maps. See
+   * "Including the SDK as a code snippet from CDN" in our README for more details.
+   *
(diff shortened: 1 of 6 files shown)


src/sdk/types.ts
```diff
@@ -100,6 +100,15 @@ type BaseSDKInitConfig = {
...
   logLevel?: DiagLogLevel;
 
+  /**
+   * bundleID should only be provided when loading the SDK from CDN through a script tag. It is used to specify a 32
+   * character placeholder string which will then be substituted by our CLI tool when uploading source maps. See
+   * "Including the SDK as a code snippet from CDN" in our README for more details.
+   *
+   * **default**: undefined
+   */
+  bundleID?: string;
```
pablomatiasgomez, line 110: Not sure I follow 100% the replacement process, but this var will always contain the template bundle ID, right? In that case, is it worth making very specific, like `templateBundleID` so that it's clear what is it used for?

pablomatiasgomez (approved):
Nice!

pablomatiasgomez (approved):

PR #405, opened 2025-05-30
feat: add instrumentation for React Router v4/v5

joaquin-diaz:
## Goal

* Add createReactRouterNavigationInstrumentation so react users can add navigation instrumentation
* Add NavigationInstrumentation as an framework-agnostic instrumentation to track navigation, used internally by the react router instrumentation

## Testing

Tested with the demo apps, adding unit tests


demo/frontend/src/otel.ts
```diff
@@ -9,6 +10,7 @@ const setupOTel = () => {
     appID: SAMPLE_APP_ID,
     spanExporters: [new ConsoleSpanExporter()],
     logExporters: [new ConsoleLogRecordExporter()],
+    instrumentations: [createReactRouterNavigationInstrumentation({})],
```
joaquin-diaz, line 13: Step 1: add the instrumentation

src/instrumentations/navigation/NavigationInstrumentation/NavigationInstrumentation.ts
```diff
@@ -0,0 +1,85 @@
...
+  EMB_TYPES,
+  KEY_EMB_TYPE,
+  KEY_VIEW_NAME,
+} from '../../../constants/index.js';
+
+// Regular expression to match path options in the format "(option)"
+// Used to clean up paths that are like "/order/:orderState(pending|shipped|delivered)" to "/order/:orderState"
+const PATH_OPTIONS_RE = /\(.*?\)/g;
+
+export class NavigationInstrumentation extends EmbraceInstrumentationBase {
```
joaquin-diaz, line 14: This can be re-used to track other react-router v6+, other libraries, frameworks, etc

demo/frontend/src/otel.ts
jpmunz: one thing that we hadn't addressed, though maybe isn't a problem in this case, is `createReactRouterNavigationInstrumentation({})` executing before `sdk.initSDK` so the tracer provider, etc. would be the default no op implementation when it first is created

probably we'd need to allow `instrumentations` to be a function that returned an array and we could execute after the needed setup for the SDK, in this case maybe it's fine if we're not relying on Embrace specific managers?

src/instrumentations/navigation/NavigationInstrumentation/react/reactRouterV5/withOTelRouting.ts
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+import hoistNonReactStatics from 'hoist-non-react-statics';
+import type {
+  RouteComponentProps,
+  SwitchedRouteComponentProps,
+} from './types.js';
+import { getNavigationInstrumentation } from '../../index.js';
+
+export const withOTelRouting = <P extends RouteComponentProps>(
```
jpmunz, line 9: I'd maybe call this `withEmbraceRouting` instead since we are applying our particular semantics on the emitted span

src/instrumentations/navigation/NavigationInstrumentation/NavigationInstrumentation.ts
```diff
@@ -0,0 +1,85 @@
...
+      this.enable();
+    }
+  }
+
+  public setCurrentRoute = (route: Route) => {
+    if (!this._config.enabled) {
+      return;
+    }
+
+    if (route.url !== this._currentRoute?.url) {
```
jpmunz, line 44: I think a couple debug logs could be helpful here going forward, one for the case where the route is the same and we don't create a span and then for indicating when we've created a new route

src/instrumentations/navigation/NavigationInstrumentation/react/reactRouterV5/withOTelRouting.ts
```diff
@@ -0,0 +1,36 @@
...
+) => {
+  const RouteWithOTelRouting: React.FC<P> = (props: P) => {
+    const navigationInstrumentation = getNavigationInstrumentation();
+    // Make sure this is Route component
+    if (props.path) {
+      // Routes get injected with computedMatch when they are children of a <Switch> but the types do not reflect that
+      // Manually setting the type here.
+      const routeProps = props as SwitchedRouteComponentProps;
+
+      if (routeProps.computedMatch) {
```
jpmunz, line 20: do we need to be careful about either routeProps or navigationInstrumentation being undefined here?

package.json
```diff
@@ -96,6 +105,7 @@
     "@opentelemetry/otlp-exporter-base": "0.57.2",
     "@opentelemetry/sdk-trace-web": "1.30.1",
     "@opentelemetry/web-common": "0.57.2",
+    "hoist-non-react-statics": "^3.3.2",
```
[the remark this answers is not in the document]
joaquin-diaz: This is a needed dependency, but I didn't want to add it as a peer dependency because is not common for react apps to have this installed and it would require users to add it to their package json maybe creating confusion. The package is ~1kg gzipped so it should be fine

package.json
```diff
@@ -132,5 +144,11 @@
     "typedoc": "0.28.4",
     "typescript": "5.8.3",
     "typescript-eslint": "8.32.1"
+  },
+  "peerDependencies": {
+    "react": "^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0"
+  },
+  "optionalDependencies": {
```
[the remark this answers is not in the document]
joaquin-diaz: Ah yeah, I read about both and I thought they served the same purpose but they don't. The `optionalDependencies` are mean for packages that optional `dependencies` but it doesn't work for `peerDependencies`. I'll change it

src/instrumentations/navigation/NavigationInstrumentation/react/reactRouterV5/withOTelRouting.ts
joaquin-diaz: Yep. I was trying to avoid tying the instrumentation to us as much as possible but you're right that we're using our own internal semantics

src/instrumentations/navigation/NavigationInstrumentation/react/reactRouterV5/withOTelRouting.ts
joaquin-diaz: Props are never going to be undefined. If for whatever reason users are skipping ts type checks and wrapping a non `Route` component with `withEmbraceRouting` we should be safe by checking first for `props.path` and then `routeProps.computedMatch`. `getNavigationInstrumentation` always returns an instance of `navigationInstrumentation`

rollup.config.js
```diff
@@ -12,10 +12,15 @@ const externalDeps = [
   ...Object.keys(pkg.peerDependencies || {}),
 ];
 
+const input = {
+  index: 'src/index.ts',
+  'react-instrumentation': 'src/react/index.ts',
```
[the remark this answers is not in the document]
joaquin-diaz: Right. If the host app doesn't call `withEmbraceRouting` in theory it should be fine to even import it in the code since it should get tree-shaken but I wanted to be extra careful and explicit, I also think it looks closer to what is going to be when we have a standalone package

PR #426, opened 2025-06-06
chore: add update-types for otel packages in dependabot.yml and move @opentelemetry/api and @opentelemetry/sdk-trace-web to peerDependencies

joaquin-diaz:
## Goal

* add update-types for otel packages in dependabot.yml 
* move @opentelemetry/api and @opentelemetry/sdk-trace-web to peerDependencies

## Testing

n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,14 +39,20 @@ Currently, only Spans and Logs are supported, but other signals will be added in
 npm:
 
 ```sh
-npm install @embrace-io/web-sdk
+npm install --save @embrace-io/web-sdk @opentelemetry/api@^1 @opentelemetry/sdk-trace-web@^1
+
 ```
 
 yarn:
 
 ```sh
-yarn add @embrace-io/web-sdk
+yarn add @embrace-io/web-sdk @opentelemetry/api@^1 @opentelemetry/sdk-trace-web@^1
 ```
+
+#### Why do I need to install `@opentelemetry/api` and `@opentelemetry/sdk-trace-web`?
+
+These packages are used internally to emit telemetry in the OpenTelemetry format. We list them as `peerDependencies` to 
+avoid bundling them and prevent version conflicts if you use the OpenTelemetry SDK directly in your application.
 
 > [!TIP]
 > For CDN installs, see [Including the SDK as a code snippet from CDN](#including-the-sdk-as-a-code-snippet-from-cdn).
--- a/package.json
+++ b/package.json
@@ -110,14 +110,12 @@
     "telemetry"
   ],
   "dependencies": {
-    "@opentelemetry/api": "1.9.0",
     "@opentelemetry/instrumentation": "0.57.0",
     "@opentelemetry/instrumentation-document-load": "0.44.1",
     "@opentelemetry/instrumentation-fetch": "0.57.2",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -160,6 +160,8 @@
     "typescript-eslint": "8.33.0"
   },
   "peerDependencies": {
+    "@opentelemetry/api": ">=1.0.0 <2.0.0",
```
jpmunz, line 163: should this be 1.9.0? Do we make use of anything additional that wouldn't be available on 1.0?

package.json
```diff
@@ -110,14 +110,12 @@
     "telemetry"
   ],
   "dependencies": {
-    "@opentelemetry/api": "1.9.0",
     "@opentelemetry/instrumentation": "0.57.0",
     "@opentelemetry/instrumentation-document-load": "0.44.1",
     "@opentelemetry/instrumentation-fetch": "0.57.2",
     "@opentelemetry/instrumentation-xml-http-request": "0.57.2",
     "@opentelemetry/opentelemetry-browser-detector": "0.57.2",
     "@opentelemetry/otlp-exporter-base": "0.57.2",
```
jpmunz, line 118: is `otlp-exporter-base` still ok as a direct dependencies? Are there issues with someone setting up their own OTLP export on a different version?

package.json
joaquin-diaz: Not sure tbh. I just guessed that they wouldn't have breaking changes but yeah I'll add as a minimum just to be sure. Same with the other package

package.json
joaquin-diaz: They need to import https://www.npmjs.com/package/@opentelemetry/exporter-trace-otlp-http to set up their own exporter, we use `otlp-exporter-base` internally because we use `fetch` instead of `xhr` as transport so we can support `keepalive`. I guess the only issue would be if they want to do something custom with `otlp-exporter-base` but I guess that also applies to the rest of the packages? I'm not sure how we want to proceed there, I'd say for now we only add as peer dependencies the ones that are most likely to be used outside embrace

package.json
```diff
@@ -160,6 +160,8 @@
     "typescript-eslint": "8.33.0"
   },
   "peerDependencies": {
+    "@opentelemetry/api": ">=1.9.0 <2.0.0",
+    "@opentelemetry/sdk-trace-web": ">=1.30.1 <2.0.0",
```
jpmunz, line 164: I think could relax slightly to 1.30.0, that's what we list here: https://github.com/embrace-io/embrace-web-sdk/blob/ea4f22c05d36a26f0846ef2520893e57e952e757/README.md#compatibility-with-otel-packages

package.json
jpmunz: yep agreed, we can apply the same strategy to other dependencies if needed later

README.md
```diff
@@ -39,15 +39,21 @@ Currently, only Spans and Logs are supported, but other signals will be added in
 npm:
 
 ```sh
-npm install @embrace-io/web-sdk
```
jpmunz, line 42: isn't `npm install @embrace-io/web-sdk` still valid?

joaquin-diaz:
Closing pending more research on how can we avoid peerDependencies in favour of making it less complicated to install the SDK

PR #441, opened 2025-06-11
chore: repair demo import path, fix a few lints

overbalance:
## Goal

In running the web demo to get familiar with the system, I encountered a build error because a build path pointed to an old bundle.

## Testing

I successfully ran the build step locally. Ideally an automated e2e test would cover navigating the test page and clicking buttons to ensure the sdk is actually loaded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,8 @@
 .idea/
 node_modules
 build
-demo/.DS_Store
+dist
 /docs/
 /coverage/
-**/.DS_Store
-/tsconfig.*.tsbuildinfo
-/tsconfig.tsbuildinfo
-dist
+.DS_Store
+*.tsbuildinfo


eslint.config.mjs
```diff
@@ -120,7 +120,7 @@ export default tseslint.config({
         // import type { something } from './index.js'; is valid
         {
           regex:
-            "import .* from '\\.{1,2}(?:\\/\.\.)*\\/([a-zA-Z0-9_-])+\\/([a-zA-Z0-9_-])+\\/.+'",
+            "import .* from '\\.{1,2}(?:\\/..)*\\/([a-zA-Z0-9_-])+\\/([a-zA-Z0-9_-])+\\/.+'",
```
jpmunz, line 123: could you explain this one? Previously I believe this was looking for relative imports so was trying to match ".." literally rather than any 2 characters

joaquin-diaz (approved):

jpmunz (approved):

PR #442, opened 2025-06-13
chore: refactor build pipeline

overbalance:
## Goal

Refactor the build pipeline to serve esm, commonjs, or a complete umd bundle.
- Adjusted typechecking to cover all src and curb overzealous linting of configs
- Added sourcemaps for SDK consumers
- Swapped out rollup plugins for swc
- Fewer package.json exports
- Pinned more dependencies
- Added React as a devDep because it's needed for the React demo

## Testing

Ran tests locally. Launched both frontend demos successfully.


jpmunz:
> Do you think is worth start adding some docs around the build process? i.e. how we leverage on `swc` to define the outputs instead of each `tsconfig`

ya this would help me as well, the build process is definitely a bit of a black box for me and I would find it useful to have a breakdown of the tools we're using here and what it gets us over other possible options

eslint.config.mjs
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable regex/invalid */
```
jpmunz, line 1: why was this needed?

overbalance: this file is now included when linting and when you lint the rule it says it's invalid. fractal code

rollup.config.mjs
```diff
@@ -0,0 +1,95 @@
...
+  plugins,
+  external: externalDeps,
+  onwarn,
+};
+
+const cdnConfig = {
+  input: input.index,
+  output: {
+    file: 'build/umd/index.js',
+    format: 'umd',
```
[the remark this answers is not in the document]
overbalance: UMD discovers the module system in use and permits the SDK to be imported into older systems using CommonJS, AMD, or window vars. IIFE only sets a window variable.

this is a conservative choice that provides greater compatibility. ESM has been recommended since 2015 so we should encourage its use, but if someone really can't upgrade, this gives them the chance to use embrace.

PR #457, opened 2025-06-18
chore: use npm workspaces

overbalance:
## Goal

Centralize npm deps at the root of the project to save disk space and confusion around scripts that need to install in child directories before importing the sdk. You no longer need to `npm i` in child directories and if you do it will install all dependencies in all workspaces, including the root.

## Testing

Ensured node_modules needed in workspaces are symlinked to the root when overlapping, and included in the individual workspace when not shared.


joaquin-diaz (approved):
Looks good! Just to be sure: we don't need to change anything in the build process to avoid adding unnecessary dependencies coming from the demo apps to the final build?

overbalance:
The final build remains the same, but it does mean the top-level node_modules is larger for someone who never runs the demo or the cli. What do you think?

overbalance:
@joaquin-diaz The final build remains the same, but it does mean the top-level node_modules is larger for someone who never runs the demo or the cli. What do you think?

joaquin-diaz:
> @joaquin-diaz The final build remains the same, but it does mean the top-level node_modules is larger for someone who never runs the demo or the cli. What do you think?

That's ok, I don't think it is an issue

PR #461, opened 2025-06-20
EMBR-7999 feat: data limit enforcement

jpmunz:
Adding in client-side limiting of data, there's a bunch of file changes because things started taking a limitManager in their constructor, main changes are in:
* src/managers/EmbraceLimitManager/EmbraceLimitManager.ts -> A new internal manager that enforces limits and reports diagnostics on how many things are being dropped/truncated
* src/managers/EmbraceLimitManager/constants.ts -> actual limits to enforce, still need to check this is consistent with our documentation + other SDKs
* src/managers/EmbraceLogManager/EmbraceLogManager.ts -> uses the limit manager when creating logs through our API
* src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts -> uses the limit manager for session properties and breadcrumbs, also resets the limits when a session ends
* src/processors/EmbraceSessionBatchedSpanProcessor/EmbraceSessionBatchedSpanProcessor.ts -> uses the limit manager to drop spans instead of exporting if we're over the max
* src/sdk/initSDK.ts -> setups the limit manager, also start specifying some of the existing limit configuration available from the otel log and tracer providers

This is a bit of a mix and match of the limits being applied at the Embrace APIs or at the OTel level, should discuss if we want more consistency either here or during a later iteration, part of the issue I ran into doing more things at the processor level was dealing with read-only objects at that point, currently:
* For individual spans attribute+event limits are handled by the Otel WebTracerProvider, max spans per sessions are handled in our batched session span processor
* For logs all limits are handled in the Embrace log API, logs added directly by an OTel provider would bypass the limits


src/managers/EmbraceLimitManager/EmbraceLimitManager.ts
```diff
@@ -0,0 +1,184 @@
...
+      info_log: 0,
+      breadcrumb: 0,
+      session_property: 0,
+      span: 0,
+      network_request: 0,
+    };
+  }
+
+  private _incrDiagnosticCount(type: LimitedType, operation: LimitOperation) {
+    const key = `emb.applied_limit.${type}.${operation}.count`;
```
jpmunz, line 177: @pablomatiasgomez sending things like "emb.applied_limit.breadcrumb.drop.count" as internal diagnostic attributes on the session span so that we can keep track of it later, I don't believe we currently have anything being read for that on the backend?

src/managers/EmbraceLimitManager/constants.ts
```diff
@@ -0,0 +1,27 @@
+import type { EmbraceLimitManagerArgs } from './types.js';
+
+export const DEFAULT_MAX_LOG_ATTRIBUTES = 50;
+
+export const DEFAULT_LIMITS: EmbraceLimitManagerArgs = {
+  maxAllowed: {
+    error_log: 500,
```
pablomatiasgomez, line 7: Maybe add a comment that these apply on a "per session" basis?

src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -76,9 +85,23 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
+      return;
+    }
+
+    this._sessionSpan.setAttribute(
+      KEY_PREFIX_EMB_PROPERTIES + limitedSessionProperty.key,
+      limitedSessionProperty.value
+    );
   }
 
+  // the external api doesn't include a reason, and if a users uses it to end a session, the reason will be 'user_ended'
```
overbalance, line 104: I don't see `user_ended` anywhere in the sdk ?
```suggestion
  // the external api doesn't include a reason, and if a users uses it to end a session, the reason will be 'manual'
```

overbalance (approved):

src/managers/EmbraceLimitManager/EmbraceLimitManager.ts
fnewberg: please coordinate with the ios and android teams on how you do this so we're consistent across platforms

src/managers/EmbraceLimitManager/constants.ts
```diff
@@ -0,0 +1,27 @@
...
+    session_property: 100,
+    span: 1_000,
+    network_request: 10_000,
+  },
+  maxLength: {
+    error_log: 128,
+    warning_log: 128,
+    info_log: 128,
+    breadcrumb: 256,
+    session_property: 256,
```
fnewberg, line 20: are there separate limits for session property keys and values?

similarly, do we have limits for span attribute key and value lengths?

src/sdk/initSDK.test.ts
```diff
@@ -461,6 +478,239 @@ describe('initSDK', () => {
...
+            // This instrumentation does its own patching of Fetch which interferes with our test stub
+            '@opentelemetry/instrumentation-fetch',
+            // Document load instrumentation generates a bunch of spans in this test environment
+            '@opentelemetry/instrumentation-document-load',
+          ]),
+        },
+      });
+      void expect(result).not.to.be.false;
+
+      await new Promise(r => setTimeout(r, 1));
```
joaquin-diaz, line 594: Maybe add a comment why do we need to wait here?

joaquin-diaz (approved):

src/managers/EmbraceLimitManager/EmbraceLimitManager.ts
pablomatiasgomez: Yaah we don't have anything for this right now. I would say we should add a section here https://www.notion.so/embraceio/1d77e3c9985280eb9501daed1f83f1f0?v=1d77e3c9985281a6b022000c8a520048 and describe these, and tell mobile teams so that they are aware if they want to implement it too? What do you think?

src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
jpmunz: yep looks to have been out of date, added

src/managers/EmbraceLimitManager/EmbraceLimitManager.ts
jpmunz: yep makes sense, I added a page to that table: https://www.notion.so/embraceio/Internal-application-of-data-limit-21f7e3c998528018980bfd65f410d23a?source=copy_link

src/managers/EmbraceLimitManager/constants.ts
jpmunz: this is shared between session property keys and values

no limits actually on span (or log) attribute key + values, I can look to add those though the built-in limiting from the otel tracer provider seems to only provide an attribute value length limit

jpmunz: added separate values for session property values and keys, added checks for log attribute key + values, I left off span attributes for now since we'll need to rethink using the builtin otel limiting there, filed: https://www.notion.so/embraceio/Expand-SDK-side-data-limiting-2257e3c99852808f925bd50f1afae677?source=copy_link

PR #474, opened 2025-06-23
chore: add new Sonda report - disabled by default

overbalance:
## Goal

Prepare for Sonda reports locally and in CI.

## Testing

Build is successful. Sonda does not output html or json.


overbalance:
Moved changes to a new PR

PR #477, opened 2025-06-24
chore: fix 

overbalance:
## Goal

1. Suppress lint errors for *.config.* files
2. Explicitly list all dependencies to ensure pinned versions are used
3. Explicitly list ecmascript rollup output versions
4. Update ambiguous type: `trace.setSpan`
5. Remove tsconfig.base.json inheritance scheme: not all tools support `extends`
6. Force IIFE build to pull ESNext versions of deps so the smallest amount of code is fetched before compression

## Testing

Demo runs. IIFE bundle size is down to 147kb.


tsconfig.types.json
```diff
@@ -1,5 +1,6 @@
 {
-  "extends": "./tsconfig.base.json",
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "extends": "./tsconfig.json",
```
jpmunz, line 3: is `extends` still ok to use here?

rollup.config.js
```diff
@@ -62,27 +65,33 @@ export default defineConfig([
...
-    external: externalDeps,
+    external: isExternal,
   },
 
   // CDN Build, it only exports the core web sdk and not any additional instrumentation
   {
     input: 'src/index.ts',
     plugins: [
       typescript({
-        tsconfig: './tsconfig.esm.json',
```
jpmunz, line 81: previously this `tsconfig.esm.json` config contained a couple extra keys for `"module": "ES6"` and `"moduleResolution": "bundler"`, are they no longer needed? Similar question for the other targets

rollup.config.js
```diff
@@ -34,15 +36,16 @@ export default defineConfig([
       preserveModules: true,
       preserveModulesRoot: 'src',
     },
-    external: externalDeps,
+    external: isExternal,
```
jpmunz, line 39: this looks to have been an array previously is it okay to pass a boolean here?

PR #482, opened 2025-06-24
chore: add back deleted tsconfig, cleanup

overbalance:
## Goal

Add back a deleted tsconfig we relied on.

I was trying to have rollup be a single source of truth for esm output version but it's useful to keep a tsconfig around to check compilation even when not transpiling for the browser.

Additionally:
1. Remove unused `.npmrc`
2. Add `node:` prefix to cli imports as a best practice
3. Ensure we only have one copy of Sonda and React types
4. Make the cli executable on Mac after build
5. Remove react() from vite because it already parses jsx
6. Revise eslint to focus on src files

## Testing

CI passes, lint passes locally, compilation passes locally.

### DIFF
--- a/demo/frontend/package.json
+++ b/demo/frontend/package.json
@@ -28,7 +28,6 @@
     "@types/react": "19.1.6",
     "@types/react-dom": "19.1.6",
     "@types/react-router-domv4v5": "npm:@types/react-router-dom@^5.3.3",
-    "@vitejs/plugin-react": "4.5.2",
     "globals": "16.2.0",
     "rimraf": "6.0.1",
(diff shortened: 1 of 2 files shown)


tsconfig.types.json
```diff
@@ -7,8 +7,6 @@
     "declarationMap": true,
     "emitDeclarationOnly": true,
     "declarationDir": "build/types",
-    "rootDir": "src",
```
jpmunz, line 10: is this still needed here? I see we have it specified in `tsconfig.json`

jpmunz (commented):
lgtm just the one question

PR #486, opened 2025-06-26
chore: cleanup before permanent props merge

overbalance:
## Goal

Small changes to whitespace, type imports, and docs I pulled out of a [larger PR](https://github.com/embrace-io/embrace-web-sdk/pull/478) that was making it hard to review.

## Testing

TSC, unit tests, and CI.


src/api-sessions/manager/index.ts
```diff
@@ -1,3 +1,4 @@
-export type { SpanSessionManager, ReasonSessionEnded } from './types.js';
+export type { ReasonSessionEnded, SpanSessionManager } from './types.js';
```
jpmunz, line 1: for the various import changes with reordering, combining ones from the same package, etc. is there something you ran locally that altered those automatically? We should see if we can include that as part of the repo's linting since atm they don't get flagged for me so I'd probably end up reintroducing things like this over time

overbalance: The changes you see in this PR were enacted manually. I'll make a point of not sorting until we make a decision about tooling.

I already have the config sitting on a branch. It's not much more work to modify prettier (oxc formatter is in beta) and update commitlint.

joaquin-diaz (approved):

PR #501, opened 2025-06-30
chore: add tests to analyze bundle size

joaquin-diaz:
## Goal

Add a test that builds an app using vite and two targets (for now) to measure the size of the SDK

## Testing

See comment for output

### DIFF
--- a/tests/bundle/utils/tests-shared.ts
+++ b/tests/bundle/utils/tests-shared.ts
@@ -1,0 +1,129 @@
+import test from 'node:test';
+import { processSondaReport } from './index';
+import assert from 'node:assert';
+import { dirname, resolve } from 'node:path';
+import { TOTAL_GZIP_SIZE_THRESHOLD_IN_KB } from '../config';
+import fs from 'node:fs';
+import { resultsToMarkdownTable } from '../../utils';
+import { fileURLToPath } from 'node:url';
+import { promisify } from 'node:util';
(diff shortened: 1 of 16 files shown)


tests/bundle/utils/tests-shared.ts
```diff
@@ -0,0 +1,129 @@
...
+ * It checks if the build for each target runs without errors and optionally checks the bundle size.
+ *
+ * It requires the bundler app to have a `build:clean` script to clean the build directory before running the tests.
+ * For each target, it runs the `build:<target>` script and checks the output.
+ *
+ * If `includeBundleSizeTest` is true, it processes the Sonda report to check the total gzip size against a threshold,
+ * producing a markdown report with the results in `./test-results/<bundlerName>-tests.md`.
+ * Sonda output is expected to be in the `.sonda/<target>/sonda_0.json` file in the app directory.
+ */
+const runBundlerBuildSmokeTest = async (
```
joaquin-diaz, line 35: @embrace-io/team-web the most important part of the test is this file, the rest is just boilerplate for webpack and vite. It's been a while since I set up a webpack app, so maybe @overbalance you can take a look to see if the setup makes sense and if it can be compared as is with vite. Build numbers are pretty much the same so in that way it looks fine to me

pablomatiasgomez (approved):

overbalance (approved):

PR #511, opened 2025-07-03
feat: explicitly export sdk to window

overbalance:
## Goal

In addition to using a script tag to load the bundled IIFE version of the sdk, allow users to lazy load: `await import('https://cdn.jsdelivr.net/npm/@embrace-io/web-sdk@1')` 

## Testing

CDN demo is working.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -113,8 +113,13 @@ export default defineConfig([
     output: {
       file: 'build/iife/bundle.js',
       format: 'iife',
-      name: 'EmbraceWebSdk', // Global variable name for the SDK
+      // global variable name for the SDK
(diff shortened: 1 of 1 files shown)


rollup.config.js
```diff
@@ -115,6 +115,7 @@ export default defineConfig([
       format: 'iife',
       name: 'EmbraceWebSdk', // Global variable name for the SDK
       sourcemap: true,
+      footer: 'window.EmbraceWebSdk = EmbraceWebSdk;',
```
jpmunz, line 118: I think worth a comment here adding the context we had discussed in standup, e.g. normally when including using X method sdk is not set on global but adding this workaround etc.

PR #512, opened 2025-07-03
fix: do not require token if cli is called with --no-upload

pablomatiasgomez:
## Goal

Currently we are calling with `--no-upload` in our frontend demo: https://github.com/embrace-io/embrace-web-sdk/blob/main/demo/frontend/scripts/runDemo.sh#L25 and this is failing as we are not passing a token, which is not used so shouldn't be mandatory.

### DIFF
--- a/cli/src/uploadToApi.ts
+++ b/cli/src/uploadToApi.ts
@@ -27,6 +27,13 @@ export const uploadToApi = async ({
   dryRun,
   upload,
 }: UploadToApiArgs): Promise<void> => {
+  console.log(
+    upload && !dryRun ? 'Uploading to Embrace API' : 'Dry run, skipping upload'
+  );
+  if (dryRun || !upload) {
+    return;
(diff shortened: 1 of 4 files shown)


cli/src/uploadToApi.ts
```diff
@@ -27,6 +27,13 @@ export const uploadToApi = async ({
...
 }: UploadToApiArgs): Promise<void> => {
+  console.log(
+    upload && !dryRun ? 'Uploading to Embrace API' : 'Dry run, skipping upload'
+  );
+  if (dryRun || !upload) {
+    return;
+  }
+
   // prepare the body for the API request as a gzipped JSON object
   const body = new Blob([
```
pablomatiasgomez, line 38: Moved the creation of the body after checking if we have to upload, otherwise this is being done for no reason

joaquin-diaz (approved):

PR #513, opened 2025-07-03
feat: do not end sessions if visibility didn't change within a small period of time

pablomatiasgomez:
## Goal

Change the visibility tracker to do:

- If visibility changes from hidden to visible, start the session immediately
- If visibility changes to hidden, wait `visibilityWaitTimeMs` (which defaults to 15s) to see if the visibility changes to visible again, and do not end the session if so.

## Testing

Test locally with the frontend demo, and added individual tests.


jpmunz (approved):
change itself looks good, the main danger with releasing is if the assumption that we'll still get a chance to record the session 15s after a tab or browser close breaks in some browsers or environments we will miss sending those sessions that we otherwise would have

some mitigation ideas we can discuss doing some of or all of these:
* default `visibilityWaitTimeMs` to 0 and skip the timeout logic in this case, in our own dashboard integration we can set this to 15000 for testing and then later update the default
* wait for https://github.com/embrace-io/embrace-web-sdk/pull/538 to be merged to see if this fails any of the added test cases
* include an extra condition on the length of the current session and only apply the timeout logic if it is less than a certain amount. This would give protection against losing a long session, in this case we would record 2 sessions when switching tabs from a long running session and then quickly back but would not record multiple quick switches back to back

src/instrumentations/session/SpanSessionVisibilityInstrumentation/SpanSessionVisibilityInstrumentation.ts
```diff
@@ -15,6 +20,39 @@ export class SpanSessionVisibilityInstrumentation extends EmbraceInstrumentation
...
+      ) {
+        this._currentVisibilityState = visibilityDoc.visibilityState;
+        this._onVisibilityChange();
+        return;
+      }
+
+      this._diag.debug(
+        `Visibility changed to ${visibilityDoc.visibilityState}. Will wait ${(visibilityWaitTimeMs / 1000).toString()}s, and check if visibility changed`
+      );
+      this._checkVisibilityTimeout = setTimeout(() => {
```
jpmunz, line 44: not needed here but a note for a follow-up it might be nice to record a 'tab lost focus' breadcrumb here for informational purposes

pablomatiasgomez:
I will still wait for Joaco's PR to get merged so that I can see if any test won't pass. Then I will merge this one, release, and configure it in dashjs to have a timeout of 15s. All other clients shouldn't see any change

PR #514, opened 2025-07-03
feat: add emb.session_number attribute to session span

pablomatiasgomez:


demo/frontend/src/App.tsx
```diff
@@ -304,15 +231,128 @@ const App = () => {
...
+          </button>
+          <button onClick={() => setNavigationType('declarativeV6+')}>
+            Enter react-router v6+ declarative navigation demo
+          </button>
+          <button onClick={() => setNavigationType('data')}>
+            Enter react-router v6+ data navigation demo
+          </button>
+        </div>
+
+        {/* Navigation: */}
```
pablomatiasgomez, line 343: Added this section to test how the sdk behaves when the app navigates. Tested mostly the opening on a new tab with cmd+tab very consecutive and couldn't get any session_number that overlaps so far

src/sdk/initSDK.test.ts
```diff
@@ -261,6 +261,10 @@ describe('initSDK', () => {
 
     it('should include the correct resource attributes', async () => {
       fakeFetchRespondWith('');
+
+      // Hardcode the session_number so that we validate it gets incremented by one.
+      localStorage.setItem('embrace_session_number', '15');
```
pablomatiasgomez, line 266: Not sure if there's a better way to do this without knowing the key.. I didn't want to export the const as it really shouldn't be exposed, but at the same time I need to know the value here to be able to predict the test output.. If you have a better way, let me know.

src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.test.ts
```diff
@@ -368,6 +370,20 @@ describe('EmbraceSpanSessionManager', () => {
...
+    expect(sessionSpan.attributes).to.have.property('emb.session_number', 2);
+    memoryExporter.reset();
+
+    manager.startSessionSpan();
+    manager.endSessionSpan();
+
+    finishedSpans = memoryExporter.getFinishedSpans();
+    expect(finishedSpans).to.have.lengthOf(1);
+    sessionSpan = finishedSpans[0];
+    // Any following session should have emb.cold_start = false
```
jpmunz, line 384: minor, you could add one more case here where you instantiate a new manager w/ the same storage instance, e.g. `manager = new EmbraceSpanSessionManager({ diag, storage, limitManager });`

in that case you should get a cold start again but a session number of 4

src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -75,6 +77,20 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
   }
 
+  // Increment and return the session number stored in local storage.
+  // This is not perfect in the sense that there may be a race condition between tabs.
+  // Eventually a lock could be implemented, but for now this solution should work fine.
+  public _getSessionNumber(): number {
+    const value = this._storage.getItem(EMBRACE_SESSION_NUMBER_STORAGE_KEY);
+    let number = value ? parseInt(value, 10) : 0;
+    number++;
+    this._storage.setItem(
```
jpmunz, line 87: these accesses to storage should be protected with try/catches, but we already have a separate ticket for that so I think ok as is and I'll prioritize doing that across all instances once this is merged

src/sdk/initSDK.test.ts
jpmunz: I think this is okay if it seems to behave fine in CI, alternatively in the test you could end the session twice and use the payload just for the purposes of reading 'emb.session_number' and then do the full expectation against the 2nd payload with that number incremented

jpmunz (approved):
lgtm with a couple comments

PR #524, opened 2025-07-06
chore: add test for iife bundle and cdn

overbalance:
## Goal

Ensure `window.EmbraceWebSdk` is set when loading the iife bundle via script tag and dynamic import.

## Testing

CI passes.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1,4 +1,32 @@
 import * as unused from './index.js';
+import { expect } from 'chai';
 
 // @ts-expect-error keep this import for testing coverage checking
 const _unusedReference = unused;
+
+describe('sdk browser bundle loading', () => {
+  it('jsdelivr should serve the cdn bundle under 2000ms', async () => {
+    // @ts-expect-error types cannot be resolved dynamically
+    // eslint-disable-next-line import/no-unresolved
+    await import('https://cdn.jsdelivr.net/npm/@embrace-io/web-sdk');
+
+    // @ts-expect-error sdk should exist in the window
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    void expect(window.EmbraceWebSdk.sdk).to.not.be.undefined;
+  });
+
+  it('should be able to dynamically import the cdn bundle', async () => {
+    // @ts-expect-error bundle is served on localhost by web-test-runner - not from disk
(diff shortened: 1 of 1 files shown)


eslint.config.mjs
```diff
@@ -149,11 +149,7 @@ export default tseslint.config([
     languageOptions: {
       globals: globals.browser,
       parserOptions: {
-        project: [
-          './tsconfig.json',
-          './tsconfig.test.json',
-          './cli/tsconfig.json',
-        ],
+        project: ['./tsconfig.json', './cli/tsconfig.json'],
```
overbalance, line 152: The sdk isn't big enough to warrant a restriction on type-checking all files.

src/index.test.ts
```diff
@@ -1,4 +1,32 @@
...
+
+    // @ts-expect-error sdk should exist in the window
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    void expect(window.EmbraceWebSdk.sdk).to.not.be.undefined;
+  });
+
+  it('should be able to dynamically import the cdn bundle', async () => {
+    // @ts-expect-error bundle is served on localhost by web-test-runner - not from disk
+    // eslint-disable-next-line import/no-unresolved
+    await import('/build/iife/bundle.js');
```
overbalance, line 21: This isn't really a unit test because it requires a full compilation of the sdk. Is this the right place for this?

joaquin-diaz: I think we can move them to the e2e test suite under bundle, I can help you set that up

src/index.test.ts
```diff
@@ -1,4 +1,32 @@
...
+import { expect } from 'chai';
 
 // @ts-expect-error keep this import for testing coverage checking
 const _unusedReference = unused;
+
+describe('sdk browser bundle loading', () => {
+  it('jsdelivr should serve the cdn bundle under 2000ms', async () => {
+    // @ts-expect-error types cannot be resolved dynamically
+    // eslint-disable-next-line import/no-unresolved
+    await import('https://cdn.jsdelivr.net/npm/@embrace-io/web-sdk');
```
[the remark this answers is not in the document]
joaquin-diaz: We should probably re-write as e2e test and spin up a local node server that serves this (I already did it for the lighthouse tests)

PR #551, opened 2025-07-09
chore: add dev and demo helpers

overbalance:
## Goal

1. To give users a single command to start the demo `npm run demo`
2. To make it easier to work on the SDK by launching the react demo vite dev server and rollup watch at the same time with `npm run dev`

## Testing

Both commands work without regressions to other build processes.


overbalance:
@jpmunz Definitely will update the docs since we are keeping it. Thanks!

joaquin-diaz (approved):

PR #573, opened 2025-07-14
chore: create example with otel@latest

overbalance:
## Goal

Set up a vite app using OpenTelemetry "latest" to track our OpenTelemetry 2.x support progress.

## Outcome

After testing multiple ways to bundle and export our sdk, we decided to add docs stating we only support OpenTelemetry API 1.x, with a plan to upgrade our support soon.


tests/bundle/otel-latest/README.md
jpmunz, line 1: this is great, thanks for adding!

jpmunz (approved):
lgtm and a good test bed for then seeing what gets resolved with us moving to the 2.0 packages. Filed a follow-up as well for seeing the behaviour if the host app has just `@opentelemetry/exporter-logs-otlp-http` and `@opentelemetry/exporter-trace-otlp-http` without the other dependencies and seeing if we hit the same issues there or not

overbalance:
@joaquin-diaz I moved it, thanks!

PR #579, opened 2025-07-15
chore: implement sampling

joaquin-diaz:
## Goal

* Add same sampling algorithm as the rest of the SDKs
* Avoid registering all of Embrace global managers if the SDK device is sampled out

### Future Work:

* Refresh remote config periodically. Android does it every hour
* Expose functions to enable or disable the SDK 
* Check sampling state for deviceId more regularly to avoid long living tabs to be sampled out forever  

## Testing

Added unit and e2e test


src/utils/isDeviceIdSampled.test.ts
```diff
@@ -0,0 +1,24 @@
...
+    void expect(isDeviceIdSampled(SAMPLED_UUID, 101)).to.be.false;
+  });
+
+  it('should return true for a device ID that is sampled', () => {
+    void expect(isDeviceIdSampled(SAMPLED_UUID, 50)).to.be.true;
+  });
+
+  it('should return false for a device ID that is not sampled', () => {
+    void expect(isDeviceIdSampled(NOT_SAMPLED_UUID, 50)).to.be.false;
+  });
```
jpmunz, line 23: just to make sure we've matched the algorithm I think worth including these exact test cases from the backend:
https://github.com/embrace-io/go/blob/master/lib/common/id_test.go#L276

src/sdk/initSDK.ts
```diff
@@ -137,6 +138,17 @@ export const initSDK = (
...
 
+    const sdkFeaturesManager = new EmbraceSDKFeaturesManager({
+      dynamicConfigManager,
+      deviceId: enduserPseudoID,
+    });
+
+    if (!sdkFeaturesManager.isSDKEnabled()) {
+      diagLogger.debug('SDK is disabled, skipping initialization.');
+
+      return false;
```
jpmunz, line 149: this looks good to me for this pass, let's discuss what we are leaving out here that would make sense as eventual follow-ups vs. what we never want to do, e.g. is polling periodically for new configs something we'd eventually want to include?

src/sdk/initSDK.test.ts
```diff
@@ -1062,6 +1064,30 @@ describe('initSDK', () => {
...
+      const diagLogger = new InMemoryDiagLogger();
+      const result = initSDK({
+        appID: 'abc12',
+        diagLogger,
+        dynamicSDKConfigManager: myCustomConfigManager,
+        logExporters: [logExporter],
+        spanExporters: [spanExporter],
+      });
+      void expect(result).to.be.false;
+      // We can't test the global managers not being set because other tests will have already initialized it
```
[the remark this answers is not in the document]
jpmunz: or there may be value in our APIs including `reset()` calls

src/sdk/initSDK.ts
```diff
@@ -137,6 +138,17 @@ export const initSDK = (
       });
     void dynamicConfigManager.refreshRemoteConfig();
 
+    const sdkFeaturesManager = new EmbraceSDKFeaturesManager({
+      dynamicConfigManager,
+      deviceId: enduserPseudoID,
+    });
+
+    if (!sdkFeaturesManager.isSDKEnabled()) {
```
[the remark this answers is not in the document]
joaquin-diaz: Yes, if the user doesn't provide anything yes. But they can provide a defaultConfig where they set a different sampling pct

src/sdk/initSDK.ts
joaquin-diaz: I'll add it to the PR description

src/sdk/initSDK.test.ts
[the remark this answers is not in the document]
joaquin-diaz: I'll give it a try, I wasn't sure how parallel tests are. If another test is setting the globalManager at the same time it probably won't work

src/sdk/initSDK.ts
jpmunz: filed https://www.notion.so/embraceio/Remote-config-sampling-improvements-2327e3c99852809ab216caf0e4be83d1?source=copy_link

PR #581, opened 2025-07-15
chore: include user_agent.original resource attribute

pablomatiasgomez:
## Goal

Include the stable user_agent.original attribute as described in https://opentelemetry.io/docs/specs/semconv/registry/attributes/user-agent/

### DIFF
--- a/src/resources/webSdkResource.test.ts
+++ b/src/resources/webSdkResource.test.ts
@@ -39,6 +39,7 @@ describe('webSdkResource', () => {
       'telemetry.sdk.language': 'webjs',
       'telemetry.sdk.name': 'embrace-web-sdk',
       'telemetry.sdk.version': SDK_VERSION,
+      'user_agent.original': window.navigator.userAgent,
     });
   });
 
--- a/tests/integration/tests/__golden__/next-latest-send-log.json
+++ b/tests/integration/tests/__golden__/next-latest-send-log.json
@@ -66,43 +66,13 @@
           {
             "key": "emb.app_instance_id",
             "value": {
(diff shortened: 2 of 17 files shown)


src/resources/webSdkResource.test.ts
```diff
@@ -39,6 +39,7 @@ describe('webSdkResource', () => {
       'telemetry.sdk.language': 'webjs',
       'telemetry.sdk.name': 'embrace-web-sdk',
       'telemetry.sdk.version': SDK_VERSION,
+      'user_agent.original': window.navigator.userAgent,
```
joaquin-diaz, line 42: Small: why are we not importing from semconv here?

tests/integration/tests/__golden__/next-latest-send-log.json
```diff
@@ -66,43 +66,13 @@
...
-            "key": "browser.mobile",
-            "value": {
-              "boolValue": false
-            }
-          },
-          {
-            "key": "browser.language",
-            "value": {
-              "stringValue": "en-US"
+              "stringValue": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/139.0.7258.5 Safari/537.36"
```
joaquin-diaz, line 75: You may need to ignore this attribute as it will be different in CI. There's a list in https://github.com/embrace-io/embrace-web-sdk/blob/230a79dfd52cc2df468256b09c5a923d8781a680/tests/integration/utils/test-with-mock-api.ts#L50

pablomatiasgomez: yep, added! Thanks!

PR #585, opened 2025-07-15
chore: include screen_resolution resource attribute

pablomatiasgomez:


src/resources/webSdkResource.test.ts
```diff
@@ -40,6 +40,7 @@ describe('webSdkResource', () => {
       'telemetry.sdk.name': 'embrace-web-sdk',
       'telemetry.sdk.version': SDK_VERSION,
       'user_agent.original': window.navigator.userAgent,
+      screen_resolution: `${window.screen.width.toString()}x${window.screen.height.toString()}`,
```
pablomatiasgomez, line 43: @embrace-io/team-web do I have to check if window.screen is undefined? I tried doing that but the linter was complaining that it was not required. In theory this is widely available, but I wonder if this is undefined, that the entire delivery of data may fail?

joaquin-diaz: It should be always defined

overbalance (approved):

PR #587, opened 2025-07-16
EMBR-8245 chore: avoid ending limited sessions

jpmunz:
## Goal

Attempt to address cases where there are a quick succession of visibility changes resulting in sessions that don't contain useful telemetry. Adding a `limitedSessionMaxDurationMs` config to `SpanSessionVisibilityInstrumentation` so that sessions that have durations lower than this number aren't ended. The rule is bypassed if we detect a key or mouse down since in that case we assume that even though the session is short it contained at least some user interaction that we'd want to record.

This does have the possibility of dropping the last of these limited sessions for a given tab if the visibility change indicates the tab closing rather than just being hidden, defaulting the config to off here so that we can do further testing with real traffic

### DIFF
--- a/src/instrumentations/session/SpanSessionVisibilityInstrumentation/SpanSessionVisibilityInstrumentation.ts
+++ b/src/instrumentations/session/SpanSessionVisibilityInstrumentation/SpanSessionVisibilityInstrumentation.ts
@@ -7,7 +7,7 @@ import {
 } from '../../../utils/index.js';
 import type { TimeoutRef } from '../../../utils/index.js';
 
-const SESSION_INTERACTION_EVENTS = ['keydown', 'mousedown'];
+const SESSION_INTERACTION_EVENTS = ['mousedown'];
 
 export class SpanSessionVisibilityInstrumentation extends EmbraceInstrumentationBase {
   private _currentVisibilityState: DocumentVisibilityState;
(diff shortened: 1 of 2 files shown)


joaquin-diaz (approved):

src/instrumentations/session/SpanSessionVisibilityInstrumentation/SpanSessionVisibilityInstrumentation.ts
```diff
@@ -1,28 +1,44 @@
 import { EmbraceInstrumentationBase } from '../../EmbraceInstrumentationBase/index.js';
 import type { SpanSessionVisibilityInstrumentationArgs } from './types.js';
+import {
+  bulkAddEventListener,
+  bulkRemoveEventListener,
+  throttle,
+} from '../../../utils/index.js';
 import type { TimeoutRef } from '../../../utils/index.js';
 
+const SESSION_INTERACTION_EVENTS = ['mousedown'];
```
jpmunz, line 10: @joaquin-diaz FYI after your approval I made this change: https://github.com/embrace-io/embrace-web-sdk/pull/587/commits/874e687174bac5f4d9b2fae0e169594347e8c8ab

In testing I realized if you switch tabs quickly using the keyboard shortcuts that would be considered a user interaction and still send the limited sessions so I'm thinking that `mousedown` might be enough here

PR #589, opened 2025-07-16
chore: accept undefined logs

overbalance:
## Goal

Log null and undefined error logs.

## Testing

New tests added.


src/managers/EmbraceLogManager/EmbraceLogManager.ts
```diff
@@ -58,32 +58,47 @@ export class EmbraceLogManager implements LogManager {
...
       handled = true,
       attributes = {},
       timestamp = this._perf.getNowMillis(),
     }: LogExceptionOptions = {}
   ) {
+    if (!error) {
+      error = new Error('undefined logException');
+    }
+
+    // Validate attributes parameter
```
overbalance, line 72:
```suggestion
```

src/managers/EmbraceLogManager/EmbraceLogManager.ts
```diff
@@ -146,4 +167,55 @@ export class EmbraceLogManager implements LogManager {
...
+    }
+
+    // For non-Error types, generate a new stack trace
+    const userCallStack = new Error().stack || '';
+
+    if (typeof error === 'string') {
+      return {
+        message: error.trim(),
+        type: 'String',
+        name: '',
```
jpmunz, line 193: I think for this and the remaining cases we can have `name` use the same value that we have for `type` this is consistent for example with generic instances of `Error` where name and constructor.name are both "Error"

src/managers/EmbraceLogManager/EmbraceLogManager.ts
```diff
@@ -146,4 +167,55 @@ export class EmbraceLogManager implements LogManager {
...
+      return {
+        message: error.message.trim(),
+        type: error.constructor.name,
+        name: error.name,
+        stack: error.stack || '',
+      };
+    }
+
+    // For non-Error types, generate a new stack trace
+    const userCallStack = new Error().stack || '';
```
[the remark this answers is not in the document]
jpmunz: ya I wonder if we're better off in these cases to just have an empty stack, for the ones we generate from the global error handler the stack wouldn't be particularly useful, but I guess for ones the user calls through manually invoking `logException` it would show where they called it from

src/managers/EmbraceLogManager/EmbraceLogManager.ts
```diff
@@ -93,17 +109,22 @@ export class EmbraceLogManager implements LogManager {
...
 
     this._logMessage({
-      message,
+      message: message.trim(),
       severity,
       timestamp: this._perf.getNowMillis(),
       attributes,
       stackTrace:
-        includeStacktrace && severity != 'info' ? new Error().stack : undefined,
+        includeStacktrace && severity != 'info' ? new Error().stack : '',
```
jpmunz, line 127: curious if there's a particular reason to prefer empty string over undefined here?

src/managers/EmbraceLogManager/EmbraceLogManager.ts
```diff
@@ -58,32 +58,48 @@ export class EmbraceLogManager implements LogManager {
...
-    error: Error,
+    error: unknown,
     {
       handled = true,
       attributes = {},
       timestamp = this._perf.getNowMillis(),
     }: LogExceptionOptions = {}
   ) {
+    if (!error) {
+      error = new Error('undefined logException');
```
jpmunz, line 69: won't be much we can show in any case but maybe a slightly verbose explanation here like:
```suggestion
      error = new Error('logException received a falsy error');
```

src/managers/EmbraceLogManager/EmbraceLogManager.ts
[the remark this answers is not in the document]
overbalance: chatted on slack. we agreed to include for now and review the data in the next version.

PR #595, opened 2025-07-18
chore: add some missing types

overbalance:
## Goal

Added previously missing exports from sdk: DynamicConfigManager, DynamicSDKConfig, Span

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,10 +1,15 @@
 import { log } from './api-logs/index.js';
 import { session } from './api-sessions/index.js';
-import { trace } from './api-traces/index.js';
 import type { ExtendedSpan } from './api-traces/index.js';
+import { trace } from './api-traces/index.js';
 import { user } from './api-users/index.js';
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -1,10 +1,20 @@
...
+import type {
+  DynamicConfigManager,
+  DynamicSDKConfig,
+  Span,
+} from './sdk/index.js';
+import { DiagLogLevel, initSDK } from './sdk/index.js';
+
+const sdk = {
+  initSDK,
+  DiagLogLevel,
```
jpmunz, line 16: I would also include Span, DynamicConfigManager, DynamicSDKConfig to avoid this being a breaking change, lgtm otherwise though!

jpmunz (changes requested):

src/index.ts
overbalance: I'll revert this change and just add the missing types.

jpmunz (approved):

PR #606, opened 2025-07-21
chore: accept stacktraces from the user

overbalance:
## Goal

Allow users to pass in their own stacktraces. If `includeStacktrace` and `stacktrace` are both included, `stacktrace` takes precedence.

## Testing

Included tests pass.

### DIFF
--- a/src/managers/EmbraceLogManager/EmbraceLogManager.test.ts
+++ b/src/managers/EmbraceLogManager/EmbraceLogManager.test.ts
@@ -764,7 +764,7 @@ describe('EmbraceLogManager', () => {
   });
 
   describe('messages with stacktrace passed in', () => {
-    it('should log an info log with stacktrace passed in', () => {
+    it('should log an info log but ignore stacktrace when stacktrace passed in', () => {
       expect(() => {
         manager.message('this is an info log with a stacktrace', 'info', {
           stacktrace: 'i am stacktrace passed in by the user',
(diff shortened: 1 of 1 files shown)


src/managers/EmbraceLogManager/EmbraceLogManager.ts
```diff
@@ -123,13 +127,19 @@ export class EmbraceLogManager implements LogManager {
       this._spanSessionManager.incrSessionCountForKey(KEY_EMB_ERROR_LOG_COUNT);
     }
 
+    let stacktraceString = '';
+    if (stacktrace) {
+      stacktraceString = stacktrace;
+    } else if (includeStacktrace && severity !== 'info') {
+      stacktraceString = new Error().stack || '';
```
jpmunz, line 134: I would include the `severity !== 'info'` check for the passed in stack trace as well since on the dashboard side we don't display stack traces for info logs

src/managers/EmbraceLogManager/EmbraceLogManager.ts
```diff
@@ -177,7 +189,7 @@ export class EmbraceLogManager implements LogManager {
     message: string;
     type: string;
     name: string;
-    stack: string;
+    stack: string; // 'stack' here (not 'stacktrace') to match the standard Error.stack property name
```
overbalance, line 192:
```suggestion
    stack: string; // 'stack' not 'stacktrace' here to match the standard Error.stack property name
```

PR #629, opened 2025-07-28
chore: remove tsx and rimraf

overbalance:
Makes the package smaller by running these utilities on demand.


package.json
```diff
@@ -18,7 +18,7 @@
     "compile": "npm run sdk:compile && npm run compile --workspaces --if-present",
     "dev": "npm run sdk:compile:esm:watch & npm run sdk:compile:types:watch & npm run dev --workspace=embrace-web-sdk-react-demo",
     "demo": "npm run demo --workspace=embrace-web-sdk-react-demo",
-    "sdk:analyze": "tsx scripts/measureSize.ts",
+    "sdk:analyze": "npx tsx scripts/measureSize.ts",
     "sdk:docs": "typedoc --readme none && touch docs/.nojekyll",
     "sdk:compile:clean": "npx rimraf build .sonda",
```
jpmunz, line 23: since these commands still run `tsx` + `rimraf` is this now relying on those packages being globally installed in the dev environment?

overbalance: no, `npx` looks for the nearest copy of `tsx`. if it doesn't find one, it downloads it temporarily and runs it. for rimraf, I left it installed at the project root so it will be found most of the time. in envs like ci, if node_modules is empty, it will still run because of the temporary download I described.

PR #633, opened 2025-07-28
chore: clean up iife code

overbalance:
## Goal

Use a new entrypoint for the cdn bundle so rollup is no longer in charge of adding `EmbraceWebSdk` to the window.

## Testing

CDN demo works.


rollup.config.js
```diff
@@ -108,19 +108,12 @@ export default defineConfig([
...
+      file: 'build/iife/index.js',
       format: 'iife',
-      // global variable name for the SDK
-      name: 'EmbraceWebSdk',
       sourcemap: true,
-      // TODO create a new entry file that sets the window variable explicitly and remove this line.
-      // This is a workaround to assign the SDK to the global window object so users can lazyload the SDK.
-      // By default, rollup creates the export with 'var' and assumes the SDK is being used in a script tag,
-      // which would normally be the window.
-      footer: 'window.EmbraceWebSdk = EmbraceWebSdk;',
```
jpmunz, line 123: to clarify this was the same behaviour we're just now doing it ourselves in `src/iife.ts` correct? Is there a particular advantage or more that we want to be explicit about what's happening?

PR #636, opened 2025-07-31
chore: report sdk startup duration as `emb.sdk_startup_duration` attribute

pablomatiasgomez:
## Goal

- The attribute was wrong. `emb.startup_duration` should be used for app startup duration (we are currently not measuring this)
- The type was sometimes wrong, not sure why is being reported as double so I added a Math.ceil() just in case.


src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -306,7 +306,7 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
     };
   }
 
-  public recordStartupDuration(duration: number) {
-    this._startupDuration = duration;
+  public recordSDKStartupDuration(duration: number) {
+    this._sdkStartupDuration = Math.ceil(duration);
```
jpmunz, line 310: I had seen this but thought it was a browser difference:
https://github.com/embrace-io/embrace-web-sdk/pull/548/files

I think what you have is better for consistency though, you could probably remove that fallback in the test

jpmunz (approved):

overbalance (approved):

PR #639, opened 2025-08-01
chore: publish all markdown files

overbalance:
There is a 404 when you follow relative links like REACT.md on jsdelivr.
https://www.jsdelivr.com/package/npm/@embrace-io/web-sdk


jpmunz (approved):
ah nice catch

npmjs looks like it gets around this by appending our repo URL in front of relative links

PR #679, opened 2025-08-19
feat: temporarily save spans to local storage when not exporting

pablomatiasgomez:


joaquin-diaz (commented):
As we agreed on the standup, we should group up all the temporary "ugly" code in a couple of helper functions so once we either remove or improve this we don't need to worry about changing the rest of the code

src/instrumentations/session/SpanSessionVisibilityInstrumentation/SpanSessionVisibilityInstrumentation.ts
```diff
@@ -1,31 +1,37 @@
...
+  public constructor(
+    {
+      diag,
+      perf,
+      visibilityWaitTimeMs = 0,
+      limitedSessionMaxDurationMs = 0,
+      backgroundSessions = false,
+      visibilityDoc = window.document,
+    }: SpanSessionVisibilityInstrumentationArgs = {},
+    embraceSpanProcessor?: EmbraceSessionBatchedSpanProcessor
```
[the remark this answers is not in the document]
pablomatiasgomez: My felling was that SpanSessionVisibilityInstrumentationArgs was more for configuring things.
We are also calling this by using `config['session-visibility']` so its config that the user provides

src/api-sessions/api/SessionAPI/SessionAPI.ts
```diff
@@ -58,6 +59,12 @@ export class SessionAPI implements SpanSessionManager {
     this.getSpanSessionManager().endSessionSpanInternal(reason);
   }
 
+  public endSessionSpanWithoutExporting(
```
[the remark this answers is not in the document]
joaquin-diaz: Ah, ideally your method and that one should be in `SpanSessionManagerInternal` instead of `SpanSessionManager`. All our internal stuff use `EmbraceSpanSessionManager` which implements `SpanSessionManagerInternal` and since it's called directly instead of doing `import { session } ..` it works fine

src/instrumentations/session/SpanSessionVisibilityInstrumentation/SpanSessionVisibilityInstrumentation.ts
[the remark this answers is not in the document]
joaquin-diaz: Gotcha

joaquin-diaz (approved):

src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -233,6 +228,45 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
+        startTime: this._activeSessionStartTime,
+        attributes: {
+          // Copy all current attributes from the original session span, plus the ending attributes
+          ...this._sessionSpan.attributes,
+          ...this._endSessionSpanAttributes(reason),
+          [KEY_EMB_FROM_STORAGE]: true,
+        },
+      })
+    );
+    return spanCopy.endWithoutExporting();
```
jpmunz, line 267: wondering if we can avoid the extra logic to alter the internals in `endWithoutExporting`, ultimately I guess the line we're trying to avoid is: https://github.com/open-telemetry/opentelemetry-js/blob/806e56b38845ac21e826cfd08c2f39c48bb681d2/packages/opentelemetry-sdk-trace-base/src/Span.ts#L294

The `this._spanProcessor` there is coming from the tracer so could instantiate a separate one like `this._nonProcessoringTracer = new Tracer(...)` to have available to the manager and use it to start+end the span copy here

src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -233,6 +228,45 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
+      [KEY_EMB_SESSION_REASON_ENDED]: reason,
+      ...this._activeSessionCounts,
+      ...this._limitManager.getDiagnosticCounts(),
+      [KEY_EMB_SDK_STARTUP_DURATION]: this._sdkStartupDuration,
+    };
+  }
+
+  // endSessionSpanWithoutExporting creates a copy of the current session span with the same attributes
+  // that endSessionSpanInternal would add, but does not affect the original session span which remains active.
+  public endSessionSpanWithoutExporting(
```
jpmunz, line 243: the name might be a bit misleading if it implies the current session span is being ended, what about maybe `currentSessionSpanReadonlyCopy` or `currentSessionAsReadableSpan`?

src/utils/EmbraceSpanStorage/EmbraceSpanStorage.ts
```diff
@@ -0,0 +1,149 @@
...
+    try {
+      // If this session was already stored, clear it first:
+      this.clearStoredSpans(sessionId);
+
+      const key = `${PENDING_SPANS_STORAGE_KEY_PREFIX}${sessionId}_${Date.now()}`;
+      this._storage.setItem(
+        key,
+        JSON.stringify([sessionSpan, ...pendingSpans], (key, value) =>
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+          key.startsWith('_') ? undefined : value
```
jpmunz, line 48: why were these being cleared?

src/utils/EmbraceSpanStorage/EmbraceSpanStorage.ts
```diff
@@ -0,0 +1,149 @@
...
+        }
+      }
+      const currentTime = Date.now();
+
+      keys.forEach(key => {
+        const parts = key.split('_');
+        const storedTime = parseInt(parts[parts.length - 1], 10);
+
+        if (
+          isNaN(storedTime) ||
```
jpmunz, line 106: should we also clear the ones where the time is invalid?

jpmunz, line 106: should we also clear the ones where the time is invalid?

src/utils/EmbraceSpanStorage/EmbraceSpanStorage.ts
```diff
@@ -0,0 +1,149 @@
...
+          currentTime - storedTime <= STORED_SPANS_EXPIRE_TIMEOUT_MS
+        ) {
+          return;
+        }
+
+        const storedData = this._storage.getItem(key);
+        if (!storedData) return;
+
+        try {
```
jpmunz, line 115: not sure how much complexity this would otherwise add but since we don't need to work with these as spans any longer we could consider moving what we store down to the transport level, e.g. storing this Uint8Array of the payload: https://github.com/embrace-io/embrace-web-sdk/blob/7276b54b2ca914fbd80bd863457ca06721eaeb1a/src/transport/FetchTransport/FetchTransport.ts#L48

which would avoid having to convert the JSON data back to spans, we could discuss further in standup

src/utils/EmbraceSpanStorage/EmbraceSpanStorage.ts
```diff
@@ -0,0 +1,149 @@
...
+      });
+    } catch (error) {
+      this._diag.error('Failed to clear stored spans from storage:', error);
+    }
+  }
+
+  public startExpiredSpansCheck(): void {
+    this._checkExpiredSpansInterval = setInterval(() => {
+      this.checkAndExportExpiredSpans();
+    }, 60 * 1000); // Check every minute
```
jpmunz, line 80: would be good to make this and STORED_SPANS_EXPIRE_TIMEOUT_MS configurable

src/instrumentations/session/SpanSessionVisibilityInstrumentation/SpanSessionVisibilityInstrumentation.ts
```diff
@@ -81,14 +88,20 @@ export class SpanSessionVisibilityInstrumentation extends EmbraceInstrumentation
       );
 
       const currentSessionStartTime = this.sessionManager.getSessionStartTime();
+      const pendingSpansCount =
+        this._embraceSpanProcessor?.getPendingSpansCount() || 0;
```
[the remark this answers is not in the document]
pablomatiasgomez: In line 117 I'm checking if we have the processor to use the storage or not, so here I think it's fine (if no processor, there are no pending spans)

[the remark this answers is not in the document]
pablomatiasgomez: Ah actually I think I get your point now. You would want to fully avoid doing the limited session logic if not using Embrace? I guess it makes sense, yeah

src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
pablomatiasgomez: But once you have a Span, how do you change it's _spanProcessor?

pablomatiasgomez: ACtually NoopTracer is not exported :( and neither is `new Tracer()` 😢

jpmunz: ah that's annoying, ya I was thinking to do `new Tracer`, looks like you'd have to go all the way up to creating a separate TracerProvider: https://github.com/open-telemetry/opentelemetry-js/blob/96a2ecf1d7cc7a8cbda436436347702057587a0c/packages/opentelemetry-sdk-trace-base/src/BasicTracerProvider.ts#L37

src/utils/EmbraceSpanStorage/EmbraceSpanStorage.ts
pablomatiasgomez: Because otherwise the internal objects of the Span object (like _spanProcessor) end up in a recursive call which makes JSON.stringify fail.. Honestly I don't think any of these spans where meant to be serialized as they are. Thats why the lib internal creates a different object when they are exported, but all of that is not exposed

src/utils/EmbraceSpanStorage/EmbraceSpanStorage.ts
pablomatiasgomez: yeah good point! changed

src/utils/EmbraceSpanStorage/EmbraceSpanStorage.ts
pablomatiasgomez: The thing is that we are using this to export spans: https://github.com/embrace-io/embrace-web-sdk/blob/d044928cdbdbd86dc4243fd984bc8c69b5c59646/src/processors/EmbraceSessionBatchedSpanProcessor/EmbraceSessionBatchedSpanProcessor.ts#L77

which uses `internal._export`  which expects an Array of ReadableSpan. I guess that if we save the byte array directly we would need to reimplement how the exporter does everything? (gzip, headers, retries?)

jpmunz: ya I think worth a discussion there to see how feasible it is, I was thinking less that we would change how the exporter is doing things but more that it is able to pass some sort of flag down through to the transport level so that the gzipping etc. all happens as normal but instead of a network request at the end it's a save to local storage

src/utils/EmbraceSpanStorage/EmbraceSpanStorage.ts
```diff
@@ -0,0 +1,189 @@
...
+const MAX_PENDING_SPANS_ITEMS = 10;
+
+export interface SpanStorageOptions {
+  storage?: Storage;
+  diag?: DiagLogger;
+  storedSpansExpireTimeoutMS?: number;
+  onExpiredSpansExport?: (spans: ReadableSpan[]) => void;
+}
+
+// EmbraceSpanStorage is used to keep pending spans that are not meant to be exported (yet) in localStorage,
```
jpmunz, line 16: 👍 I filed this as a follow-up

src/utils/EmbraceSpanStorage/EmbraceSpanStorage.ts
```diff
@@ -0,0 +1,189 @@
...
+      });
+    } catch (error) {
+      this._diag.error('Failed to clear stored spans from storage:', error);
+    }
+  }
+
+  public startExpiredSpansCheck(): void {
+    this._checkExpiredSpansInterval = setInterval(() => {
+      this.checkAndExportExpiredSpans();
+    }, 60 * 1000); // Check every minute
```
jpmunz, line 109: minor since it is lightweight, given the expiry I think this could be even every 5 or 10 minutes

src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -233,6 +229,47 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
+    reason: ReasonSessionEnded
+  ): ReadableSpan | null {
+    if (!this._sessionSpan || !this._activeSessionStartTime) {
+      this._diag.debug(
+        'trying to end a session, but there is no session in progress. This is a no-op.'
+      );
+      return null;
+    }
+
+    const tracer = new BasicTracerProvider().getTracer(
```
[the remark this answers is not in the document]
pablomatiasgomez: done

src/instrumentations/session/SpanSessionVisibilityInstrumentation/SpanSessionVisibilityInstrumentation.ts
```diff
@@ -1,31 +1,37 @@
...
 import {
   bulkAddEventListener,
   bulkRemoveEventListener,
   throttle,
 } from '../../../utils/index.js';
-import type { TimeoutRef } from '../../../utils/index.js';
+import type { EmbraceProcessor } from '../../../processors/index.js';
 
 const SESSION_INTERACTION_EVENTS = ['mousedown'];
+const MAX_PENDING_SPAN_COUNT = 5;
```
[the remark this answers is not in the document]
pablomatiasgomez: done!

src/utils/EmbraceSpanStorage/EmbraceSpanStorage.ts
```diff
@@ -0,0 +1,189 @@
...
+      );
+    } catch (error) {
+      this._diag.error('Failed to store spans to storage:', error);
+    }
+  }
+
+  public clearStoredSpans(sessionId: string): void {
+    try {
+      const keysToRemove: string[] = [];
+      for (let i = 0; i < this._storage.length; i++) {
```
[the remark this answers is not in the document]
pablomatiasgomez: done!

PR #684, opened 2025-08-20
feat: support running multiple instances of the SDK

joaquin-diaz:
## Goal

Allow users to create multiple instances of the Embrace SDK at the same time without registering global providers. This allows each instance to emit its own telemetry. 

Some things are still set up globally and need to be worked separately:
* Fetch instrumentation: current code is patching fetch globally so last SDK to register will only emit network spans
* Global exceptions: is catching all exceptions so each instance of the SDK will capture all the errors 
* Storage: local storage needs to be reworked so its keys are prefixed with app id to avoid conflicts

## Testing

Unit tests and created another demo app that runs this scenario


demo/frontend-sdk/sdk/bundle.js
overbalance, line 1: drop this file?

demo/frontend-sdk/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "frontend-sdk",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "setup": "cd sdk && npm install && cd ../host && npm install && cd .. && bash scripts/setupBundleSymLinks.sh",
```
overbalance, line 7: Workspaces can be installed from the top

```suggestion
    "setup": "npm install && bash scripts/setupBundleSymLinks.sh",
```

demo/frontend-sdk/package.json
```diff
@@ -0,0 +1,20 @@
...
+    "build": "cd sdk && npm run build && cd ../host && npm run build",
```
overbalance, line 8:
```suggestion
    "build": "npm run build --workspaces",
```

overbalance (approved):

demo/frontend-sdk/sdk/bundle.js
joaquin-diaz: Yep!

PR #685, opened 2025-08-21
feat(experience): add user experience tab tracking

overbalance:
## Summary
Introduces **EmbraceExperienceManager** to track user journeys across browser tabs by assigning persistent experience IDs that group related sessions.

## Key Features
- **Experience ID**: Persistent identifier shared across related tabs
- **Tab Detection**: Identifies how tabs are opened (link, manual, window.open, reload, back/forward)  
- **Cross-Tab Inheritance**: Child tabs inherit parent's experience ID via localStorage
- **Referrer Tracking**: Captures referrer type, domain (external), or path (same-origin)

## New Attributes
```typescript
emb.experience_id      // Groups related sessions
emb.tab_open_method    // How tab was opened
emb.referrer_type      // same_origin/external/none
emb.referrer_path      // For same-origin navigation
emb.referrer_domain    // For external referrers
emb.previous_tab_id    // Parent tab ID if inherited
```

## API Additions
```typescript
sessionManager.getExperienceId(): string | null
sessionManager.getPreviousTabId(): string | null
```

## Technical Details
- Dual-key storage system for accurate parent-child matching
- Auto-cleanup of data older than 24 hours
- Comprehensive test coverage (779 lines, all passing)
- Performance optimized with pre-computed referrer info

## Testing
✅ All tests passing - covers detection, inheritance, storage quota handling, and edge cases


src/resources/experienceId.ts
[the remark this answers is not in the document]
overbalance: I'm going to refactor and we can review together.

src/resources/experienceId.ts
```diff
@@ -0,0 +1,111 @@
+import { generateUUID } from '../utils/index.js';
+
+const EXPERIENCE_ID_KEY = 'embrace_experience_id';
```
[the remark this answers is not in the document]
overbalance: I'm going to refactor and we can review together.

src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -208,6 +212,7 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
 
     this._sessionSpan.setAttributes({
       ...this._getPermanentAttributes(),
+      ...this._experienceManager.getExperienceAttributes(),
```
overbalance, line 215: Calling this twice: once at startSession and again at endSession. Only because you may want to read the attributes in the middle of a session. The second call is needed to update the lastActivity update.

src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -257,6 +266,7 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
 
     const attributes: Attributes = {
       ...this._getPermanentAttributes(),
+      ...this._experienceManager.getExperienceAttributes(),
```
overbalance, line 269: Here is the second call.

src/managers/EmbraceSpanSessionManager/types.ts
```diff
@@ -3,13 +3,15 @@ import type { PerformanceManager } from '../../utils/index.js';
...
 import type { LimitManagerInternal } from '../EmbraceLimitManager/index.js';
+import type { EmbraceExperienceManager } from '../EmbraceExperienceManager/index.js';
 
 export interface EmbraceSpanSessionManagerArgs {
   diag?: DiagLogger;
   perf?: PerformanceManager;
   visibilityDoc?: VisibilityStateDocument;
   storage?: Storage;
   limitManager: LimitManagerInternal;
+  experienceManager?: EmbraceExperienceManager;
```
overbalance, line 14: Set as optional because I only foresee this used in testing.

src/managers/EmbraceExperienceManager/EmbraceExperienceManager.ts
```diff
@@ -0,0 +1,576 @@
...
+    try {
+      // Store with tab-specific key
+      this._storage.setItem(
+        `${INHERITANCE_KEY_PREFIX}${this._appInstanceId}`,
+        experienceDataStr
+      );
+
+      // Store with referrer-based key for accurate child matching
+      const referrerKey = EmbraceExperienceManager._createReferrerKey(
+        window.location.href
```
joaquin-diaz, line 338: Should we also provide a location object via constructor as we do in other places?

src/managers/EmbraceExperienceManager/EmbraceExperienceManager.ts
```diff
@@ -0,0 +1,576 @@
...
+      // Ignore URL parsing errors
+    }
+  }
+
+  /**
+   * Update the last activity timestamp
+   * Called on session start/end events
+   */
+  public updateLastActivity(): void {
+    this._experienceData.lastActivityAt = Date.now();
```
joaquin-diaz, line 542: If lastActivityAt is only based on session, maybe we should rename it to something else. Activity can be confused as user activity like moving the cursor or clicking, also if it only depends on session activity maybe we can use the session lifecycle listeners? we have `addSessionEndedListener` and `addSessionStartedListener`

src/managers/EmbraceExperienceManager/EmbraceExperienceManager.ts
```diff
@@ -0,0 +1,576 @@
...
+    this._storeExperienceData();
+  }
+
+  /**
+   * Get all experience attributes for session spans
+   * Updates last activity timestamp on each call
+   */
+  public getExperienceAttributes(): Record<string, AttributeValue> {
+    // Update last activity timestamp
+    this.updateLastActivity();
```
joaquin-diaz, line 552: It feels weird that this gets updated when you get the attributes, this is only done this way because we know we call it on session start / end?

src/managers/EmbraceExperienceManager/EmbraceExperienceManager.test.ts
```diff
@@ -0,0 +1,789 @@
...
+      expect(manager.getTabOpenMethod()).to.equal('manual_new_tab');
+    });
+  });
+
+  describe('inheritance', () => {
+    it('should inherit experience ID from parent tab via window.opener', () => {
+      // First ensure window.opener is null (from beforeEach)
+
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      expect(window.opener).to.be.null;
```
joaquin-diaz, line 215: I think you can avoid all the `eslint-disable-next-line @typescript-eslint/no-unused-expressions` by doing `void expect(window.opener).to.be.null;`

src/api-sessions/api/SessionAPI/SessionAPI.ts
```diff
@@ -62,6 +63,10 @@ export class SessionAPI implements SpanSessionManager {
     return this.getSpanSessionManager().getSessionId();
   }
 
+  public getExperienceManager(): EmbraceExperienceManager | null {
+    return this.getSpanSessionManager().getExperienceManager();
```
joaquin-diaz, line 67: I'm wondering if the experience should be it's own global API, given that sessions are current-tab related and experiences is cross tab

joaquin-diaz (commented):
A few questions around activity but the rest looks amazing!

src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
overbalance: my mistake. I did this when I decided to report last activity time, but decided to only use it internally to determine which experience. There *are* no updates to the values within a session. I'll remove.

src/managers/EmbraceExperienceManager/EmbraceExperienceManager.ts
overbalance: My idea is that updates would come from multiple places in the code depending on what we consider activity. This happens to be the only one so far. I think it looks weird because we aren't tracking other user interactions yet.

src/managers/EmbraceExperienceManager/EmbraceExperienceManager.ts
overbalance: My comment missed your point @joaquin-diaz, I was responding to @jpmunz.

I did remove the attribute from the data, but last activity is used to calculate the experience of the next tab. Removing it wouldn't make sense as this is architected currently. It's a combination of referrer and most recent using a best-candidate algo.

PR #707, opened 2025-08-26
EMBR-8477 chore: 2.0 release prep

jpmunz:
## Goal

Additional breaking changes to include as part of 2.0.0 along with documentation for how to upgrade

### DIFF
--- a/tests/integration/platforms/webpack-4/src/index.js
+++ b/tests/integration/platforms/webpack-4/src/index.js
@@ -1,4 +1,4 @@
-const { sdk } = require('@embrace-io/web-sdk');
+const { initSDK, DiagLogLevel } = require('@embrace-io/web-sdk');
 const { ZoneContextManager } = require('@opentelemetry/context-zone');
 const { registerInstrumentations } = require('@opentelemetry/instrumentation');
 const {
@@ -20,10 +20,10 @@ registerInstrumentations({
   instrumentations: [new DocumentLoadInstrumentation()],
 });
 
-sdk.initSDK({
+initSDK({
   appID: '',
   appVersion: 'YOUR_APP_VERSION',
-  logLevel: sdk.DiagLogLevel.INFO,
+  logLevel: DiagLogLevel.INFO,
   spanExporters: [new ConsoleSpanExporter()],
   logExporters: [new ConsoleLogRecordExporter()],
 });


README.md
```diff
@@ -464,15 +464,20 @@ We also provide a CDN version that is transpiled down to ES6/ES2015 for maximum
...
 The SDK is built on top of OpenTelemetry and as such it is possible to use it alongside other OTel libraries. **Important:** The Embrace Web SDK only supports OpenTelemetry 1.x packages. OpenTelemetry 2.x (and above) is **not supported** and will not work with this SDK at this time.
 
 If you wish to customize the SDK behavior by configuring custom resources, exporters, processors, or instrumentations, you must ensure that you are using versions of the OTel packages that are compatible with our SDK:
 
-| Open Telemetry APIs | Core    | Instrumentations & Contrib |
-|---------------------|---------|----------------------------|
-| 1.9.0               | 1.30.1  | 0.57.2                     |
+| Embrace Web SDK | Open Telemetry APIs | Core   | Instrumentations & Contrib |
+|-----------------|---------------------|--------|----------------------------|
+| 2.x             | 1.9.x               | 2.0.x  | 0.203.x                    |
```
jpmunz, line 479: are we fine with the "x" patch versions here or does the compatibility need to be more restrictive?

src/index.ts
```diff
@@ -4,12 +4,15 @@ import type { ExtendedSpan } from './api-traces/index.js';
...
-  DynamicSDKConfig,
-  Span,
-} from './sdk/index.js';
-import * as sdk from './sdk/index.js';
+import type { DynamicConfigManager, DynamicSDKConfig } from './sdk/index.js';
+import type { Span } from '@opentelemetry/api';
+import { DiagLogLevel } from '@opentelemetry/api';
+import { initSDK } from './sdk/index.js';
+
+const sdk = {
```
jpmunz, line 12: @overbalance was this enough to get the typing on `sdk`? I wasn't sure how to test seeing that

src/index.ts
```diff
@@ -4,12 +4,15 @@ import type { ExtendedSpan } from './api-traces/index.js';
...
+  initSDK,
+  DiagLogLevel,
+};
 
 export { getNavigationInstrumentation, log, sdk, session, trace, user };
```
jpmunz, line 17: @overbalance what were the reasons for including the export as `sdk.initSDK` vs `initSDK`?

README.md
overbalance: the last one is technically `0.20x.0` which is strange because they only bump the minor and never the patch. What about using `^` for the others and a range for the last one?
```suggestion
| ^2.0.3          | ^1.9.0              | ^2.0.3 | >=0.203.0 <= 0.209.0      |
```

UPGRADING.md
```diff
@@ -0,0 +1,91 @@
...
+import { log } from '@embrace-io/web-sdk';
+
+log.message('Loading not finished in time.', 'error', {
+  keyA: 'valueA',
+  keYB: 'valueB'
+}, true);
+
+// ... 
+
+log.logException(err, true, { keyA: 'valueA' }, ts);
```
overbalance, line 68: `ts`?

overbalance:
What do you think about renaming `log.logException` to `log.exception`? We could keep backcompat by keeping the old name but encouraging the new one?

README.md
jpmunz: updated though I'm not sure I follow the point about `>=0.203.0 <=0.209.0`, why would we not support a 0.210.0 version in that case? Or did you mean `>=0.203.0 <=0.299.0` since they would reserve bumping to 0.300.0 if they ever did a new major core version bump? I set as `~0.203.0` for now

src/index.ts
jpmunz: removed `sdk` and switched the rest to top level exports

README.md
overbalance: If we set it to `~0.203.0` then we are restricting it to patches, which means it is effectively equivalent to pinning it at `0.203.0` because experimental libs don't have patches.

Consider the core goes to `2.0.10`. Would they set experimental to `0.210.0` or `0.2010.0`? I'll ask them on Slack what the plan is because I'm really not sure.

pablomatiasgomez (approved):

PR #713, opened 2025-08-28
refactor: use rolldown instead of rollup

overbalance:
## Goal

Speed up the build with less packages. Because Rolldown does esm/cjs interop and node resolution without plugins, it's a better version of Rollup. It's still in beta but it works perfectly as a drop-in replacement today.

https://rolldown.rs/guide/ 

Roadmap:
https://github.com/rolldown/rolldown/discussions/153


package.json
```diff
@@ -164,7 +162,7 @@
     "playwright": "1.55.0",
     "prettier": "3.6.2",
     "rimraf": "6.0.1",
-    "rollup": "4.49.0",
+    "rolldown": "^1.0.0-beta.34",
```
jpmunz, line 165: I'm hesitant to include this while still in beta, do they have a list of what would need to be updated before they would consider it stable?

overbalance:
Closing in favor of https://github.com/embrace-io/embrace-web-sdk/pull/805

PR #714, opened 2025-08-28
feat: add browser and default fields to exports

overbalance:
## Goal

- Add browser field and export condition for better bundler compatibility
- Add engines field requiring Node.js 18+
- Add default export condition as fallback
- Move React instrumentation from src/react to src/react-instrumentation
- Update build to output CJS files to build/cjs/ instead of build/src/
- Remove esnext build target to simplify build pipeline
- Update measureSize script to work with new build structure

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,9 @@
   "description": "Embrace Web SDK",
   "type": "module",
   "jsdelivr": "build/iife/bundle.js",
-  "main": "build/src/index.js",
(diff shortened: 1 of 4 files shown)


overbalance:
@jpmunz after more research we should keep cjs for webpack 4. i'll make a new branch with optimizations.

overbalance:
Closing to recreate with correct branch name feat/browser-export-fields

PR #715, opened 2025-08-28
EMBR-7318 feat: rewrite network instrumentation

jpmunz:
Putting this up for reference but doesn't seem like we can move forward with this as replacement for the existing network instrumentations:
* HTTP request method used is unavailable from PerformanceResourceTiming
* Request size is unavailable (current instrumentation doesn't report this by default but can be enabled)
* In the case of errors response status text is unavailable
* Status code is currently unavailable in Safari (https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/responseStatus)
* Response size is unavailable if the request is CORS and `Timing-Allow-Origin` is not present

### DIFF
--- a/demo/frontend/src/App.tsx
+++ b/demo/frontend/src/App.tsx
@@ -136,6 +136,12 @@ const App = () => {
     req.send();
   };
 
+  const handleSendXMLNetworkRequest404 = () => {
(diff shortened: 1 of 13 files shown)


jpmunz:
closing as won't do

PR #739, opened 2025-09-03
EMBR-8855 feat: config to avoid repatching global in fetch instrumentation

jpmunz:
In the case that the SDK is included twice in both a host application and a library want to give a config options to the network instrumentations to choose not to patch fetch / xhr if they find it already has been patched.

This just has the `fetch` update, putting this up for review now since the approach for xhr will be very similar and will work on that piece next


demo/frontend-sdk/shared/otel.ts
```diff
@@ -62,11 +70,22 @@ const addEmbraceSDK = () => {
...
 
-const initSDK = (appID: string) => {
+const initSDK = (
+  appID: string,
+  appVersion: string,
+  omitNetworkIfAlreadyPatched: boolean
+) => {
   window.EmbraceWebSdkOnReady.onReady(() => {
     const { initSDK } = window.EmbraceWebSdk;
+    console.log(
```
joaquin-diaz, line 80: Clean up?

joaquin-diaz (approved):

demo/frontend-sdk/shared/otel.ts
jpmunz: I was about to but actually since I found it useful for debugging and it's just in the demo app maybe worth keeping?

jpmunz:
@joaquin-diaz fyi you already approved but I just added a 2nd commit for the equivalent changes for XHR: https://github.com/embrace-io/embrace-web-sdk/pull/739/commits/7c5bc68ad644087593bfb66307a67c9dd8ea63bf

PR #779, opened 2025-09-09
chore: update README with privacy faq and network instrumentation config for multiple instances

jpmunz:


src/sdk/types.ts
```diff
@@ -324,11 +324,11 @@ export interface DefaultInstrumentationConfig {
     config has enabled=false. Instead, use `omit` to specify which default instrumentations should be turned off.
    */
   '@opentelemetry/instrumentation-fetch'?: Omit<
-    FetchInstrumentationConfig,
+    EmbraceFetchInstrumentationArgs,
```
jpmunz, line 327: I realized these types were wrong when writing the docs, `EmbraceFetchInstrumentationArgs` + `EmbraceXHRInstrumentationArgs` is the same except it adds in the optional `omitIfAlreadyPatched`

PR #781, opened 2025-09-10
EMBR-6945 feat: support network span forwarding

jpmunz:
## Goal

Support network span forwarding, when enabled this feature does two things:
1. Injects a 'traceparent' header to all outgoing network requests (see the existing [apple SDK implementation](https://github.com/embrace-io/embrace-apple-sdk/blob/main/Sources/EmbraceCore/Capture/Network/URLSessionTaskHandler.swift#L267) as an example)
2. Adds a 'emb.w3c_traceparent' attribute to the network span using the same value as the header, this allows us to forward the spans to other data destinations on the backend (see the existing [apple SDK implementation](https://github.com/embrace-io/embrace-apple-sdk/blob/main/Sources/EmbraceCore/Capture/Network/URLSessionTaskHandler.swift#L145) as an example)

In this implementation we can leverage the existing [W3CTraceContextPropagator](https://github.com/open-telemetry/opentelemetry-js/blob/2197129a1d73fb3c465527ac98156f7de2ecfc0c/packages/opentelemetry-core/src/trace/W3CTraceContextPropagator.ts#L74) to do 1). Both the [xhr](https://github.com/open-telemetry/opentelemetry-js/blob/2197129a1d73fb3c465527ac98156f7de2ecfc0c/experimental/packages/opentelemetry-instrumentation-xml-http-request/src/xhr.ts#L148) and [fetch](https://github.com/open-telemetry/opentelemetry-js/blob/2197129a1d73fb3c465527ac98156f7de2ecfc0c/experimental/packages/opentelemetry-instrumentation-fetch/src/fetch.ts#L208) instrumentations already call the globally registered propagator to inject additional headers.

For 2) added `EmbraceW3CTraceContextPropagator` which inherits from the base propagator and just does the extra step of adding the 'emb.w3c_traceparent'  span attribute

The other piece here is deciding if the feature is enabled, this is true when:
* We get a non-zero enabled % from remote config for the network span forwarding feature and the current device is considered part of that enabled percent
* The feature is not explicitly blocked with `blockNetworkSpanForwarding: true`
* There isn't some other config set that would interfere with the feature, e.g. passing a custom `propagator` isn't supported since we specifically need our `EmbraceW3CTraceContextPropagator` registered for it to work


src/sdk/initSDK.test.ts
```diff
@@ -1268,6 +1287,228 @@ describe('initSDK', () => {
...
+    };
+
+    type NSFTest = {
+      name: string;
+      sdkConfig: SDKInitConfig;
+      networkType: 'fetch' | 'xhr';
+      expectInjection: boolean;
+    };
+
+    const tests: NSFTest[] = [
```
jpmunz, line 1335: we normally just rewrite test cases even if there's duplicated code, in this case I found it extra verbose since I was repeating each case for both fetch and XHR but I can go back to that approach if this seems less readable

overbalance (approved):
what do you think about inverting the name since it's off by default? `blockNetworkSpanForwarding` -> `forwardNetworkSpans`

PR #783, opened 2025-09-11
chore: avoid upgrading webpack 4 dependencies

joaquin-diaz:


joaquin-diaz:
fixed in https://github.com/embrace-io/embrace-web-sdk/pull/810

PR #789, opened 2025-09-14
refactor(cli): faster tools, lower node req to 18

overbalance:
## Goal

I was looking for opportunities to make our bundle smaller and it turned it into a full review of the cli package. There are no functional changes, just housekeeping.

  - Switched from TypeScript compiler to tsup bundler for faster builds
  - Reduced Node.js requirement from 20.0.0 to 18.0.0 for broader compatibility
  - Removed unnecessary configuration files (.gitignore, .prettierrc.json)
  - Streamlined package.json scripts and removed redundant exports
  - Added es-check and publint for package validation
  - Updated tsconfig to ES2022 target with noEmit for type checking only
  - Reduced dev dependencies from 10 to 4 packages

## Testing
I ran the demo with an appID but I would like someone to do a prod upload to fully test.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,41 +1,28 @@
 {
   "name": "@embrace-io/web-cli",
   "bin": {
-    "embrace-web-cli": "build/index.js"
+    "embrace-web-cli": "dist/index.js"
   },
   "version": "2.2.0",
-  "description": "Embrace Web CLI to help setup the Embrace SDK in your web app",
+  "description": "Embrace Web CLI for uploading sourcemaps and versioning bundles",
   "type": "module",
-  "module": "build/index.js",
-  "esnext": "build/index.js",
-  "types": "build/index.d.ts",
   "publishConfig": {
     "access": "public"
   },
   "scripts": {
-    "clean": "npm run cli:install:clean && npm run cli:compile:clean",
-    "compile": "npm run cli:compile:clean && npm run cli:compile",
-    "prepublishOnly": "npm run cli:compile:clean & npm run cli:compile",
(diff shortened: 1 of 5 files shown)


cli/package.json
```diff
@@ -59,19 +46,12 @@
...
-    "eslint-plugin-prettier": "5.5.4",
-    "eslint-plugin-require-extensions": "0.1.3",
-    "globals": "16.3.0",
-    "prettier": "3.6.2",
-    "typescript": "5.8.3",
-    "typescript-eslint": "8.43.0"
+    "@types/node": "^18.0.0",
+    "es-check": "9.3.1",
+    "publint": "0.3.12",
+    "tsup": "8.5.0"
```
jpmunz, line 52: I think we should avoid this one, repo warns that it is no longer maintained: https://github.com/egoist/tsup

jpmunz (commented):
Ideally would want to keep the tooling consistent between this and the SDK, for the changes here are any relevant to also make at the top-level? Specifically:
* using es-check + publint
* specifying `engines.node`
* different values for `module` + `moduleResolution` here vs ./tsconfig.json

cli/package.json
jpmunz: they suggest `tsdown` as an alternative

overbalance:
> Ideally would want to keep the tooling consistent between this and the SDK, for the changes here are any relevant to also make at the top-level? Specifically:
> 
> * using es-check + publint
> * specifying `engines.node`
> * different values for `module` + `moduleResolution` here vs ./tsconfig.json

Absolutely but I wanted to keep the PR scope specific.

jpmunz (approved):

PR #805, opened 2025-09-17
refactor(build): replace rollup with esbuild+terser

overbalance:
## Summary
- Switch from rollup to esbuild for bundling (faster builds)
- Add terser for minification to maintain bundle size (~46KB gzipped)  
- Optimize Vite config to exclude Embrace SDK from dependency pre-bundling
- Simplify build configuration

## Details
This PR replaces our rollup build system with esbuild + terser. The main benefits are:
- Much faster build times (1.2s → 74ms)
- Simpler configuration
- Same output size as before
- Better Vite integration - SDK is now excluded from pre-bundling to avoid double-processing

## Test plan
- [x] Build runs successfully (`npm run compile`)
- [x] Bundle size remains the same (~46KB gzipped)
- [x] Sourcemaps work correctly
- [x] Vite demo works without pre-bundling issues
- [x] All tests pass


demo/frontend/vite.config.ts
```diff
@@ -1,9 +1,19 @@
-import { defineConfig } from 'vite';
+import { resolve } from 'node:path';
 import Sonda from 'sonda/vite';
+import { defineConfig } from 'vite';
 
 // https://vite.dev/config/
 export default defineConfig({
   plugins: [Sonda({ enabled: false })],
+  resolve: {
```
overbalance, line 8: aliasing means we dont have to put the sdk in watch mode which is way easier and less resource-intensive.

tsconfig.json
```diff
@@ -3,31 +3,32 @@
   "compilerOptions": {
     "allowUnreachableCode": false,
     "allowUnusedLabels": false,
+    "declaration": true,
+    "declarationMap": true,
+    // "erasableSyntaxOnly": true,
```
overbalance, line 8: I'd like to implement this soon. The only rule we violate is the presence of enums.

package.json
```diff
@@ -3,89 +3,60 @@
   "version": "2.2.0",
   "description": "Embrace Web SDK",
   "type": "module",
-  "jsdelivr": "build/iife/bundle.js",
-  "types": "build/types/index.d.ts",
-  "browser": "build/esm/index.js",
-  "module": "build/esm/index.js",
-  "main": "build/cjs/index.cjs",
+  "jsdelivr": "./dist/embrace-web-sdk.js",
+  "module": "./dist/index.js",
```
overbalance, line 7: This enables importing the sdk in Webpack 4.

PR #806, opened 2025-09-17
feat(build): add build validation scripts

overbalance:
Once I added es-check and publint, it made sense to merge all the kinds of validation into one command.

- Add validation scripts for SDK and CLI builds to ensure quality and catch regressions
- Scripts run automatically in CI after builds complete
- Configurable size limits via environment variables

## Changes
- Add `scripts/validate-build.sh` for SDK validation
- Add `scripts/validate-cli.sh` for CLI validation  
- Add `validate` commands to package.json
- Integrate validation into CI workflow

## SDK Validation Checks
- ES6/ES2022 compatibility via es-check
- Package publishing readiness via publint
- Bundle size limits (configurable, default 47KB gzipped)
- ESM and CommonJS exports work correctly
- Sourcemap validity and completeness
- Load performance benchmarking
- No unexpected require() calls in distribution files

## CLI Validation Checks
- ES2022 compatibility
- Package publishing readiness
- Bundle size limits (configurable, default 15KB)
- Executable has proper shebang
- CLI executes and shows help
- Package.json bin field configured correctly
- ES module compliance

## Test plan
- [x] Run `npm run validate` - both SDK and CLI validations pass
- [x] Run `npm run sdk:validate` - SDK validation passes
- [x] Run `npm run cli:validate` - CLI validation passes
- [x] Scripts work with both old (build/) and new (dist/) directory structures
- [x] Size limits are enforced and configurable via environment variables

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,8 +46,8 @@
     "sdk:lint:eslint": "eslint --max-warnings=0",
     "sdk:lint:eslint:fix": "eslint --fix --max-warnings=0 --no-warn-ignored",
     "sdk:tsc:check": "tsc --build --noEmit tsconfig.json cli/tsconfig.json",
-    "sdk:validate": "MAX_BUNDLE_SIZE_KB=47 bash scripts/validate-build.sh",
-    "cli:validate": "MAX_CLI_SIZE_KB=15 bash scripts/validate-cli.sh",
+    "sdk:validate": "bash scripts/validate-build.sh",
+    "cli:validate": "bash scripts/validate-cli.sh",
     "commitlint": "commitlint",
     "playwright": "playwright",
     "prepare": "if [ -d .git ]; then husky; fi"
--- a/package.json
+++ b/package.json
@@ -13,9 +13,10 @@
   },
   "sideEffects": false,
   "scripts": {
-    "prepublishOnly": "npm run sdk:compile:clean && npm run sdk:compile",
+    "prepublishOnly": "npm run sdk:compile:clean && npm run sdk:compile && npm run sdk:validate",
     "clean": "npm run clean --workspaces --if-present && npx rimraf .sonda build node_modules",
     "compile": "npm run sdk:compile && npm run compile --workspaces --if-present",
+    "validate": "npm run sdk:validate && npm run cli:validate",
     "dev": "npm run sdk:compile:esm:watch & npm run sdk:compile:types:watch & npm run dev --workspace=embrace-web-sdk-react-demo",
     "demo": "npm run demo --workspace=embrace-web-sdk-react-demo",
     "sdk:analyze": "npx tsx scripts/measureSize.ts",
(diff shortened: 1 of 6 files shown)


scripts/validate-cli.sh
```diff
@@ -0,0 +1,109 @@
...
+  echo "  ✅ CLI has command structure"
+else
+  echo "  ⚠️  No commands found in CLI help"
+fi
+
+cd - > /dev/null
+
+# 5. Verify package.json bin field
+echo "Checking package.json bin configuration..."
+if grep -q '"embrace-web-cli"' $CLI_DIR/package.json; then
```
jpmunz, line 88: technically "embrace-web-cli" could show up somewhere somewhere else in package.json and still pass here, would it have been enough that we were able to run `npx embrace-web-cli` above to confirm the bin configuration?

scripts/validate-cli.sh
```diff
@@ -0,0 +1,109 @@
...
+  echo "  ✅ CLI bin field configured"
+else
+  echo "  ❌ Missing bin configuration in package.json"
+  exit 1
+fi
+
+# 6. Check ES module syntax
+echo "Checking for ES module compliance..."
+if grep -q "require(" "$CLI_FILE" | grep -v "#!/usr/bin/env node"; then
```
jpmunz, line 97: should this just be `grep -q "require(" "$CLI_FILE"`? What is the additional pipe to `| grep -v "#!/usr/bin/env node"`?

scripts/validate-cli.sh
```diff
@@ -0,0 +1,109 @@
...
+  echo "  ⚠️  Found require() calls in ESM CLI"
```
jpmunz, line 98: should this be an error like the equivalent check in `scripts/validate-build.sh`?

scripts/validate-cli.sh
overbalance: Thanks, that's just wrong (but passing). It's supposed to find the shebang and not find require. I'll fix.

scripts/validate-cli.sh
overbalance: yes, this should have been error and not warning.

scripts/validate-cli.sh
overbalance: yeah, that is redundant

package.json
```diff
@@ -45,6 +46,8 @@
     "sdk:lint:eslint": "eslint --max-warnings=0",
     "sdk:lint:eslint:fix": "eslint --fix --max-warnings=0 --no-warn-ignored",
     "sdk:tsc:check": "tsc --build --noEmit tsconfig.json cli/tsconfig.json",
+    "sdk:validate": "MAX_BUNDLE_SIZE_KB=47 bash scripts/validate-build.sh",
```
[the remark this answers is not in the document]
joaquin-diaz: Integration tests will also fail if the build is above the threshold, why do we need this extra validation?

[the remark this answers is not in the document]
overbalance: I agree it's redundant having two values to set. It it nice to be able to quickly see the validation and size without running the integration. I'm open to removing it.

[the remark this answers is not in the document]
overbalance: I'm going to report the size but remove the condition.

PR #852, opened 2025-09-29
EMBR-9083 chore: attach session.previous_id attribute to emitted logs

jpmunz:
## Goal

Adding session.previous_id to logs to help diagnose issues where we emit a log when there's no active session


src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -251,6 +252,7 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
     this._sessionSpan.end();
     this._sessionSpan = null;
     this._activeSessionStartTime = null;
+    this._previousSessionId = this._activeSessionId;
```
jpmunz, line 255: this won't track across cold starts, I think fine for now since if we do get a cold start that means we'll also have a fresh active session ID

overbalance (approved):

PR #853, opened 2025-09-29
EMBR-9083 chore: add state attribute to emitted logs

jpmunz:
## Goal

Adding emb.state to logs to help diagnose issues with logs so that we can tell if they happened in the foreground or background


src/utils/getState.ts
```diff
@@ -0,0 +1,7 @@
+import { EMB_STATES } from '../constants/index.js';
+import type { VisibilityStateDocument } from '../common/index.js';
+
+export const getState = (visibilityDoc: VisibilityStateDocument) =>
```
overbalance, line 4: I see that this is the current name but it reads too generic. Does this need to match mobile or can it be modified to something like `getVisibilityState`?

jpmunz: we should keep 'emb.state' as the final attribute name to match mobile but ya can change this method to be clear, updated

PR #855, opened 2025-09-30
EMBR-9105 fix: include resource when sending expired pending spans

jpmunz:
## Goal

When exporting pending spans from local storage we were not including the same `resource` object that was setup in `initSDK` which prevented these spans from being processed in the same way since they were missing app_version, etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -183,6 +183,5 @@
   "engines": {
     "node": ">=20",
     "npm": "11.4.2"
-  },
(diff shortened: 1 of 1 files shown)


pablomatiasgomez (approved):
nice catch. I guess this explains why I was not seeing any attribute with `from_storage` but I cannot understand why the missing sessions actually dropped? That is actually really weird

PR #856, opened 2025-10-01
EMBR-9074 chore: session diagnostic attributes for export failure

jpmunz:
## Goal

Additional diagnostic attributes on the session span to track when we have failures in recording. Since these failures could mean we never get the session sent to the backend this is adding "emb.export_failed.*" on the current session but also "emb.previous_export_failed.*" on the following session so that we have an extra chance to be notified about the failure


src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -325,6 +326,8 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
     this._activeSessionId = generateUUID();
     this._activeSessionStartTime = this._perf.getNowHRTime();
     this._activeSessionCounts = {};
+    const previouslyRecordCounts = this._nextSessionCounts;
```
overbalance, line 329: should this be `previousRecordCounts`?

overbalance (approved):
One question.

src/processors/EmbraceSessionBatchedSpanProcessor/EmbraceSessionBatchedSpanProcessor.test.ts
```diff
@@ -116,40 +128,76 @@ describe('EmbraceSessionBatchedSpanProcessor', () => {
...
+      expect(sessionSpan.attributes).to.have.property(
+        `emb.export_failed.${test.expectedAttributeSuffix}`,
+        1
+      );
+      expect(sessionSpan.attributes).not.to.have.property(
+        `emb.previous_export_failed.${test.expectedAttributeSuffix}`
+      );
 
-    await Promise.resolve();
+      const nexSessionSpan = finishedSpans[1];
```
joaquin-diaz, line 187:
```suggestion
      const nextSessionSpan = finishedSpans[1];
```

src/processors/EmbraceSessionBatchedSpanProcessor/EmbraceSessionBatchedSpanProcessor.ts
```diff
@@ -9,22 +9,35 @@ import { EMB_TYPES, KEY_EMB_TYPE } from '../../constants/index.js';
...
 const isSessionSpan = (span: ReadableSpan | SessionSpan): span is SessionSpan =>
   span.attributes[KEY_EMB_TYPE] === EMB_TYPES.Session;
 
+type ExportFailureReason = 'concurrent_limit' | 'fetch_error' | 'unknown';
+
+const exportFailureAttributeKey = (
+  reason: ExportFailureReason,
+  session: 'current' | 'previous'
+) =>
+  `emb.${session === 'current' ? 'export_failed' : 'previous_export_failed'}.${reason}`;
```
pablomatiasgomez, line 27: I would probably change this so that reason is the value instead of a different key.
e.g.:
`emb.export_failure = fetch_error` / `emb.previous_export_failure = fetch_error`

thoughts?

pablomatiasgomez (approved):

PR #859, opened 2025-10-03
feat: tag all logs with current route

joaquin-diaz:
## Goal

Tag all logs with current route if provided

## Testing

Unit tests


src/api-page/manager/types.ts
```diff
@@ -0,0 +1,14 @@
+export interface Route {
+  // This is the path of the route before replacing the URL params. i.e. /products/:productId
+  path: string;
```
jpmunz, line 3: should this be optional?

src/managers/EmbracePageManager/EmbracePageManager.ts
```diff
@@ -0,0 +1,19 @@
+import type { PageManager, Route } from '../../api-page/index.js';
+import { generateUUID } from '../../utils/index.js';
+
+export class EmbracePageManager implements PageManager {
+  private _currentRoute: Route | null = null;
+  private _currentPageId: string = generateUUID();
```
jpmunz, line 6: should this be `null` to start with as well? Just thinking about the initial case where `getCurrentRoute` returns null I would expect there to be a valid PageId either

src/constants/attributes.ts
```diff
@@ -23,6 +23,8 @@ export const KEY_EMB_W3C_TRACEPARENT = 'emb.w3c_traceparent';
 export const KEY_VIEW_NAME = 'view.name';
 export const KEY_EMB_NAVIGATION_SOURCE = 'emb.navigation_source';
 export const KEY_EMB_REFERRER_URL = 'emb.referrer_url';
+export const KEY_EMB_SUFRACE_NAME = 'app.surface.name';
+export const KEY_EMB_SUFRACE_ID = 'app.surface.id';
```
jpmunz, line 27:
```suggestion
export const KEY_EMB_SURFACE_NAME = 'app.surface.name';
export const KEY_EMB_SURFACE_ID = 'app.surface.id';
```

I'm a bit torn since this is the only place we refer to surface I also wonder if overall the SDK logic would be clearer with something like:

```suggestion
export const KEY_EMB_PAGE_PATH = 'app.surface.name';
export const KEY_EMB_PAGE_ID = 'app.surface.id';
```

either way I think could use a quick comment here explaining we're mapping the web specific page/route onto the more general client-side surface concept

jpmunz (commented):
lgtm just a few questions, no need to block on my review

overbalance (approved):

PR #860, opened 2025-10-03
[EMBR-8969] feat(document-load): collect CORS and cache resource status

overbalance:
## Summary
- Adds `cors_restricted` and `failed_request` quality flags to identify resources with timing/size limitations
- Adds `cache_validated` flag for 304 responses
- Collects extended PerformanceResourceTiming attributes: `deliveryType`, `renderBlockingStatus`, `responseStatus`

## Test plan
- [x] Tests added for quality flags and new attributes
- [x] Lint and compile passed
- [x] All tests passed


src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
```diff
@@ -399,6 +450,91 @@ export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<Docu
...
+
+      if (hasTimingData) {
+        resourceQualityFlags.push('cors_restricted');
+      } else {
+        // No timing or size data - request likely failed or was aborted
+        resourceQualityFlags.push('failed_request');
+      }
+    }
+
+    // 304 Not Modified responses typically show 300 bytes transferSize (HTTP headers only, no body)
```
jpmunz, line 524: is there a reference we could link to here for the 300 bytes magic number?

src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
```diff
@@ -399,6 +450,91 @@ export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<Docu
...
+    // This indicates cache revalidation occurred but deliveryType may not reflect it in older browsers
+    const deliveryType =
+      typeof resource.deliveryType === 'string' ? resource.deliveryType : '';
+
+    if (transferSize === 300 && deliveryType !== 'cache') {
+      resourceQualityFlags.push('cache_validated');
+    }
+
+    if (resourceQualityFlags.length > 0) {
+      span.setAttribute(ATTR_RESOURCE_FLAGS, resourceQualityFlags.join(','));
```
jpmunz, line 534: I'd double check what's easiest for parsing on the backend here, it may be preferable to have individual attribute keys for each quality type

jpmunz (commented):
lgtm just a few questions, no need to block on my review

src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
overbalance: Talked to Pablo and Santi and this is better as 3 attributes.

pablomatiasgomez:
> @pablomatiasgomez following up from planning discussion here, any particular reason to hold-off? We generally have merged new attributes in the SDK prior to being able to ingest them unless there's an open question here about what the final values should be?

looks like I confused these changes with the ones we mentioned about renaming the existent attributes we have
My bad, now I understand what @overbalance was referring to before, this should be good to merge

PR #877, opened 2025-10-07
chore: add dependabot groups

overbalance:


.github/dependabot.yml
```diff
@@ -3,7 +3,12 @@ updates:
   - package-ecosystem: 'github-actions'
     directory: '/'
     schedule:
-      interval: 'weekly'
+      interval: 'monthly'
```
vitaliyf, line 6:
```suggestion
      interval: 'weekly'
```

these are the easiest to merge - if job passes then safe to merge. you guys should just spent 5 seconds to click approve on those instead of leaving them off forever :)

.github/dependabot.yml
```diff
@@ -3,7 +3,12 @@ updates:
...
+    groups:
+      github-actions:
+        patterns:
+          - '*'
```
vitaliyf, line 10: i wouldn't do that - it makes it super confusing since you are combining so many different things together. if you really want to - then group "actions/*" (since those are all from GitHub) and leave everything else ungroup (since they are all unrelated to each other)

.github/dependabot.yml
```diff
@@ -13,22 +14,59 @@ updates:
...
       interval: 'weekly'
-    open-pull-requests-limit: 50
+    open-pull-requests-limit: 5
+
     groups:
-      vite:
+      # Production dependencies - keep separate and prioritized
+      open-telemetry:
         patterns:
+          - '@opentelemetry/*'
```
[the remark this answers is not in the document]
vitaliyf: No it should not! The original pattern matches all packages that specifically belong to NPM organization "opentelemetry", and the suggested pattern will match all organizations whose name starts with that prefix.

overbalance:
@vitaliyf Ive had a word with the 🤖. Thanks for the review.

PR #899, opened 2025-10-15
feat(session): properly handle back-forward cache (bfcache) transitions

overbalance:

Browser back/forward cache (bfcache) restores pages from memory without reloading, causing sessions to never end and navigation source to be misreported as "reload" instead of "back_forward".

- Listen for `pagehide`/`pageshow` events with `event.persisted === true` to detect bfcache transitions
- End sessions with reason `'bfcache'` when entering bfcache (or `'state_changed'` if visibility instrumentation handles it first)
- Start new sessions with reason `'bfcache_restore'` and navigation source `'back_forward'` when restoring
- Add coordination in `SpanSessionVisibilityInstrumentation` to skip visibility change handling after bfcache restoration (prevents double-session creation)
- Add 8 tests for bfcache functionality across session manager and visibility instrumentation
- Document bfcache behavior and coordination in README

## Testing

- All 560 tests pass
- Navigate to external site, press back button
- Verify: New session created with `'back_forward'` navigation source
- Verify: Previous session exported (reason depends on whether visibility instrumentation is enabled)
- Verify: Only one session created on restoration (no double-handling)

### DIFF
--- a/packages/web-sdk/src/api-sessions/manager/types.ts
+++ b/packages/web-sdk/src/api-sessions/manager/types.ts
@@ -39,7 +39,8 @@ export type ReasonSessionEnded =
   | 'timer' // max_time_reached limit
   | 'manual' // using the public api
   | 'max_size_reached'
-  | 'state_changed'; // visibility change
+  | 'state_changed' // visibility change
+  | 'bfcache'; // page entering back-forward cache
 
 export type PropertyOptions = {
   lifespan?: 'permanent';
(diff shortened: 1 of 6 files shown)


jpmunz (commented):
Logic for the fix lgtm, a couple overall comments:
* I would move the logic and documentation from `EmbraceSpanSessionManager` to a standalone instrumentation like we have for `SpanSessionVisibilityInstrumentation`, etc.
* Is the issue specific to Chrome? If that's the case I'm wondering if it would be more explicit to hook into the resume/freeze events, would these map 1:1 with pagehide/pageshow triggered with persisted=true?

PR #1220, opened 2026-03-20
chore: use activationStart and pageshow for T=0

joaquin-diaz:

\`OTelPerformanceManager\` used \`window.performance.timeOrigin\` as T=0, which doesn't account for when the user actually starts seeing content. This leads to inaccurate session start times for prerendered pages and back-forward cache (BFCache) restores.

- Added \`getPageStartMillis()\` to \`OTelPerformanceManager\` and \`PerformanceManager\` interface — returns \`max(timeOrigin + activationStart, pageShowEpochMs)\` as the effective T=0
- \`epochMillisFromOriginOffset\` now uses \`getPageStartMillis()\` as the base so all timestamps are relative to the effective page start
- \`millisSinceHRTime\` is clamped to 0 to prevent negative durations for events recorded before the effective page start
- Module-level \`_pageShowMillis\` (updated via exported \`updatePageShowMillis\`, aliased as \`updatePageStartMillis\`) is shared across all \`OTelPerformanceManager\` instances
- \`initSDK\` registers a \`pageshow\` listener after all SDK validation succeeds to update T=0 on BFCache restores

### DIFF
--- a/packages/web-sdk/src/instrumentations/clicks/ClicksInstrumentation/ClicksInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/clicks/ClicksInstrumentation/ClicksInstrumentation.ts
@@ -64,7 +64,7 @@ export class ClicksInstrumentation extends EmbraceInstrumentationBase {
               ),
               'tap.coords': `${event.x.toString()},${event.y.toString()}`,
             },
-            this.perf.epochMillisFromOriginOffset(event.timeStamp),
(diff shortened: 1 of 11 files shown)


andydavies:
The `activationStart` code looks OK to me

@joseph-embrace how do we handle BFCache restores in Lux – I had a read of the code but couldn't completely follow how we were deriving the epoch when the page was restored

joseph-embrace (approved):
🎉 lgtm!

jpmunz (approved):

PR #1274, opened 2026-04-13
feat: add UserTimingInstrumentation

joaquin-diaz:

Captures `performance.mark()` and `performance.measure()` entries from the browser's User Timing API and forwards them as Embrace logs, giving customers visibility into their custom performance markers.

- **New**: `utils/performanceObserver/` — shared `isEntryTypeSupported` and `createPerformanceObserver<T>` helpers (used by all Browser Timing instrumentations)
- **New**: `UserTimingInstrumentation` — extends `EmbraceInstrumentationBase`; two separate buffered `PerformanceObserver`s (one for `'mark'`, one for `'measure'`); URL-scoped deduplication; fixed volume caps (200 marks / 100 measures); `entry.detail` serialized as JSON string in log body; `emb.type: ux.user_timing`
- **Refactor**: `LoafInstrumentation` updated to use the new shared helpers
- **Demo**: Browser Timing fieldset with Trigger Mark / Trigger Measure buttons


demo/frontend/src/App.tsx
```diff
@@ -514,6 +514,29 @@ const App = () => {
...
+              performance.mark('demo-mark', { detail: { source: 'demo' } })
+            }
+          >
+            Trigger Mark
+          </button>
+          <button
+            type="button"
+            onClick={() => {
+              performance.mark('demo-measure-start');
+              performance.measure('demo-measure', 'demo-measure-start');
```
jpmunz: does this end up having a non-zero duration? Should it be delayed slightly?

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
```diff
@@ -0,0 +1,151 @@
...
+      this.enable();
+    }
+  }
+
+  public enable(): void {
+    if (this._isEnabled) {
+      return;
+    }
+
+    if (!isEntryTypeSupported('mark')) {
```
jpmunz: is a separate check for 'measure' required or is that part of 'mark'?

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
```diff
@@ -0,0 +1,151 @@
...
+    this._measureCount = 0;
+
+    if (this._markObserver) {
+      this._markObserver.disconnect();
+    }
+    if (this._measureObserver) {
+      this._measureObserver.disconnect();
+    }
+
+    this._markObserver = createPerformanceObserver<PerformanceMark>(
```
jpmunz, line 66: ~are 2 separate observers needed here since they use a common `_processEntry` function? Could the type here be `PerformanceMark | PerformanceMeasure` and use a single one?~

ah I guess this isn't possible, you can't call `observe` with multiple types AND `buffered`: https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/observe#entrytypes

maybe worth adding that restriction as a comment in packages/web-sdk/src/utils/performanceObserver/performanceObserver.ts since it may not be obvious when looking at the code later

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    const key = `${location.href}::${entry.name}`;
+    if (this._seenEntries.has(key)) {
+      return;
+    }
+    this._seenEntries.add(key);
+
+    if (entry.entryType === 'mark') {
+      if (this._markCount >= MARK_CAP) {
+        return;
```
jpmunz: can be a separate ticket, something we should consider is applying these limits through the LimitManager, that way we can leverage the diagnostic logic to have a debug attribute included on the session to let us know how many of these we're dropping in practice: https://github.com/embrace-io/embrace-web-sdk/blob/70c310af2a9d6d9d5a26b8ea45ff859f0d2e228e/packages/web-sdk/src/managers/EmbraceLimitManager/EmbraceLimitManager.ts#L205

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
```diff
@@ -0,0 +1,151 @@
...
+      ),
+      [ATTR_USER_TIMING_DURATION]: entry.duration,
+      [ATTR_USER_TIMING_ENTRY_TYPE]: entry.entryType,
+    };
+
+    const detail = (entry as PerformanceMark).detail;
+    const body = detail != null ? JSON.stringify(detail) : undefined;
+
+    this.logger.emit({
+      ...(body !== undefined ? { body } : {}),
```
jpmunz: wouldn't this be equivalent?

```suggestion
      body,
```

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
```diff
@@ -0,0 +1,151 @@
...
+    this._measureCount = 0;
+  }
+
+  private _processEntry(entry: PerformanceMark | PerformanceMeasure): void {
+    if (!this._isEnabled) {
+      return;
+    }
+
+    const key = `${location.href}::${entry.name}`;
+    if (this._seenEntries.has(key)) {
```
jpmunz, line 111: can you add a comment for this requirement? Is it that we're only ever allowing one mark or measure with the same name per url?

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
```diff
@@ -0,0 +1,151 @@
...
+      this._markObserver.disconnect();
+    }
+    if (this._measureObserver) {
+      this._measureObserver.disconnect();
+    }
+
+    this._markObserver = createPerformanceObserver<PerformanceMark>(
+      'mark',
+      (entries) => {
+        for (const entry of entries) {
```
jpmunz: I wonder if this loop should be part of the helper as well, not sure if we'd ever have any other pattern, if not we could do `createPerformanceObserver<PerformanceMark>('mark', this._processEntry, this._diag);`

demo/frontend/src/App.tsx
joaquin-diaz: Ah yeah, good call

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
joaquin-diaz: Is not but measure wouldn't work without mark. I'll check for both just in case

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
joaquin-diaz: Yep

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
joaquin-diaz: Yes, we only keep track of the first ocurrence by URL

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
joaquin-diaz: Added the limit manager

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/constants.ts
```diff
@@ -0,0 +1,9 @@
+export const USER_TIMING_EVENT_NAME = 'emb-user-timing';
+
+export const ATTR_USER_TIMING_NAME = 'emb.user_timing.name';
+export const ATTR_USER_TIMING_START_TIME = 'emb.user_timing.start_time';
+export const ATTR_USER_TIMING_DURATION = 'emb.user_timing.duration';
+export const ATTR_USER_TIMING_ENTRY_TYPE = 'emb.user_timing.entry_type';
```
overbalance: Our namespaced keys typically start with `KEY_EMB_`

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/constants.ts
```diff
@@ -0,0 +1,9 @@
...
+
+export const MARK_CAP = 200;
+export const MEASURE_CAP = 100;
```
overbalance: In other places we prefix limits with MAX
```suggestion
export const MAX_MARK_ENTRIES = 200;
export const MAX_MEASURE_ENTRIES = 100;
```

joaquin-diaz: Moved to the limit mananger

packages/web-sdk/src/instrumentations/loaf/LoafInstrumentation/LoafInstrumentation.ts
```diff
@@ -62,43 +66,31 @@ export class LoafInstrumentation extends EmbraceInstrumentationBase {
...
+    }
 
-      this._observer.observe({
-        type: 'long-animation-frame',
-        buffered: true,
-      });
+    this._observer =
+      createPerformanceObserver<PerformanceLongAnimationFrameTiming>(
+        'long-animation-frame',
+        (entry) => this._processEntry(entry),
```
jpmunz, line 83: I think can be simplified to just:

```suggestion
        this._processEntry,
```

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
```diff
@@ -0,0 +1,139 @@
...
+      return;
+    }
+
+    const attributes: Record<string, string | number> = {
+      [KEY_EMB_TYPE]: EMB_TYPES.UserTiming,
+      [KEY_EMB_USER_TIMING_NAME]: entry.name,
+      [KEY_EMB_USER_TIMING_START_TIME]: this.perf.millisFromZeroTime(
+        entry.startTime,
+      ),
+      [KEY_EMB_USER_TIMING_DURATION]: entry.duration,
```
jpmunz: will this be undefined for `PerformanceMark` entries then?

jpmunz (approved):

packages/web-sdk/src/instrumentations/loaf/LoafInstrumentation/LoafInstrumentation.ts
joaquin-diaz: We would actually need to do `this._processEntry.bind(this)` to keep the `this` context in the other function. I feel like this is more readable

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
joaquin-diaz: Duration is 0 for marks. In the backend we then filter out marks when aggregating to avoid 0 affecting the duration metrics

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
```diff
@@ -0,0 +1,160 @@
...
+      return;
+    }
+
+    if (entry.entryType === 'measure') {
+      const span = this.tracer.startSpan(entry.name, {
+        startTime: this.perf.epochMillisFromOriginOffset(entry.startTime),
+        attributes: {
+          // Until we decided if we want to create a new pipeline for Measures,
+          // make sure we add Perf as emb type to use the existing spans pipeline for these entries
+          [KEY_EMB_TYPE]: EMB_TYPES.Perf,
```
jpmunz: I would double-check this, I think it might go through that pipeline even without this type being explicitly set

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
```diff
@@ -0,0 +1,160 @@
...
+        },
+      });
+      span.end(
+        this.perf.epochMillisFromOriginOffset(entry.startTime + entry.duration),
+      );
+      return;
+    }
+
+    const detail = (entry as PerformanceMark).detail;
+    const body = detail != null ? JSON.stringify(detail) : undefined;
```
jpmunz, line 144: should we try and include this as an attribute on the measure span as well?

joaquin-diaz: Yeah, I realized after pushing

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
joaquin-diaz: It doesn't need to be `EMB_TYPES.Perf` specifically but it needs to have an emb_type. See https://github.com/embrace-io/go/blob/master/lib/codec/spans.go#L187. But regardless of that I removed it from here since we're using the EmbraceTraceManager which already adds `EMB_TYPES.Perf`. If we decide in the future we want a dedicated loader we can change the emb type for something else

PR #1275, opened 2026-04-14
fix: add emb.web_vital.id to LoafInstrumentation report

joaquin-diaz:
LoAF reports were missing the `emb.web_vital.id` attribute

- Add `generateWebVitalID` utility (matching the web-vitals library ID format) and export it from `utils/index.ts`
- Use it in `LoafInstrumentation` to set `emb.web_vital.id` on each session's report, regenerated on every session reset

### DIFF
--- a/packages/web-sdk/src/instrumentations/loaf/LoafInstrumentation/LoafInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/loaf/LoafInstrumentation/LoafInstrumentation.ts
@@ -47,7 +47,6 @@ export class LoafInstrumentation extends EmbraceInstrumentationBase {
   private _longestDurationExcludingFirst = 0;
   private _totalBlockingDuration = 0;
   private _scriptSummaries: ScriptSummaries = new Map();
-  private _id: string = generateWebVitalID();
 
   public constructor({ diag, perf }: LoafInstrumentationArgs = {}) {
     super({
@@ -212,7 +211,7 @@ export class LoafInstrumentation extends EmbraceInstrumentationBase {
 
     const attrs: Record<string, string | number> = {
       [KEY_EMB_TYPE]: EMB_TYPES.WebVital,
-      ['emb.web_vital.id']: this._id,
+      ['emb.web_vital.id']: generateWebVitalID(),
       ['emb.web_vital.name']: 'TBD',
       ['emb.web_vital.value']: Math.round(this._totalBlockingDuration),
       ['emb.web_vital.rating']: rating,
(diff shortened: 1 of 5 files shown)


packages/web-sdk/src/utils/generateWebVitalID.ts
```diff
@@ -0,0 +1,5 @@
+// Generates a unique ID for web vitals metrics, following the format "v5-{timestamp}-{randomNumber}"
+// https://github.com/GoogleChrome/web-vitals/blob/main/src/lib/generateUniqueID.ts
+export const generateWebVitalID = () => {
+  return `v5-${Date.now()}-${Math.floor(Math.random() * (9e12 - 1)) + 1e12}`;
```
jpmunz: what do we think about a 'embrace-v1-' prefix instead to differentiate telemetry coming from this sdk vs the web-vitals library?

packages/web-sdk/src/instrumentations/loaf/LoafInstrumentation/LoafInstrumentation.ts
```diff
@@ -218,6 +220,7 @@ export class LoafInstrumentation extends EmbraceInstrumentationBase {
 
     const attrs: Record<string, string | number> = {
       [KEY_EMB_TYPE]: EMB_TYPES.WebVital,
+      ['emb.web_vital.id']: this._id,
```
jpmunz: what about just `['emb.web_vital.id']:  generateWebVitalID(),` here instead of keeping the instance variable?

packages/web-sdk/src/utils/generateWebVitalID.ts
joaquin-diaz: Yeah sounds good. But I would use `embracev1` instead of `embrace-v1` to avoid creating an extra "part" and keeping the same `<version>-<ts>-<random_number>` structure

packages/web-sdk/src/utils/generateWebVitalID.ts
```diff
@@ -0,0 +1,5 @@
+// Generates a unique ID for web vitals metrics, following the format "v5-{timestamp}-{randomNumber}"
+// https://github.com/GoogleChrome/web-vitals/blob/main/src/lib/generateUniqueID.ts
+export const generateWebVitalID = () => {
+  return `embracev1-${Date.now()}-${Math.floor(Math.random() * (9e12 - 1)) + 1e12}`;
```
overbalance: should this version match the sdk?
```suggestion
  return `embracev2-${Date.now()}-${Math.floor(Math.random() * (9e12 - 1)) + 1e12}`;
```

joaquin-diaz: Yeah, we might end up using this ID for other things as well

jpmunz (approved):

PR #1276, opened 2026-04-15
refactor(web-vitals): generalize attribution extraction to capture all primitives

overbalance:

The per-metric-type hardcoded attribution lists required manual updates whenever the web-vitals library added fields, causing some primitive values to be silently dropped.

- Replace hardcoded per-metric attribution lists with a generic `Object.entries()` loop that captures all primitive values (string, number, boolean)
- Extract LoAF scripts (INP) and TTFB sub-parts into standalone functions, preserving their derived/computed attribute logic
- Extract attribute key constants into `attributes.ts`
- Use attribute key constants instead of inline strings

## Testing

- Updated existing test expectations to include attribution values that the generic loop now captures (e.g., `largestShiftTime`, `loadState`, `interactionTime`, `waitingDuration`)
- Added tests for non-primitive filtering, falsy primitive preservation, null attribution, boolean values, and `urlAttribution=false`
- All 663 unit tests pass, lint clean, validate clean


packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -40,196 +51,161 @@ type AttributedPage = {
   label?: string;
 };
 
+const isPrimitiveValue = (value: unknown): value is AttributeValue => {
+  if (value === undefined || value === null) {
+    return false;
+  }
```
jpmunz: this wouldn't be needed since the next checks on type would return false for these cases correct?

jpmunz (approved):
lgtm w/ a small comment

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
overbalance: It was a micro optimization for readability but you're right it can be simplified

PR #1283, opened 2026-04-20
feat: add ElementTimingInstrumentation

joaquin-diaz:

Implements the Element Timing piece of the Browser Timing epic. Captures `PerformanceElementTiming` entries (HTML elements tagged with the `elementtiming` attribute) and emits one span per entry, enabling analysis of how long individual elements take to render from navigation start.

- Add `ElementTimingInstrumentation` extending `EmbraceInstrumentationBase`
  - Guards with `isEntryTypeSupported('element')` — Chromium-only, skips silently in Firefox/Safari
  - Span start anchored to navigation start (`epochMillisFromOriginOffset(0)`), end at element render time (`entry.startTime = renderTime || loadTime`)
  - Adds a `'load'` span event at `loadTime` when present (non-zero), for image resources
  - `emb.type: ux.element_timing` with full attribute set: identifier, element tag, render_time, load_time, start_time, url, natural_width, natural_height
  - Opt-out via `'element-timing'` key in `DefaultInstrumentationConfig.omit`
- Add `EMB_TYPES.ElementTiming = 'ux.element_timing'` to attributes enum
- Wire into `setupDefaultInstrumentations` and `sdk/types.ts`
- Add Element Timing demo section in `demo/frontend`
- Fix consistency: measure spans in `UserTimingInstrumentation` now also carry `emb.type: ux.user_timing`


packages/web-sdk/src/instrumentations/element-timing/ElementTimingInstrumentation/ElementTimingInstrumentation.ts
```diff
@@ -0,0 +1,108 @@
...
+      this._observer.disconnect();
+    }
+
+    this._observer = createPerformanceObserver<PerformanceElementTiming>(
+      'element',
+      (entry) => this._processEntry(entry),
+      { diag: this._diag },
+    );
+
+    if (!this._observer) {
```
jpmunz, line 67: we should include an equivalent check in UserTimingInstrumentation

packages/web-sdk/src/instrumentations/element-timing/ElementTimingInstrumentation/ElementTimingInstrumentation.ts
```diff
@@ -0,0 +1,108 @@
...
+        [KEY_EMB_ELEMENT_TIMING_ELEMENT]: entry.element?.tagName?.toLowerCase(),
+        [KEY_EMB_ELEMENT_TIMING_RENDER_TIME]: entry.renderTime,
+        [KEY_EMB_ELEMENT_TIMING_LOAD_TIME]: entry.loadTime,
+        [KEY_EMB_ELEMENT_TIMING_START_TIME]: entry.startTime,
+        [KEY_EMB_ELEMENT_TIMING_URL]: entry.url,
+        [KEY_EMB_ELEMENT_TIMING_NATURAL_WIDTH]: entry.naturalWidth,
+        [KEY_EMB_ELEMENT_TIMING_NATURAL_HEIGHT]: entry.naturalHeight,
+      },
+    });
+    if (entry.loadTime > 0) {
```
jpmunz, line 108: is this needed given we're capturing as KEY_EMB_ELEMENT_TIMING_LOAD_TIME on the attributes?

jpmunz (commented):
lgtm just a couple small comments

packages/web-sdk/src/instrumentations/element-timing/ElementTimingInstrumentation/ElementTimingInstrumentation.ts
joaquin-diaz: It helps visualize the load time when you're looking at the span in a tree (like our spans page or an external tool like Tempo)

jpmunz (commented):
@joaquin-diaz one thing I missed from my initial review was making use of the limit manager, we should probably have some guard against a large number of spans being generated for element timing

jpmunz (approved):

PR #1284, opened 2026-04-21
feat: add ServerTimingInstrumentation

joaquin-diaz:

Captures `Server-Timing` response header data from the navigation entry so teams can observe backend performance metrics (db latency, cache hit/miss, CDN edge timings) alongside frontend session data.

- Add `ServerTimingInstrumentation` extending `EmbraceInstrumentationBase`
- Reads `performance.getEntriesByType('navigation')[0]?.serverTiming` after page load
- Emits one log per `PerformanceServerTiming` entry with `emb.server_timing.name`, `emb.server_timing.duration`, `emb.server_timing.description` attributes and `emb.type: ux.server_timing`
- Lifecycle mirrors `DocumentLoadInstrumentation`: reads immediately if `document.readyState === 'complete'`, otherwise attaches a `load` event listener removed by `disable()`
- Empty or missing `serverTiming` (e.g. CORS without `Timing-Allow-Origin`) → silent no-op
- Registered as a default instrumentation with opt-out key `'server-timing'`
- Add `EMB_TYPES.ServerTiming = 'ux.server_timing'` to attributes constants


packages/web-sdk/src/instrumentations/server-timing/ServerTimingInstrumentation/ServerTimingInstrumentation.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    window.addEventListener('load', this._onLoad);
+  }
+
+  public disable(): void {
+    window.removeEventListener('load', this._onLoad);
+  }
+
+  private _readServerTiming(): void {
+    const navEntries = performance.getEntriesByType(
```
jpmunz, line 60: we could add a guard similar to https://github.com/embrace-io/embrace-web-sdk/blob/70c310af2a9d6d9d5a26b8ea45ff859f0d2e228e/packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts#L131 to prevent this from triggering more than once (I think would only happen if it `enable` was called multiple times)

packages/web-sdk/src/instrumentations/server-timing/ServerTimingInstrumentation/ServerTimingInstrumentation.ts
```diff
@@ -0,0 +1,71 @@
...
+      'navigation',
+    ) as PerformanceNavigationTiming[];
+    const serverTiming = navEntries[0]?.serverTiming;
+
+    if (!serverTiming?.length) {
+      return;
+    }
+
+    for (const entry of serverTiming) {
+      this.logger.emit({
```
jpmunz, line 74: maybe less important in this case but we could consider applying the limit manager here as well

packages/web-sdk/src/instrumentations/server-timing/ServerTimingInstrumentation/ServerTimingInstrumentation.ts
joaquin-diaz: Yep, good call

jpmunz (approved):

PR #1287, opened 2026-04-21
REFERENCE EMBR-11208 add user session lifecycle and session part attributes

overbalance:
> **Stacked on #1295.** That prep PR carries the orthogonal utils, demo styling, and `hasFocus` plumbing; this PR is the session redefinition itself. Land #1295 first.

Aligns the SDK with the common user session spec. What we previously called a "span session" was a foreground interval; this PR renames it to "session part" and adds a real user session that spans tabs, backgrounding, and part boundaries. User-session attributes are stamped on every span and log, and a public API is exposed for managing user sessions.

## Public surface

- `EmbraceUserSessionManager` and the public `UserSessionManager` API. Public methods: `endUserSession()` (spec §1.4 cooldown applies), `setSessionId()` (override `session.id`; must be called after `initSDK` returns), property/breadcrumb helpers, user-session listeners. Session parts are hidden behind `getSessionPartManager()`.
- Deprecated aliases preserved: `SpanSessionManager` type, `endSessionSpan()`, `addSessionStartedListener` / `addSessionEndedListener`. Listener aliases now fire on user-session boundaries (~12h or `endUserSession`); for the previous per-foreground-interval semantics use `getSessionPartManager().addSessionPartStartedListener` / `addSessionPartEndedListener`. Listeners must not synchronously re-enter lifecycle methods.
- Documented in `packages/web-sdk/src/api-sessions/README.md`.

## Lifecycle and attribution

- `session.id`, `session.previous_id`, `emb.user_session_id`, `emb.user_session_previous_id` are stamped on **every** span and log; empty strings when no current/previous user session. `emb.user_session_id` and `emb.user_session_previous_id` are snapshotted at span `onStart` so a user-session termination during a long-running span (cross-tab `storage` event or max-duration rollover) cannot strip the attribution at `onEnd`. The other five user-session attributes (`emb.user_session_number`, `emb.user_session_part_number`, `emb.user_session_start_ts`, `emb.user_session_max_duration_seconds`, `emb.user_session_inactivity_timeout_seconds`) and the part attributes (`emb.session_part_id`, `emb.session_part_start_reason`, `emb.session_part_end_reason`) live on the session-part span only.
- A tab is "engaged" only when `document.visibilityState === 'visible'` AND `document.hasFocus()` is true. `startSessionPart` is a no-op when either fails; `visibilitychange` to hidden, `blur`, or `pagehide` ends the active part with reason `visibility_change`. If max-duration rollover fires while not engaged, the user session ends and storage is cleared but the new part's start is deferred until engagement returns.
- Inactivity has two layers (see `packages/web-sdk/src/api-sessions/README.md` for the diagram):
  - **Part-level (eager, web-specific):** a 30-min `setTimeout` in `SessionPartActivityInstrumentation`, reset on each input event. When it fires, the part ends with reason `'inactivity'`. This is the source of the `SessionPartEndReason='inactivity'` value.
  - **User-session-level (lazy, spec §1.1):** at part end the SDK writes `inactivityDeadlineTs = now + inactivityTimeoutMs` to localStorage and walks away. The check fires only at the next `onSessionPartStart`; no JS timer at this layer, so backgrounded tabs and reloads are correctly handled.
- Max-duration uses an active `setTimeout` (spec §1.3, 12h default).
- Cross-tab user-session termination propagates via the browser `storage` event (works under namespaced localStorage when `registerGlobally: false`).

## New instrumentation

`SessionPartActivityInstrumentation` consolidates engagement tracking, the part-inactivity timer, and BFCache handling that were previously spread across three separate instrumentations. Listens to `visibilitychange`, `focus`, `blur`, `pageshow` (canonical BFCache-restore signal), `pagehide`, and the activity events (`keydown`, `mousedown`, `mousemove`, `scroll`, throttled to 30s). Registered unconditionally; the part lifecycle depends on it.

## Removed

- `SpanSessionTimeoutInstrumentation` (4h part timer), redundant with user-session max-duration.
- `SpanSessionBrowserActivityInstrumentation` and `SpanSessionVisibilityInstrumentation`, replaced by `SessionPartActivityInstrumentation` (see "New instrumentation" above). The old debounce / mousedown-throttle / sub-5s-filter logic that suppressed tiny sessions under the per-foreground-interval model is gone; the 30-min part-inactivity behavior carries forward.
- `SpanSessionOnLoadInstrumentation` (lifecycle hook with no listeners or timers; initial part start is now inlined into `initSDK`).
- `EmbraceSpanStorage` and the `storePendingSpans` / `clearStoredSpans` machinery (a 5-min poll + 1h TTL buffer that hoarded snapshots to avoid flooding the backend with tiny sessions). The new model groups parts under a user session, so the buffer is unneeded; parts now flush on `visibilitychange` to hidden, which exports faster.
- `backgroundSessions` flag, non-spec `emb.session_part_number` attribute, dead `lastPartEndTs` / `lastPartEndType` state.
- Edge case lost: a tab opened, briefly hidden, never returned to, kept alive >1h, then closed previously exported a frozen TTL-sweep snapshot; now ends and exports on hide. Same data, earlier.

## `emb.session_part_start_reason` / `emb.session_part_end_reason`

Spec §5 names these but leaves enums unspecified. Values:

- **Start:** `init` · `visibility_change` · `activity` (input resumed after a part ended for inactivity) · `user_session_rollover`
- **End:** `manual` · `visibility_change` · `inactivity` (30-min part-inactivity timer fired with no input) · `user_session_ended`

## Deviations from spec

- **Universal session IDs (§2.1):** the four session ids stamp on every span/log, empty-string when unavailable. Foreground-only on web means off-part intervals are larger here than on mobile, and correlation is more valuable than strict alignment.
- **`emb.user_session_previous_id` / `session.previous_id`:** not in §5 but `session.previous_id` exists in OTel semantic conventions and aids backend stitching. User-set values are respected.
- **`emb.session_part_end_reason`:** renamed from `emb.session_end_type` to align with `emb.user_session_termination_reason`.
- **`emb.user_session_termination_reason`:** only `manual` and `max_duration_reached` are emitted. Spec also names `inactivity`, but lazy detection means the prior part's span is already ended by the time inactivity is observed.
- **Ordering (§4):** retry transport may reorder; backends should sort by part-span start time and `emb.user_session_start_ts` (ms since epoch).
- **Foreground-only (§1.2):** spec allows background parts; web does not emit them.
- **Remote configuration (§3):** `user_session.max_duration_seconds` and `user_session.inactivity_timeout_seconds` keys are not yet wired into dynamic config; values can only be set via `initSDK({ userSessionConfig })`.

## Testing

- `npm run validate`
- `npm run test` from `packages/web-sdk/`
- `npm run test:integration`


packages/web-sdk/src/utils/EmbraceSpanStorage/EmbraceSpanStorage.ts
```diff
@@ -143,6 +143,7 @@ export class EmbraceSpanStorage {
             storedTime,
           );
           this._storage.removeItem(key);
+          return;
```
overbalance: existing bug fixed

PR #1288, opened 2026-04-22
fix: add start_time and duration attributes to measure spans

joaquin-diaz:

`UserTimingInstrumentation` creates spans for `performance.measure` entries but was missing `emb.user_timing.start_time`, `emb.user_timing.duration`, and the `emb.user_timing.detail` attribute key lacked the `emb.` prefix used by all other keys.

- Added `emb.user_timing.start_time` and `emb.user_timing.duration` to measure span attributes, consistent with mark logs
- Renamed `user_timing.detail` → `emb.user_timing.detail` and its constant `KEY_USER_TIMING_DETAIL` → `KEY_EMB_USER_TIMING_DETAIL`
- Added test assertions for the new attributes, including a cross-check that `emb.user_timing.duration` equals the span's own `endTime - startTime` delta

### DIFF
--- a/packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
@@ -8,11 +8,11 @@ import {
 import { createPerformanceObserver } from '../../../utils/index.ts';
 import { EmbraceInstrumentationBase } from '../../EmbraceInstrumentationBase/index.ts';
 import {
+  KEY_EMB_USER_TIMING_DETAIL,
   KEY_EMB_USER_TIMING_DURATION,
   KEY_EMB_USER_TIMING_ENTRY_TYPE,
   KEY_EMB_USER_TIMING_NAME,
   KEY_EMB_USER_TIMING_START_TIME,
-  KEY_USER_TIMING_DETAIL,
   USER_TIMING_EVENT_NAME,
 } from './constants.ts';
 import type {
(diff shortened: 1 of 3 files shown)


packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
```diff
@@ -129,6 +129,10 @@ export class UserTimingInstrumentation extends EmbraceInstrumentationBase {
           [KEY_EMB_INSTRUMENTATION]:
             EMB_PERFORMANCE_INSTRUMENTATIONS.UserTiming,
           [KEY_EMB_USER_TIMING_ENTRY_TYPE]: entry.entryType,
+          [KEY_EMB_USER_TIMING_START_TIME]: this.perf.millisFromZeroTime(
```
overbalance, line 133: This should probably be timestamp to reduce redundancy

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
```diff
@@ -129,6 +129,10 @@ export class UserTimingInstrumentation extends EmbraceInstrumentationBase {
...
+            entry.startTime,
+          ),
+          [KEY_EMB_USER_TIMING_DURATION]: entry.duration,
```
overbalance, line 136: Do spans need duration?

joaquin-diaz: I'm adding duration for convenience so we have the raw value coming directly from the browser in case something goes wrong, since we're making up the span based on the measure

packages/web-sdk/src/instrumentations/userTiming/UserTimingInstrumentation/UserTimingInstrumentation.ts
joaquin-diaz: Start time is the value reported by the observer, we just adjust it to what we consider our zero time. There might be value on the user having metrics around how much time it passed since the mark/measure is reported. No way of getting that from span start time or log timestamp without having a hard reference to the zero time

overbalance (approved):

PR #1292, opened 2026-04-26
EMBR-11208 refactor(session): remove empty session avoidance feature

overbalance:

Removing the empty-session-avoidance feature (which buffered short, no-interaction sessions in localStorage instead of ending them) per EMBR-11208. The visibility instrumentation now ends a user session on visibility change unconditionally, the same way it does for interactions and timeouts. Several adjacent dead/dangling pieces are cleaned up while the area is being touched.

- Visibility instrumentation: drop \`visibilityWaitTimeMs\` debounce option, interaction-tracking, and limited-session config; visibility changes always end the current user session and start a new one for \`visible\` (or for \`hidden\` with \`backgroundSessions\`).
- Wrap the visibility-change handler in try/catch so a throw in \`endSessionSpanInternal\` / \`startSessionSpan\` cannot escape into the browser's \`visibilitychange\` event.
- Remove the pending-span storage subsystem entirely: \`EmbraceSpanStorage\`, the 5-minute \`setInterval\`, and the \`getPendingSpansCount\` / \`storePendingSpans\` / \`clearStoredSpans\` methods (plus tests).
- Remove \`EmbraceSessionBatchedSpanProcessor\`'s unused \`resource\` constructor arg.
- Remove \`SDKFeaturesManager.isEmptySessionAvoidanceEnabled()\` and the \`RemoteConfig.empty_session_avoidance_enabled_pct\` field; \`parseRemoteConfig\` silently ignores the field if the backend still sends it.
- Drop the \`emb.from_storage\` attribute key.
- \`SpanSessionManager.currentSessionAsReadableSpan\` is preserved as a \`@deprecated\` no-op (always returns \`null\`) to avoid a public API break; the supporting \`_noExportTracer\` / \`BasicTracerProvider\` are removed.
- \`DynamicSDKConfig.emptySessionAvoidanceEnabledPct\` is preserved as a \`@deprecated\` field that is ignored at runtime.
- Demo \`defaultInstrumentationConfig\` no longer sets \`'session-visibility'\`; \`setupDefaultInstrumentations.test.ts\` asserts the visibility instrumentation is wired in.

## Testing

- \`npm run check\` and \`npm run test\` from \`packages/web-sdk/\` pass locally.
- New unit test: visibility change to \`hidden\` with default config (synchronous path, \`backgroundSessions=false\`) ends the user session with reason \`state_changed\` and does not start a new one.
- Manually verified that a visibility change produces a \`state_changed\` user session end and that no \`embrace_pending_*\` keys are written to \`localStorage\` after init.


packages/web-sdk/src/instrumentations/session/SpanSessionVisibilityInstrumentation/SpanSessionVisibilityInstrumentation.ts
```diff
@@ -1,39 +1,20 @@
...
-      maxPendingSpanCount = 5,
-      visibilityDoc = window.document,
-    }: SpanSessionVisibilityInstrumentationArgs = {},
-    featureManager: SDKFeaturesManager,
-    embraceSpanProcessor?: EmbraceSessionBatchedSpanProcessor,
-  ) {
+  public constructor({
+    diag,
+    perf,
+    visibilityWaitTimeMs = 0,
```
jpmunz: @pablomatiasgomez can confirm but I believe `visibilityWaitTimeMs` was related to the same logic and could be removed along with the other changes

pablomatiasgomez: yeah this was related. That flag was meant to configure an interval where if the user goes to a tab and comes back quickly, the session "continued".

If we want to remove all this feature we should "forward events" as they happen and remove this flag

overbalance: This is removed in a PR farther up the stack but I'll remove here and rebase all.

jpmunz (approved):

PR #1294, opened 2026-04-26
4/7 EMBR-11208 refactor(session): remove SpanSessionTimeoutInstrumentation

overbalance:
> Stacked on #1293. Part of the EMBR-11208 split: each PR carves a coherent slice from the user-session redefinition WIP (#1287, kept open for reference until the stack lands).

`SpanSessionTimeoutInstrumentation` is a 4-hour session-part timer that becomes redundant once user-session max-duration (12h default) lands as part of the user-session model in a later PR.

- Delete `packages/web-sdk/src/instrumentations/session/SpanSessionTimeoutInstrumentation/` (impl, tests, constants, types, barrel)
- Drop the `SpanSessionTimeoutInstrumentation` / `SpanSessionTimeoutInstrumentationArgs` re-exports from `instrumentations/index.ts` and `instrumentations/session/index.ts`
- Drop the instantiation from `setupDefaultInstrumentations.ts`
- Drop the `'session-timeout'` config field from `DefaultInstrumentationConfig` in `sdk/types.ts`

## Intermediate behavior

Until the user-session max-duration cap ships in a later PR, session parts no longer have a time-based termination. They still rotate on visibility/activity events, so the SDK remains functional.

## Testing

- npm run validate
- npm run test from packages/web-sdk/


jpmunz (commented):
I wouldn't do this drop separately from the replacement so we can view behaviour difference together

PR #1295, opened 2026-04-26
EMBR-11208 chore(prep): orthogonal utils, demo, and test-infra improvements

overbalance:

Splits orthogonal improvements out of PR #1287 (EMBR-11208) so each lands cleanly and the parent diff shrinks.

- `NamespacedStorage`: add `getStorageEventKey(logicalKey)` so callers filtering cross-tab `storage` events can compare against the namespaced underlying key.
- `getIncrementedCount`: warn and reset to 0 when stored value is non-numeric.
- `VisibilityStateDocument`: add required `hasFocus(): boolean`; update all test mocks.
- Demo: add `.info-list-grid` CSS class and apply it to the Experience fieldset.
- Integration test utility: expand ignored-attribute list (UUID/timestamp session keys, FCP `loadState`, `session.previous_id`).
- `webpack-4` fixture: set placeholder `appID: '11111'` so init succeeds.
- `.gitignore`: ignore `.claude/worktrees/`.

## Testing

- [x] `npm run check`
- [x] `npm run lint`
- [x] `npm test` from `packages/web-sdk/` (730 pass)

### DIFF
--- a/packages/web-sdk/src/utils/getIncrementedCount.ts
+++ b/packages/web-sdk/src/utils/getIncrementedCount.ts
@@ -21,6 +21,6 @@ export const getIncrementedCount = (
     return number;
   } catch (e) {
     diag.warn(`Failed to retrieve ${key} from storage: `, e);
-    return 0;
+    return 1;
   }
 };
--- a/packages/web-sdk/src/utils/getIncrementedCount.ts
+++ b/packages/web-sdk/src/utils/getIncrementedCount.ts
@@ -10,11 +10,17 @@ export const getIncrementedCount = (
   try {
     const value = storage.getItem(key);
     let number = value ? parseInt(value, 10) : 0;
+    if (Number.isNaN(number)) {
+      diag.warn(
+        `Non-numeric value stored at ${key} (${String(value)}); resetting counter.`,
+      );
+      number = 0;
+    }
     number++;
     storage.setItem(key, number.toString());
     return number;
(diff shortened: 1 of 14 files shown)


packages/web-sdk/src/utils/getIncrementedCount.ts
```diff
@@ -10,11 +10,17 @@ export const getIncrementedCount = (
...
+      );
+      number = 0;
+    }
     number++;
     storage.setItem(key, number.toString());
     return number;
   } catch (e) {
     diag.warn(`Failed to retrieve ${key} from storage: `, e);
-    return 1;
+    return 0;
```
jpmunz: I think `1` was the correct default here, e.g. when getting the value to use for session_number in an environment without access to local storage we would expect all sessions to have session_number=1

overbalance: We can't guarantee it's the first session. Is that alright?

jpmunz: it feels consistent with the other cases, e.g. if NaN is read then the next session number will be 1, or likewise if the browser's local storage is cleared we're back to zero and will return 1 from the function. If we want session_number=0 specifically as a signal that we weren't able to read from storage it may be better to do that as a separate diagnostic attribute

jpmunz (approved):

PR #1296, opened 2026-04-26
EMBR-11208 refactor(session): inline SpanSessionOnLoadInstrumentation into initSDK

overbalance:
> Stacked on #1294. Part of the EMBR-11208 split - each PR carves a coherent slice from the user-session redefinition WIP (#1287, kept open for reference until the stack lands).

`SpanSessionOnLoadInstrumentation` and `SpanSessionTimeoutInstrumentation` are thin wrappers that, between them, only start the initial user session and schedule a 4-hour cap to roll it over. Inlining the start and dropping the timeout entirely matches #1287's design and removes dead complexity.

It also fixes a latent bug: the on-load wrapper read its session manager from the global session API (`session.getSpanSessionManager()`) at construction time. For SDK instances created with `registerGlobally: false`, the global API was never wired to the instance's manager, so the init session start went to a no-op. Calling `startSessionSpan` directly on the instance's manager fixes this.

- Inline `spanSessionManager.startSessionSpan({ reason: 'init' })` into initSDK, just before `setupDefaultInstrumentations`
- Default `emb.session_start_type` to `'manual'` when `startSessionSpan` is called without an explicit reason
- Delete `SpanSessionOnLoadInstrumentation/` and `SpanSessionTimeoutInstrumentation/` directories (impl, test, index, types, constants)
- Drop their re-exports from `instrumentations/index.ts` and `instrumentations/session/index.ts`
- Drop their instantiations from `setupDefaultInstrumentations.ts`
- Drop the `'session-on-load'` and `'session-timeout'` config fields from `DefaultInstrumentationConfig` in `sdk/types.ts`
- Update isolated instance test to expect 4 spans (init user session + manual user session) instead of 3, and assert `emb.session_start_type` on each

## Testing

- npm run validate
- npm run test from packages/web-sdk/


packages/web-sdk/src/sdk/initSDK.test.ts
```diff
@@ -1946,10 +1946,14 @@ describe('isolated instances', () => {
 
       const finishedSpans = spanExporter.getFinishedSpans();
 
-      expect(finishedSpans).to.have.lengthOf(3);
+      // initSDK starts an init session per instance; the manual
+      // startSessionSpan ends it and starts another which endSessionSpan
+      // closes - so two emb-session spans are expected.
```
jpmunz: it seems incorrect to have 2, so we have the one being created by `spanSessionManager.startSessionSpan({ reason: 'init' });` in `initSDK`, which line is creating the other? If there's already a session span is `spanSessionManager.startSessionSpan({ reason: 'init' });` needed at all?

joaquin-diaz: So if we run multiple SDK instances we'd still have 1 session shared across them?

overbalance: The second `emb-session` comes from the `sdkInstance.session.startSessionSpan()` call on line 1935: `EmbraceSpanSessionManager.startSessionSpan` ends any in-progress session before opening a new one (see `EmbraceSpanSessionManager.ts:253-256`), so it ends the init session (emitting first) and starts a fresh one, which `endSessionSpan()` then closes (emitting second).

The init `startSessionSpan({ reason: 'init' })` is still needed because in production users rarely call `startSessionSpan` manually; the init session is the default that runs from page load. The test happens to exercise both paths. I tightened the inline comment in a follow-up to spell this out.

overbalance: @joaquin-diaz  No, sessions are per-instance. Each `initSDK` call creates its own `spanSessionManager` via `setupSession()` (`initSDK.ts:217`) and the init session is started against that instance's manager (`initSDK.ts:282`). The test confirms this: `checkInstanceTelemetry` runs against each instance's own exporter and independently sees the 4 spans (including 2 `emb-session`s).

Before this PR the init session start was wired through the global session API, so for `registerGlobally: false` instances the init start went to a no-op; that latent bug is fixed here.

overbalance: @jpmunz Do you think we should conditionally check for an active session instead of calling init by default?

jpmunz: nope this is fine, I hadn't realized that test was calling startSessionSpan itself I thought under normal initialization of the sdk we were getting 2 session spans but looks like that's not the case. It might make the test a bit clearer if the "reason" attribute for each session were expected on, if I'm understanding correctly we should get one as "init" and another as "manual"? Looks good in any case though

jpmunz (approved):

packages/web-sdk/src/sdk/initSDK.test.ts
overbalance: Good idea. I added explicit reason assertions and set the default reason to `manual`.

PR #1307, opened 2026-05-04
1/4 EMBR-11208 refactor(storage): replace NamespacedStorage with EmbraceStorage

overbalance:

Foundation for the user-session redefinition stack. The session feature in PR 2 needs a storage wrapper that survives quota errors, disabled storage, and other environmental failures without crashing the SDK. Self-contained and orthogonal to the session work.

- Replaces `NamespacedStorage` with `EmbraceStorage`: a `Storage`-shaped wrapper that combines safe error handling with an optional namespace.
- `setItem`/`removeItem`/`clear` return `boolean` (`true` on success, `false` on caught failure or when writes are disabled), so callers can branch on persist-vs-fall-back without polling state.
- Reads return `null` and warn on failure. The first environmental write failure flips a sticky `writeDisabled` flag and emits one error; subsequent `setItem` calls silently return `false` without attempting the underlying write. `TypeError` from `setItem` propagates without disabling writes (a code bug at one call site shouldn't poison the whole session).
- `keys()` skips poisoned indices; `key(i)` uses `?? null` so empty-string keys are preserved. `getStorageEventKey()` exposes the underlying key for `storage` event filtering.
- `initSDK` always wraps `localStorage`/`sessionStorage` in `EmbraceStorage`, with the namespace applied only when `registerGlobally` is false and an `appID` is present.

## Stack

- 1/4 (this PR), base: ``main``
- 2/4 EMBR-11208 feat(sessions): user session lifecycle and session part attributes (#1308)
- 3/4 EMBR-11208 refactor(sessions): rename SpanSessionManager to UserSessionManager (#1309)
- 4/4 EMBR-11208 docs(sessions): README and demo migration (#1310)

## Testing

- [x] ``npm run validate``
- [x] ``npm run test`` from ``packages/web-sdk/``


packages/web-sdk/src/utils/NamespacedStorage/NamespacedStorage.ts
```diff
@@ -1,4 +1,23 @@
-export class NamespacedStorage implements Storage {
+/**
+ * Storage that exposes its underlying prefixed key for cross-tab event matching.
+ * Use `isNamespacedStorageLike` to detect; structural casts will miss it.
+ */
+export interface NamespacedStorageLike {
```
jpmunz: I'm not sure about the naming here and in `packages/web-sdk/src/utils/SafeStorage/SafeStorage.ts` since we don't otherwise have `Like` suffixed on types, what about having the type be `NamespacedStorage` and the class `EmbraceNamespacedStorage` ?

packages/web-sdk/src/utils/EmbraceNamespacedStorage/EmbraceNamespacedStorage.ts
```diff
@@ -7,34 +26,45 @@ export class NamespacedStorage implements Storage {
...
-    return this.getNamespacedKeys().length;
+    return this._getNamespacedKeys().length;
   }
 
   public clear(): void {
-    this.getNamespacedKeys().forEach((key) => {
-      this._storage.removeItem(`${this._keyPrefix}${key}`);
-    });
+    // Snapshot before removal: removeItem shifts indices mid-iteration.
+    const toRemove: string[] = [];
```
jpmunz: I'm not sure I see the functional difference from before, wouldn't this just be the same as calling `_keys` to get the string array?

packages/web-sdk/src/utils/EmbraceSafeStorage/EmbraceSafeStorage.ts
```diff
@@ -0,0 +1,123 @@
...
+  remove(key: string): boolean;
+  has(key: string): boolean;
+  /**
+   * Logical keys when namespaced; raw keys otherwise (may include foreign
+   * keys sharing the origin's storage). Best-effort: partial results on
+   * per-index errors in plain storage, empty on namespaced bulk failure.
+   */
+  keys(): string[];
+  isWriteDisabled(): boolean;
+  getStorageEventKey(logicalKey: string): string;
```
jpmunz: we could do `& NamespacedStorage` rather than having `keys` + `getStorageEventKey` again

packages/web-sdk/src/utils/EmbraceSafeStorage/EmbraceSafeStorage.ts
```diff
@@ -0,0 +1,123 @@
+import type { DiagLogger } from '@opentelemetry/api';
+import { isNamespacedStorageLike } from '../NamespacedStorage/NamespacedStorage.ts';
+
+export interface SafeStorageLike {
+  read(key: string): string | null;
```
jpmunz: why not have these function names match the `Storage` interface, e.g. `getItem`, `setItem`, etc.

packages/web-sdk/src/utils/EmbraceNamespacedStorage/EmbraceNamespacedStorage.ts
overbalance: yes, reverted.

packages/web-sdk/src/utils/EmbraceSafeStorage/EmbraceSafeStorage.ts
overbalance: decided to combine the classes. much less code now

packages/web-sdk/src/utils/EmbraceStorage/EmbraceStorage.ts
```diff
@@ -0,0 +1,145 @@
...
+  public setItem(key: string, value: string): boolean {
+    if (this._writeDisabled) {
+      return false;
+    }
+    try {
+      this._storage.setItem(this._keyName(key), value);
+      return true;
+    } catch (e) {
+      if (e instanceof TypeError) {
+        throw e;
```
jpmunz, line 64: why re-throw this type specifically?

jpmunz (commented):
lgtm overall just the one question around rethrowing

packages/web-sdk/src/utils/EmbraceStorage/EmbraceStorage.ts
overbalance: TypeError here means an error at the call site as opposed to Quota or Security. I wanted to surface author errors early and suppress environmental errors

jpmunz: but that means we'd potentially be propagating an unhandled exception up to the user's application code if the caller doesn't have their own guard in place and causing a runtime issue which we're otherwise trying to insulate the app from

overbalance:
Changes here were fixed directly https://github.com/embrace-io/embrace-web-sdk/pull/1308

packages/web-sdk/src/utils/EmbraceStorage/EmbraceStorage.ts
overbalance: removed entirely

PR #1308, opened 2026-05-04
EMBR-11208 feat(sessions): add user session lifecycle and session part attributes

overbalance:

Implements the user-session redefinition (EMBR-11208) under the existing `SpanSession*` filenames and classnames so reviewers can read this as a behavior change without rename noise. A separate follow-up will rename to `UserSession` / `SessionPart`.

References WIP: #1298.

## What this PR adds

- **Public methods on `SpanSessionManager`**: `getUserSessionId`, `getPreviousUserSessionId`, `getUserSessionStartTime`, `endUserSession`, `setSessionId`.
- **New internal interface** `SpanSessionManagerInternal extends SpanSessionManager` exposing part-level operations to instrumentations and processors: `getUserSessionAttributes`, `getUserSessionIdOverride`, `getSessionPartId`, `getSessionPartSpan`, `startSessionPartInternal`, `endSessionPartInternal`, `addSessionPartStartedListener`, `addSessionPartEndedListener`, plus tracer-provider wiring and counter helpers.
- **Types**: `UserSessionState`, `UserSessionConfig`, `UserSessionAttributes`, `SessionPartStartReason`, `SessionPartEndReason`, `UserSessionEndReason`. `UserSessionState` carries a `schemaVersion` stamp so a future incompatible shape change discards old persisted blobs cleanly instead of leaking partially-parsed fields downstream.
- **Processors**: `processors/PropertySpanProcessor/` (stamps user-session and permanent properties onto session-part spans, replacing inline `setAttributes` calls in the manager). `IdentifiableSessionLogRecordProcessor` is updated to consume `UserSessionAttributes` from the internal manager.
- **Util**: `utils/applyUserSessionAttributes.ts` (shared stamping helper for spans and logs).
- **Attributes**: `emb.session_part_id`, `emb.session_part_start_reason`, `emb.session_part_end_reason`, `emb.user_session_id`, `emb.user_session_previous_id`, `emb.user_session_number`, `emb.user_session_part_number`, `emb.user_session_start_ts`, `emb.is_final_session_part`, `emb.user_session_termination_reason`, `emb.user_session_max_duration_seconds`, `emb.user_session_inactivity_timeout_seconds`. `emb.type` value `ux.session` becomes `ux.session_part`. Removed: `emb.session_number`, `emb.session_start_type`, `emb.session_end_type`.

**Naming, kept old in this PR (renamed in the follow-up):** file paths `managers/EmbraceSpanSessionManager/`, `api-sessions/manager/{Proxy,NoOp}SpanSessionManager/`, `instrumentations/session/SpanSessionBrowserActivityInstrumentation/` (rewritten in place; consolidates the deleted `SpanSessionVisibilityInstrumentation/`). Class names `EmbraceSpanSessionManager`, `ProxySpanSessionManager`, `NoOpSpanSessionManager`, `SpanSessionBrowserActivityInstrumentation`. Public interface name `SpanSessionManager` (gains the new methods). Log text "non-session span ended".

**Internal SessionAPI tightening:** `setGlobalSessionManager`'s parameter and `getSpanSessionManager`'s return type tighten from `SpanSessionManager` to `SpanSessionManagerInternal`. Method names are unchanged; the public-facing `SessionAPIInstance` continues to expose only the public `SpanSessionManager` surface.

**Backwards compat (deprecated forwarders on `SpanSessionManager`):** `getSessionId`, `getPreviousSessionId`, `getSessionStartTime`, `getSessionSpan`, `currentSessionAsReadableSpan` always return `null`. `startSessionSpan` is a no-op (the spec has no public start API). `endSessionSpan` forwards to `endUserSession`. `addSessionStartedListener` / `addSessionEndedListener` are no-ops with no-op unsubscribers. `endSessionSpanInternal` is removed (replaced by internal `endSessionPartInternal`).

## Cross-tab safety

- Part finalization isolates a throwing `setAttributes` from `span.end()` so a poisoned permanent-property value can't drop the part span from the export.
- Peer-tab user-session rollover doesn't stamp the dying tab's inactivity deadline on the new session's storage row, and clears the dying max-duration timer.
- Cleared storage is honored: a part-end persist won't resurrect in-memory state after the row was wiped externally. The manager tracks "have we persisted yet?" locally so the rule fires only after a successful initial write.
- Storage I/O failures: the user session itself continues in-memory via `NamespacedStorage`'s sticky `_writeDisabled` (telemetry keeps flowing). Property writes (both `addProperty` permanent and user-session-scoped) are **rejected outright** when storage is unavailable, a value the caller can set but can't persist would silently diverge from what other tabs and the next page load see.
- `endUserSession` has a 5-second anti-spam cooldown anchored on a persisted timestamp (`embrace_last_end_user_session_ts`), so a refresh-and-click loop cannot manufacture session boundaries any faster than the in-tab gap.
- Tests cover cross-tab races, storage failures, BFCache disengagement, and the full peer-rollover cascade.

## Inactivity deadline

- **Anchored to last user input, not part end.** Previously `inactivityDeadlineTs` was anchored to `partEndTs`, so a user who idled in-page before disengaging got a longer effective grace than one who used the page right up to disengage. Now anchored to `_lastActivityTs`, the total tolerance from the last real input (mouse / keyboard / scroll) is exactly `inactivityTimeoutSeconds`, regardless of disengage timing.
- **Continuously persisted while a part is active.** The deadline is seeded at part start to `now + inactivityTimeoutSeconds * 1000` and refreshed on each throttled activity tick via `_refreshInactivityDeadline`. A browser crash, force-quit, or tab discard that skips `pagehide` now leaves a current deadline in storage so the next page load expires the session correctly. Previously the row held `inactivityDeadlineTs = null` during an active part, so a crash mid-idle silently resumed a stale session.
- **Engagement reactivation counts as activity.** The defensive `engaged && active` branch of `_handleEngagementTransition` (reachable on BFCache restores that leave the part flag set, or on redundant browser focus/visibility events) refreshes the deadline and re-arms the in-page inactivity timer so the disengaged interval doesn't count.
- **Listener firing order.** Session-part ended listeners now fire inside the part-end `finally` block, after end attributes are stamped on the span and before `span.end()`. Subscribers can read `emb.session_part_end_reason` synchronously and emit any final spans/logs while the part is conceptually still active.

## Reason enums

Cross-platform values (shared with mobile) are unprefixed: `manual`, `inactivity`, `max_duration_reached`, `init`, `user_session_rollover`, `user_session_ended`. Web-specific values take a `web_` prefix: `web_foreground`, `web_background`, `web_activity`.

- `emb.session_part_start_reason`: `init`, `web_foreground`, `web_activity`, `user_session_rollover`.
- `emb.session_part_end_reason`: `web_background`, `inactivity`, `user_session_ended`.
- `emb.user_session_termination_reason` (stamped on the final part only): `manual` (public `endUserSession` call), `max_duration_reached` (max-duration timer), `inactivity` (inactivity timer).

### Behavior matrix

| User behavior | Part START | Part END | User session END |
|---|---|---|---|
| Hard nav (new page) | `init` | | |
| Hard nav (old page) | | `web_background` | |
| Soft nav | | | |
| Tab background (no unload) | | `web_background` | |
| Tab foreground (no input yet) | `web_foreground` | | |
| Manual end (engaged) | `user_session_rollover` | `user_session_ended` | `manual` |
| Manual end (disengaged) | | `user_session_ended` (if part active) | `manual` |
| Inactivity timer | next input → `web_activity` | `inactivity` (ts at last activity) | `inactivity` |
| Max duration (engaged) | `user_session_rollover` | `user_session_ended` | `max_duration_reached` |

`_onEngagementChange` is split into explicit per-event handlers (`_onVisibilityChange`, `_onFocus`, `_onBlur`, `_onPageShow`, `_onPageHide`) so each event's semantics are documented at the call site, with `pageshow`/`pagehide` receiving the `PageTransitionEvent` and recording `event.persisted` in the debug log. All five delegate to one shared `_handleEngagementTransition` because the outcome is keyed on `isTabEngaged()` and whether a part is currently active.

## Testing

- `npm run check`
- Full unit suite from `packages/web-sdk/`: 783 passing, 0 failing. Fixed 5 stragglers in `NamespacedStorage.test.ts` that the `f77356c1 "safer prefix"` source change (single-`_` → double-`__` namespace separator) plus its alignment commit `960083e2` had left behind: contract-level assertions now use the `prefix__` literal that matches the source, and priming-style setups go through `storage.setItem` so they don't recouple to the separator.


packages/web-sdk/src/api-sessions/manager/types.ts
```diff
@@ -1,53 +1,168 @@
...
+   * call site.
+   */
+  getUserSessionId: () => string | null;
+  getPreviousUserSessionId: () => string | null;
+  getUserSessionStartTime: () => number | null;
 
-  getSessionSpan: () => ExtendedSpan | null;
+  /**
+   * Ends the current user session. The next foreground session part will
+   * begin a new user session. There is no public start API per spec 1.4;
```
jpmunz: we should avoid references like spec 1.4 since this isn't publicly documented and instead just explain the reasoning here

packages/web-sdk/src/api-sessions/manager/types.ts
```diff
@@ -1,53 +1,168 @@
...
+   * starting is an internal consequence of the next foreground part.
```
jpmunz: to clarify the wording here, so is it the case that if you manually end a user session while the user is engaged with the app that a new user session will then start immediately?

packages/web-sdk/src/api-sessions/manager/types.ts
```diff
@@ -1,53 +1,168 @@
...
+  /**
+   * @deprecated The session part span is no longer part of the public API
+   * and this method always returns `null`. Use getUserSessionId() for user
+   * session identity.
+   */
+  getSessionSpan: () => ExtendedSpan | null;
+  /**
+   * @deprecated **Behavior changed**: forwards to
+   * `addUserSessionStartedListener`. Fires on user-session boundaries
+   * (max-duration rollover, manual `endUserSession()`, lazy inactivity
```
jpmunz: this explanation should probably go in the description of `addUserSessionStartedListener` rather than the deprecation notice here. I don't fully understand the "lazy inactivity detection on the next part start, clock-jump-backwards detection, or  cross-tab termination cascade)" part

packages/web-sdk/src/api-sessions/manager/types.ts
```diff
@@ -1,53 +1,168 @@
...
    */
-  currentSessionAsReadableSpan: (
-    reason: ReasonSessionEnded,
-  ) => ReadableSpan | null;
-
   addSessionStartedListener: (listener: () => void) => () => void;
-
+  /**
+   * @deprecated **Behavior changed**: forwards to
+   * `addUserSessionEndedListener`. Fires on user-session boundaries
```
jpmunz: similar

packages/web-sdk/src/instrumentations/EmbraceInstrumentationBase/EmbraceInstrumentationBase.ts
```diff
@@ -39,13 +41,14 @@ export abstract class EmbraceInstrumentationBase<
...
     this._perf = perf ?? new OTelPerformanceManager();
     this._limitManager = limitManager;
-    this._sessionManager = session.getSpanSessionManager();
     this._logManager = log.getLogManager();
   }
 
-  /* Returns session provider */
-  protected get sessionManager(): SpanSessionManager {
-    return this._sessionManager;
+  // Falls back to the session proxy's current delegate so instrumentations
```
jpmunz: do we have cases where instrumentations are constructed before initSDK?

packages/web-sdk/src/instrumentations/session/SpanSessionBrowserActivityInstrumentation/SpanSessionBrowserActivityInstrumentation.ts
```diff
@@ -1,84 +1,199 @@
...
+      this._onPartEnded,
+    );
+    for (const event of this._events) {
+      this._target.addEventListener(event, this._onActivityThrottled);
+    }
+    this._visibilityDoc.addEventListener?.(
+      'visibilitychange',
+      this._onEngagementChange,
+    );
+    this._target.addEventListener('focus', this._onEngagementChange);
```
jpmunz: I think we should allow a configuration to control whether focus+blur + pageshow/pagehide trigger engagement changes, I'd like to discuss that one in particular a bit more since its new behaviour and may be something where we'd want to flexibility to control with remote configuration

packages/web-sdk/src/instrumentations/session/SpanSessionBrowserActivityInstrumentation/SpanSessionBrowserActivityInstrumentation.ts
```diff
@@ -1,84 +1,199 @@
...
+    this._clearPartInactivityTimer();
   };
 
-  private readonly _onInactivity = () => {
-    this._diag.debug('Inactivity detected');
-    if (this._activityTimeout) {
-      clearTimeout(this._activityTimeout);
+  private readonly _onActivity = (): void => {
+    try {
+      if (!this._isEngaged()) {
```
jpmunz: would it be possible to have activity but not have the page be considered engaged?

packages/web-sdk/src/instrumentations/session/SpanSessionBrowserActivityInstrumentation/SpanSessionBrowserActivityInstrumentation.ts
```diff
@@ -1,84 +1,199 @@
...
+      }
+      if (engaged && active) {
+        this._armPartInactivityTimer();
+      }
+    } catch (e) {
+      this._diag.warn('Error handling engagement change', e);
+    }
+  };
+
+  private _armPartInactivityTimer(): void {
```
jpmunz: should we just call this `_startPartInactivityTimer`? I find the use of `arm` a bit confusing since we don't have that term otherwise in the repo

packages/web-sdk/src/instrumentations/session/SpanSessionBrowserActivityInstrumentation/SpanSessionBrowserActivityInstrumentation.ts
```diff
@@ -1,84 +1,199 @@
...
+      this._target.removeEventListener(event, this._onActivityThrottled);
+    }
+    this._visibilityDoc.removeEventListener?.(
+      'visibilitychange',
+      this._onEngagementChange,
+    );
+    this._target.removeEventListener('focus', this._onEngagementChange);
+    this._target.removeEventListener('blur', this._onEngagementChange);
+    this._target.removeEventListener('pageshow', this._onEngagementChange);
+    this._target.removeEventListener('pagehide', this._onPageHide);
```
jpmunz: why does page hide need a special case here, could it just call `_onEngagementChange`?

packages/web-sdk/src/managers/EmbraceSpanSessionManager/constants.ts
```diff
@@ -1 +1,33 @@
...
+// Holds the absolute timestamp at which the active user session expires from
+// inactivity (`part_end_ts + inactivityTimeoutMs`). Written on every part-end
+// that does NOT terminate the user session; cleared whenever a part starts
+// (the session is engaged again) or the user session ends. Stored in its own
+// key so part-end writes never clobber a peer tab's pn bump in the main
+// state key.
+export const EMBRACE_USER_SESSION_INACTIVITY_DEADLINE_KEY =
+  'embrace_user_session_inactivity_deadline';
+
+// 12 hours in milliseconds (default max user session duration)
```
jpmunz: I'd leave the specific times out of the comments here since they could drift with the actual values set

packages/web-sdk/src/managers/EmbraceTraceManager/EmbraceTraceManager.ts
```diff
@@ -28,9 +32,20 @@ export class EmbraceTraceManager implements TraceManager {
...
+    // Default parent is the active session-part span so customer-created
+    // spans are correlated under the part in the trace tree. Explicit
+    // parentSpan / context still take precedence so callers can override.
+    let activeContext: Context | undefined;
+    if (options.parentSpan) {
+      activeContext = trace.setSpan(context.active(), options.parentSpan);
+    } else if (ctx) {
+      activeContext = ctx;
+    } else {
+      const partSpan = this._spanSessionManager?.getSessionPartSpan();
```
jpmunz: I'd leave this out of this PR, parenting each span to the session span I think will have downstream implications that we should discuss separately

packages/web-sdk/src/processors/IdentifiableSessionLogRecordProcessor/IdentifiableSessionLogRecordProcessor.ts
```diff
@@ -1,39 +1,45 @@
...
 
   public onEmit(logRecord: SdkLogRecord) {
-    logRecord.setAttributes({
-      [ATTR_LOG_RECORD_UID]: generateUUID(),
-      [ATTR_SESSION_ID]: this._spanSessionManager.getSessionId(),
-      [ATTR_SESSION_PREVIOUS_ID]:
-        this._spanSessionManager.getPreviousSessionId(),
-    });
+    try {
+      logRecord.setAttributes({
```
jpmunz: could still do this in a single `setAttributes` call

packages/web-sdk/src/sdk/initSDK.ts
```diff
@@ -316,6 +321,18 @@ export const initSDK = (
 
     diagLogger.info('successfully initialized the SDK');
 
+    // SDKControl.session is always a ProxySpanSessionManager so deprecated
```
jpmunz: I'm not sure I follow the need here, what would be the downside if `spanSessionManager` was just used directly as before?

packages/web-sdk/src/processors/UserSessionLogRecordProcessor/UserSessionLogRecordProcessor.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  public onEmit(logRecord: SdkLogRecord) {
+    try {
+      const userSessionId = this._spanSessionManager.getUserSessionId() ?? '';
+      const userSessionPreviousId =
+        this._spanSessionManager.getPreviousUserSessionId() ?? '';
+      const userSessionIdOverride =
+        this._spanSessionManager.getUserSessionIdOverride();
+      // Spec 2.1: events without a part id must not have a session id.
+      const hasPartId = !!logRecord.attributes[KEY_EMB_SESSION_PART_ID];
```
jpmunz: I think this would an argument in favour of combining with `IdentifiableSessionLogRecordProcessor` and having a single processor handle both sets of tagging, this would rely on `IdentifiableSessionLogRecordProcessor` having run first correct?

packages/web-sdk/src/processors/UserSessionLogRecordProcessor/UserSessionLogRecordProcessor.ts
```diff
@@ -0,0 +1,101 @@
...
+
+      logRecord.setAttributes({
+        [KEY_EMB_USER_SESSION_ID]: hasPartId ? userSessionId : '',
+        [KEY_EMB_USER_SESSION_PREVIOUS_ID]: hasPartId
+          ? userSessionPreviousId
```
jpmunz: it might be valid to tag previous user session id even without a part id

packages/web-sdk/src/processors/UserSessionLogRecordProcessor/UserSessionLogRecordProcessor.ts
```diff
@@ -0,0 +1,101 @@
...
+          : '',
+      });
+
+      if (logRecord.attributes[ATTR_SESSION_ID] === undefined) {
```
jpmunz: I would try and do everything in a single `setAttributes` call in the processor

packages/web-sdk/src/processors/UserSessionSpanProcessor/UserSessionSpanProcessor.ts
```diff
@@ -0,0 +1,162 @@
...
+  // max-duration rollover) clears the manager's in-memory state before the
+  // span finalizes; without a snapshot, onEnd would overwrite the legitimate
+  // attribution with empty strings.
+  public onStart(span: Span): void {
+    try {
+      const userSessionId = this._spanSessionManager.getUserSessionId() ?? '';
+      const userSessionPreviousId =
+        this._spanSessionManager.getPreviousUserSessionId() ?? '';
+      const hasPartId = this._spanSessionManager.getSessionPartId() !== null;
+      span.setAttribute(
```
jpmunz: replace with a single setAttributes call

packages/web-sdk/src/processors/UserSessionSpanProcessor/UserSessionSpanProcessor.ts
```diff
@@ -0,0 +1,162 @@
...
+            : '';
+
+      span.attributes[KEY_EMB_USER_SESSION_ID] = userSessionId;
+      span.attributes[KEY_EMB_USER_SESSION_PREVIOUS_ID] = userSessionPreviousId;
+
+      if (span.attributes[ATTR_SESSION_ID] === undefined) {
+        if (userSessionIdOverride !== null) {
+          span.attributes[ATTR_SESSION_ID] = userSessionIdOverride;
+        } else {
+          span.attributes[ATTR_SESSION_ID] = userSessionId;
```
jpmunz: minor, it took me a bit to confirm this was the same logic as `UserSessionLogRecordProcessor` (e.g. only include if there is a part and user session id but not override Id), it might be worth a helper that could be shared between the processors that take in the various raw values for session id/part id/override, etc. and returns an object with the attributes set correctly

packages/web-sdk/src/api-sessions/manager/types.ts
overbalance: Definitely. The markers were included only for reviewers and will be removed

packages/web-sdk/src/api-sessions/manager/types.ts
overbalance: That's right. Ending a user session will immediately start a new user session.

packages/web-sdk/src/api-sessions/manager/types.ts
overbalance: After speaking offline, this will likely be removed pending approval from the internal team

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -74,50 +140,363 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
+
+  public getSessionPartStartTime(): HrTime | null {
+    return this._activeSessionPartStartTime;
+  }
+
+  public getSessionPartSpan(): ExtendedSpan | null {
+    return this._sessionPartSpan;
+  }
+
+  public startSessionPart(reason: SessionPartStartReason = 'init'): void {
```
jpmunz: maybe default to 'unknown' to avoid conflating this with the start in initSDK if we forget to include the reason in some calls

packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
```diff
@@ -74,50 +140,363 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
+
+  public startSessionPart(reason: SessionPartStartReason = 'init'): void {
+    if (this._sessionPartSpan) {
+      this.endSessionPartInternal('manual');
+    }
+
+    // Parts are foreground-only: a part corresponds to a tab that is both
+    // visible AND the most recently focused window.
+    if (
+      getVisibilityState(this._visibilityDoc) === EMB_STATES.Background ||
```
jpmunz, line 304: similar to comment in `SpanSessionBrowserActivityInstrumentation` I think we should try and keep the focus part of the criteria controlled through a configuration parameter

packages/web-sdk/src/instrumentations/EmbraceInstrumentationBase/EmbraceInstrumentationBase.ts
overbalance: good call, there isn't a current case for this (subclassing by users isn't possible). and I just noticed the fallback would also misroute under `registerGlobally: false`. reverted this change in a new commit so the getter doesn't stay no-op forever

packages/web-sdk/src/instrumentations/session/SpanSessionBrowserActivityInstrumentation/SpanSessionBrowserActivityInstrumentation.ts
overbalance: Simultaneous parts aren't currently supported by the dash and I'd say is out of scope for this project. A local option for developers does seem useful though to prevent noisy focus events when using DevTools

packages/web-sdk/src/instrumentations/session/SpanSessionBrowserActivityInstrumentation/SpanSessionBrowserActivityInstrumentation.ts
overbalance: I would need to test but my instinct is that non-user activity (a script auto-scrolling a hidden window) would not fire a visibilitychange event or activate `document.hasFocus`

packages/web-sdk/src/instrumentations/session/SpanSessionBrowserActivityInstrumentation/SpanSessionBrowserActivityInstrumentation.ts
overbalance: agreed, this was an `init` call in a previous iteration. I'll update.

packages/web-sdk/src/instrumentations/session/SpanSessionBrowserActivityInstrumentation/SpanSessionBrowserActivityInstrumentation.ts
overbalance: That was only for more specific messaging but it's not different enough to warrant the complexity. Removed.

packages/web-sdk/src/managers/EmbraceSpanSessionManager/constants.ts
overbalance: I disagree but will change it if you like. I think it's convenient so you don't have to do the math in your head

packages/web-sdk/src/processors/IdentifiableSessionLogRecordProcessor/IdentifiableSessionLogRecordProcessor.ts
overbalance: sounds good. if either one does throw and is missing, the entire log will be rejected by the BE

packages/web-sdk/src/sdk/initSDK.ts
overbalance: you're right, mirroring those again has no benefit. removed

packages/web-sdk/src/processors/UserSessionLogRecordProcessor/UserSessionLogRecordProcessor.ts
overbalance: separating the concerns here isn't worth the complexity. combined, thanks

packages/web-sdk/src/processors/UserSessionLogRecordProcessor/UserSessionLogRecordProcessor.ts
overbalance: it's not. the spec calls for "" explicitly

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -74,50 +140,363 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
       });
     this._perf = perf ?? new OTelPerformanceManager();
+    this._storage =
+      storage ?? new EmbraceStorage(window.localStorage, this._diag);
     this._visibilityDoc = visibilityDoc;
-    this._storage = storage;
-    this._limitManager = limitManager;
-    this._tracer = trace.getTracer('embrace-web-sdk-sessions');
+    this._limitManager = limitManager ?? null;
+    this._tracer = trace.getTracer('SpanSessionBrowserActivityInstrumentation');
```
jpmunz: tracer name incorrect here

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -74,50 +140,363 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
+    this._activeSessionPartId = activeSessionPartId;
+    let span: EmbraceExtendedSpan;
+    try {
+      span = new EmbraceExtendedSpan(
+        this._tracer.startSpan('emb-session', { attributes }),
+      );
     } catch (error) {
-      this._diag.warn('Error loading permanent session properties', error);
+      this._activeSessionPartId = null;
+      throw error;
```
jpmunz: we shouldn't rethrow here since that could end up affecting the user's application

packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
```diff
@@ -74,50 +140,363 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
+    }
+
+    // Spec 1.1: the inactivity deadline is computed from the part end, not
+    // from after the SpanProcessor.onEnd chain completes.
+    const partEndTs = this._perf.getNowMillis();
+    const span = this._sessionPartSpan;
+    try {
+      const endAttrs = this._endSessionPartAttributes(reason);
+      if (terminationInfo) {
+        endAttrs[KEY_EMB_IS_FINAL_SESSION_PART] = 1;
```
jpmunz, line 377: shouldn't this be based on `terminationInfo.isFinal` specifically?

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -6,66 +6,132 @@ import type {
...
+  private readonly _userSessionStartedListeners: Array<() => void> = [];
+  private readonly _userSessionEndedListeners: Array<() => void> = [];
+  private _lastEndUserSessionTs: number | null = null;
+  private _inEndUserSession = false;
+  private _userSessionIdOverride: string | null = null;
+  // User-session-scoped properties added via addProperty without
+  // lifespan: 'permanent'. Keys are already prefixed with
+  // KEY_PREFIX_EMB_PROPERTIES and length-limited. Cleared on user-session
+  // end. Permanent properties live in localStorage instead.
+  private readonly _userSessionProperties: Map<string, string> = new Map();
```
jpmunz: having these in memory doesn't quite work since a single user session could cross tabs, e.g. if a user closed all browser tabs and then re-opened a few minutes later non-permanent attributes would still be valid for that same user session

jpmunz: one simplification we could make is only supporting permanent properties for now and then revisit supporting user-session-scoped ones

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -126,213 +505,595 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
-    return this._previousSessionId;
+    return this.getPreviousUserSessionId();
   }
 
+  public getSessionStartTime(): HrTime | null {
+    const ms = this.getUserSessionStartTime();
+    if (ms === null) {
+      return null;
+    }
+    const seconds = Math.floor(ms / 1000);
```
jpmunz: there is `millisToHrTime` that could be used here: https://github.com/embrace-io/embrace-web-sdk/blob/31868891e9c885f6b478bcab923282ee8e5a7895/packages/web-sdk/src/utils/PerformanceManager/OTelPerformanceManager.ts#L15

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -126,213 +505,595 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
+
+  public addSessionEndedListener(listener: () => void): () => void {
+    return this.addUserSessionEndedListener(listener);
   }
 
-  public startSessionSpan(options?: StartSessionOptions) {
-    // if there is a session already in progress, end it first
-    if (this._sessionSpan) {
-      this.endSessionSpanInternal('manual');
+  private readonly _onStorage = (event: StorageEvent): void => {
```
jpmunz: I think worth a sync to review this part and below, I want to get a better handle on why the cross tab storage syncing is needed, if there's at most 1 active session part couldn't that part be checking for user session termination due to inactivity/max duration/manual? And then every other part would just need to check on their start rather than being kept in sync?

packages/web-sdk/src/managers/EmbraceSpanSessionManager/constants.ts
jpmunz: you could introduce constants for the readability while keeping the value just in code, e.g.:

```
export const DEFAULT_USER_SESSION_MAX_DURATION_MS = 12 * HOURS_IN_MILLISECONDS;
```

packages/web-sdk/src/processors/UserSessionSpanProcessor/UserSessionSpanProcessor.ts
overbalance: good idea. I made a helper and eliminated 80 lines

packages/web-sdk/src/api-sessions/manager/types.ts
overbalance: Confirmed and has been removed

packages/web-sdk/src/api-sessions/manager/types.ts
overbalance: Removed, see above

packages/web-sdk/src/managers/EmbraceSpanSessionManager/constants.ts
overbalance: I'll remove the redundant comments like you suggested to limit scope

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
overbalance: renamed to startSessionPartInternal and the default reason was removed

packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
overbalance: refactored to rely on user session end, terminationInfo was redundant

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
overbalance: at the beginning of the project there could be more than one active part, but that's no longer the case. I ripped out all the storage event code

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
overbalance: we already have the state container in storage. I'll add support to put properties in there as well, mirroring what already happens with ids - including the same warnings about persistence you mentioned

packages/web-sdk/src/api-sessions/manager/ProxySpanSessionManager/ProxySpanSessionManager.ts
```diff
@@ -1,81 +1,81 @@
...
   }
-
-  public startSessionSpan(options?: StartSessionOptions) {
-    this.getDelegate().startSessionSpan(options);
+  getSessionStartTime(): HrTime | null {
+    const ms = this.getUserSessionStartTime();
+    if (ms === null) {
+      return null;
+    }
+    const seconds = Math.floor(ms / 1000);
```
jpmunz: can millisToHrTime be used here?

jpmunz: also isn't this already handled by `web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts`, should the proxy just call `this._spanSessionManager.getSessionStartTime()`? Same question for the other deprecated methods

packages/web-sdk/src/instrumentations/session/SpanSessionBrowserActivityInstrumentation/SpanSessionBrowserActivityInstrumentation.ts
jpmunz: makes sense, but ya agreed let's keep it separate from this and file a follow-up

packages/web-sdk/src/sdk/initSDK.ts
```diff
@@ -322,7 +328,7 @@ export const initSDK = (
       },
       log: embraceLogManager,
       trace: embraceTraceManager,
-      session: spanSessionManager,
+      session: registerGlobally ? session : spanSessionManager,
```
jpmunz: what was the need for this change?

packages/web-sdk/src/utils/EmbraceStorage/EmbraceStorage.ts
```diff
@@ -0,0 +1,137 @@
...
+  public isWriteDisabled(): boolean {
+    return this._writeDisabled;
+  }
+
+  /**
+   * Returns the underlying-storage key for a logical key. Browser `storage`
+   * events fire with the underlying key, so callers that filter those events
+   * need to compare against this value rather than the logical key.
+   */
+  public getStorageEventKey(logicalKey: string): string {
```
jpmunz: no longer used

packages/web-sdk/src/processors/UserSessionLogRecordProcessor/UserSessionLogRecordProcessor.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  public forceFlush(): Promise<void> {
+    return Promise.resolve(undefined);
+  }
+
+  public onEmit(logRecord: SdkLogRecord) {
+    try {
+      logRecord.setAttribute(ATTR_LOG_RECORD_UID, generateUUID());
+      applyUserSessionAttributesToLogRecord(
```
jpmunz: I think it's enough that the helper defines `createUserSessionAttributes`, then instead of having `applyUserSessionAttributesToLogRecord` we're able to use a single setAttributes call here:
```
logRecord.setAttributes({ATTR_LOG_RECORD_UID, generateUUID()}, ...createUserSessionAttributes(logRecord.attributes, this._spanSessionManager));
```

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -74,50 +116,303 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
+          `falling back to default inactivity timeout (${(DEFAULT_USER_SESSION_INACTIVITY_TIMEOUT_MS / 1000).toString()}s).`,
+      );
+      this._inactivityTimeoutMs = DEFAULT_USER_SESSION_INACTIVITY_TIMEOUT_MS;
+    }
   }
 
-  // Collects all permanent session properties from localStorage
-  private _getPermanentAttributes(): Attributes {
-    const permanentAttributes = new Map();
+  public dispose(): void {
```
jpmunz: looks like this is only used in tests, is that expected?

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -126,213 +421,523 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
       }
     }
-  }
 
-  public removeProperty(propertyKey: string) {
-    if (!this._sessionSpan) {
-      this._diag.debug(
-        'trying to remove a session property, but there is no session in progress. This is a no-op.',
+    if (this._sessionPartSpan) {
+      this._sessionPartSpan.setAttribute(
```
jpmunz: is this needed since we write them on span end as well?

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -126,213 +421,523 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
     }
 
-    this._activeSessionId = generateUUID();
-    this._activeSessionStartTime = this._perf.getNowHRTime();
-    this._activeSessionCounts = {};
-    const previouslyRecordedCounts = this._nextSessionCounts;
-    this._nextSessionCounts = {};
+    // Merge the persisted snapshot (which may include values written by
+    // another tab on the same user session) with anything queued in this
+    // tab since the last part end. Local writes win per key.
```
jpmunz: shouldn't non-local writes win here? Thinking of a case where this tab (Tab1) was inactive and had something like {foo: "tab1"}, if another tab (Tab2) was active and the property was updated to {foo: "tab2"} and then Tab2 is hidden and Tab1 is refocused then we would want {foo: "tab2"} to be the final value see on this part start right?

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -126,213 +421,523 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
-        listener();
-      } catch (error) {
-        this._diag.warn(
-          'Error while executing session started listener',
-          error,
-        );
-      }
+    // Spec 1.3: the user-session max-duration timer must fire even while
+    // no part is active.
+    this._setupMaxDurationTimer(updated, this._perf.getNowMillis());
```
jpmunz: isn't the existing max duration timer still valid? Why clear and restart it in `_continueUserSessionAfterPartEnd` ?

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -126,213 +421,523 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
+  private _continueUserSessionAfterPartEnd(partEndTs: number): void {
+    if (!this._state) {
+      this._clearMaxDurationTimer();
+      return;
+    }
 
-    this._coldStart = false;
+    const updated: UserSessionState = {
+      ...this._state,
+      inactivityDeadlineTs: partEndTs + this._state.inactivityTimeoutMs,
```
jpmunz: should there be an exception in the case that the session part end reason was "inactivity"? In that case wouldn't we be waiting twice the inactivity time to end the user session? (e.g. there was 30 minute of user inactivity during the part, then an additional 30 minutes of inactivity after the part ended)

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -126,213 +421,523 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
+  ): void {
+    try {
+      this.endSessionPartInternal('user_session_ended', userSessionEndReason);
+    } catch (e) {
+      this._diag.error('Error finalizing part during user session end', e);
+    }
+
+    this._previousUserSessionId = this._state?.userSessionId ?? null;
+    this._state = null;
+    this._clearStoredState();
```
jpmunz: wouldn't this also clear the user session number?

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -126,213 +421,523 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
+    // no part is active.
+    this._setupMaxDurationTimer(updated, this._perf.getNowMillis());
+  }
+
+  /**
+   * Ends the active part as final, clears state, and starts a fresh part
+   * for the next user session. Used by manual endUserSession and
+   * max-duration expiry.
+   */
+  private _terminateUserSession(
```
jpmunz: I'd maybe call this `_rolloverUserSession` to make it clear it always creates a new one and shouldn't be used in the case of ending a user session due to inactivity

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -126,213 +421,523 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
 
-  public addSessionEndedListener(listener: SessionEndedListener) {
-    const listenerIndex = this._sessionEndedListeners.push(listener);
+  private _getPermanentAttributes(): Attributes {
+    const permanentAttributes: Record<string, string> = {};
+    if (this._storage.isWriteDisabled()) {
+      return permanentAttributes;
+    }
+    for (const key of this._storage.keys()) {
+      if (key.startsWith(KEY_PREFIX_EMB_PROPERTIES)) {
```
jpmunz: minor potential follow-up refactor, I wonder if we have EmbraceSpanSessionManager.ts setup a separate `this._storage` instance for the permanent properties that is namespaced, then we can take advantage of the EmbraceStorage methods that already take namespace into account for these operations

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
```diff
@@ -126,213 +421,523 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
 
-  public addSessionStartedListener(listener: SessionStartedListener) {
-    const listenerIndex = this._sessionStartedListeners.push(listener);
+  private _buildUserSessionAttributes(
+    state: UserSessionState,
+  ): UserSessionAttributes {
+    return {
+      [ATTR_SESSION_ID]: this._userSessionIdOverride ?? state.userSessionId,
+      [KEY_EMB_USER_SESSION_ID]: state.userSessionId,
+      [KEY_EMB_USER_SESSION_NUMBER]: state.userSessionNumber,
```
jpmunz: is UserSessionSpanProcessor.ts needed since these attributes are all being set here on the session part span anyway?

packages/web-sdk/src/sdk/initSDK.ts
overbalance: removes one level of proxy indirection but not worth the churn. removed

packages/web-sdk/src/utils/EmbraceStorage/EmbraceStorage.ts
overbalance: missed that because it still had a test

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
overbalance: yes, only for tests. I refactored to expose the actual function and marked as `@internal`

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
overbalance: This is a holdover from allowing reads of a span that is not ended. OTel discourages using spans as stores so it's a bit of an anti-pattern already. I recommend leaving for now to maintain current functionality, but definitely something to follow up on

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
overbalance: I removed the entire function. If partStartTime is requested, we can add later

packages/web-sdk/src/api-sessions/manager/ProxySpanSessionManager/ProxySpanSessionManager.ts
overbalance: yes, I switched it back to delegating properly. poor refactor

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
overbalance: you're right. I inverted it

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
overbalance: I was trying to get it to clean up after itself by not firing in an unfocused tab but youre right it just restarts with the same time left on the clock. dropped

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
overbalance: yes, I designed it that way. but I'm going to end both at 30m instead. change incoming

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
overbalance: Yes, it stamps all non-session-part spans on start/end. The name wasn't very precise so I renamed it to IdentifiableSessionSpanProcessor to match logs

packages/web-sdk/src/utils/NamespacedStorage/NamespacedStorage.ts
```diff
@@ -36,24 +38,36 @@ export class NamespacedStorage implements Storage {
     this._diag =
       diagParam ??
       diag.createComponentLogger({ namespace: 'NamespacedStorage' });
-    this._keyPrefix = namespace ? `${namespace}_` : '';
+    this._keyPrefix = namespace ? `${namespace}__` : '';
```
overbalance, line 41: Switched to the clearer double `__` prefix so startsWith has less potential false finds

packages/web-sdk/src/api-sessions/manager/ProxySpanSessionManager/ProxySpanSessionManager.ts
```diff
@@ -41,41 +46,104 @@ export class ProxySpanSessionManager implements SpanSessionManager {
...
-    this.getDelegate().endSessionSpanInternal(reason);
-  }
-
-  public currentSessionAsReadableSpan(
-    reason: ReasonSessionEnded,
-  ): ReadableSpan | null {
-    return this.getDelegate().currentSessionAsReadableSpan(reason);
+  /** @deprecated Will be removed in a future major version, always returns null. */
+  public currentSessionAsReadableSpan(): null {
+    return null;
```
jpmunz: these could continue to call the delegate rather than re-implementing the no-op

packages/web-sdk/src/processors/UserSessionLogRecordProcessor/UserSessionLogRecordProcessor.ts
```diff
@@ -25,11 +30,18 @@ export class IdentifiableSessionLogRecordProcessor
...
+    const userSessionId = this._spanSessionManager.getUserSessionId() ?? '';
+    const previousUserSessionId =
+      this._spanSessionManager.getPreviousUserSessionId() ?? '';
+
     logRecord.setAttributes({
       [ATTR_LOG_RECORD_UID]: generateUUID(),
-      [ATTR_SESSION_ID]: this._spanSessionManager.getSessionId(),
-      [ATTR_SESSION_PREVIOUS_ID]:
-        this._spanSessionManager.getPreviousSessionId(),
+      [ATTR_SESSION_ID]: userSessionId,
```
jpmunz, line 37: should this be the user specified session id if set?

packages/web-sdk/src/sdk/initSDK.ts
```diff
@@ -149,14 +150,16 @@ export const initSDK = (
       );
     }
 
-    const namespace = !registerGlobally && appID ? appID : '';
+    const namespace = appID ?? undefined;
```
jpmunz, line 157: why the change here, would we still only want it namespaced if `!registerGlobally` as before?

packages/web-sdk/src/utils/NamespacedStorage/NamespacedStorage.ts
jpmunz: I'm not sure this is worth the change without a migration, all current namespaced keys in storage would get lost on an sdk update wouldn't they?

packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
```diff
@@ -142,196 +462,468 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
+      // persist would silently diverge from what other tabs (and the next
+      // page load) see, which is worse than no property at all.
+      if (!storePermanentProperties(this._storage, candidate)) {
+        this._diag.warn(
+          'Storage unavailable; rejecting permanent property write.',
+        );
+        return;
       }
+      this._permanentProperties = candidate;
+      this._removeFromUserSessionProperties(bareKey);
```
jpmunz, line 483: why is the property being removed here?

packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
```diff
@@ -142,196 +462,468 @@ export class EmbraceSpanSessionManager implements SpanSessionManagerInternal {
...
+          error,
+        );
+      }
+    }
+  }
 
-    return () => {
-      this._sessionEndedListeners.splice(listenerIndex - 1, 1);
-    };
+  private _storeState(): boolean {
```
jpmunz, line 893: can be a separate follow-up, I think a few more things here could be moved to `state.ts` or `activity.ts`, like `_storeState` here to keep those operations contained within the helpers

packages/web-sdk/src/managers/EmbraceUserSessionManager/types.ts
```diff
@@ -1,21 +1,163 @@
...
+  /**
+   * Null on the first user session since the first visit or after a
+   * permanent storage clear.
+   */
+  readonly previousUserSessionId: string | null;
+  readonly userSessionStartTs: number;
+  readonly userSessionMaxEndTs: number;
+  readonly userSessionNumber: number;
+  /** Bumped on every part start. */
+  readonly userSessionPartIndex: number;
```
jpmunz, line 33: was there also a requirement to track the number of parts within a given user session?

packages/web-sdk/src/api-sessions/manager/ProxySpanSessionManager/ProxySpanSessionManager.ts
overbalance: Doing it this way saves one level of indirection and having to add the parameter and type import. I'll change it.

packages/web-sdk/src/processors/UserSessionLogRecordProcessor/UserSessionLogRecordProcessor.ts
overbalance: I cut that feature to reduce scope. It's on the list as a follow up.

packages/web-sdk/src/sdk/initSDK.ts
overbalance: Always scoping makes it easier for users to debug locally when switching from prod/dev appIDs on localhost (or any domain that may host more than one appID). And while rare, if someone decides to convert a globally scoped appID to non-global, this keeps from losing total counts.

packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
overbalance: If you had previously set a user session-scoped property and then promoted it to permanent, you would only want a permanent entry. There's only one eviction function so unless we want to add a scoped property manager, I thought this was the best option.

packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
overbalance: agreed. these snuck back in, in a refactor.

packages/web-sdk/src/instrumentations/session/SpanSessionBrowserActivityInstrumentation/SpanSessionBrowserActivityInstrumentation.ts
overbalance: So I tested two windows open side-by-side, not overlapping. If you focus A and then mouseover to B and use the scrollwheel on the mouse, it does not `hasFocus`. The scroll event does fire but our code requires both `visible` conditions. We could change that but we would also have to add support for multiple active parts. If we wanted to do that without adding multi-part support we would likely have to use BroadcastChannel to inform all tabs which tab is designated the primary/active.

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
overbalance: refactored out

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
overbalance: terminate was still hanging around because that's the name of the key. renamed.

packages/web-sdk/src/managers/EmbraceSpanSessionManager/EmbraceSpanSessionManager.ts
overbalance: i like that idea. i'll add it to the list

packages/web-sdk/src/utils/NamespacedStorage/NamespacedStorage.ts
overbalance: Every key we have is new so there isn't anything to migrate.

packages/web-sdk/src/instrumentations/session/SpanSessionBrowserActivityInstrumentation/SpanSessionBrowserActivityInstrumentation.ts
jpmunz: got it, okay I think how it currently is makes sense then

packages/web-sdk/src/api-sessions/manager/ProxySpanSessionManager/ProxySpanSessionManager.ts
jpmunz: this is true of a few of the other calls in here as well

jpmunz (approved):

packages/web-sdk/src/sdk/initSDK.ts
```diff
@@ -102,6 +101,8 @@ export const initSDK = (
     blockNetworkSpanForwarding = false,
     restrictedProtocols = new Set(['file:']),
     useDocumentTitleAsPageLabel = true,
+    maxUserSessionDurationSeconds = 43200,
```
jpmunz: do these have to be set here since we have `DEFAULT_USER_SESSION_MAX_DURATION_SECONDS` and `DEFAULT_ACTIVITY_THROTTLE_MS` to use as the fallback defaults? Or if they do can we have them use the constants here?

packages/web-sdk/src/api-sessions/manager/types.ts
```diff
@@ -1,41 +1,97 @@
...
+   * the current user session. Properties added without `lifespan: 'permanent'`
+   * survive across foreground/background transitions but are cleared when the
+   * user session ends (manual `endUserSession()`, max-duration rollover, or a
+   * cross-tab end). They are persisted alongside the user-session state in
+   * localStorage, so other tabs sharing the same user session pick them up
+   * on their next part start. With `lifespan: 'permanent'` the property is
+   * stored as its own localStorage entry and reapplied across user sessions
+   * until removed via `removeProperty(key)`.
+   *
+   * If localStorage is unavailable the property degrades to in-memory only:
```
jpmunz: I believe this behaviour was changed, it now simply rejects the write altogether in this case right?

packages/web-sdk/src/sdk/initSDK.ts
overbalance: switched to the constant so we can keep these keys optional

PR #1312, opened 2026-05-06
EMBR-11944 chore: fork upstreamed web vitals instrumentation

joaquin-diaz:

Web vitals were captured as span events on the session span using `emb.web_vital.*` attribute names. The upstream OTel browser project now ships a web vitals instrumentation that emits log records using standardized `browser.web_vital.*` attribute names. We want to adopt the OTel implementation as our local fork to align with the emerging standard.

- Forked the OTel browser `WebVitalsInstrumentation` as the new base, switching from `sessionSpan.addEvent()` to `this.logger.emit()` (log records instead of span events)
- Renamed attribute keys: `emb.web_vital.*` → `browser.web_vital.*` for standard fields; `emb.web_vital.attribution.*` prefix retained for Embrace-specific computed extras (LoAF scripts, TTFB sub-parts)
- Event name unified to `browser.web_vital` (metric identified via `browser.web_vital.name` attribute) instead of per-metric names like `emb-web-vitals-report-LCP`
- Log body set to `JSON.stringify(metric.attribution)` — captures same primitive data as the old `webVitalAttributionToReport()` helper without redundancy
- Kept injectable `listeners` pattern for testability (not in OTel upstream)

## Integration test changes
  - Added `SCOPES_WITH_SORTED_COMPARISON` to handle non-deterministic web vitals arrival order in golden comparisons
  - Added `LOGS_WITH_IGNORED_BODY` to skip body comparison for web vitals logs (timing values change each run)
  - Added `browser.web_vital.delta`, `browser.web_vital.id`, `browser.web_vital.value` to ignored attributes list
  - Added bind mounts in `e2e-test.sh` so `:test` picks up local test file changes without rebuilding the container image

### DIFF
--- a/packages/web-sdk/src/instrumentations/loaf/LoafInstrumentation/LoafInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/loaf/LoafInstrumentation/LoafInstrumentation.ts
@@ -214,7 +214,7 @@ export class LoafInstrumentation extends EmbraceInstrumentationBase {
     const attrs: Record<string, string | number> = {
       [KEY_EMB_TYPE]: EMB_TYPES.WebVital,
       ['browser.web_vital.id']: generateWebVitalID(),
-      ['browser.web_vital.name']: 'TBD',
+      ['browser.web_vital.name']: 'tbd',
       ['browser.web_vital.value']: Math.round(this._totalBlockingDuration),
       ['browser.web_vital.rating']: rating,
       [ATTR_TBD_LOAF_TOTAL_DURATION]: Math.round(this._totalDuration),
--- a/packages/web-sdk/src/instrumentations/loaf/LoafInstrumentation/LoafInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/loaf/LoafInstrumentation/LoafInstrumentation.ts
@@ -213,10 +213,10 @@ export class LoafInstrumentation extends EmbraceInstrumentationBase {
 
     const attrs: Record<string, string | number> = {
       [KEY_EMB_TYPE]: EMB_TYPES.WebVital,
-      ['emb.web_vital.id']: generateWebVitalID(),
-      ['emb.web_vital.name']: 'TBD',
-      ['emb.web_vital.value']: Math.round(this._totalBlockingDuration),
-      ['emb.web_vital.rating']: rating,
+      ['browser.web_vital.id']: generateWebVitalID(),
+      ['browser.web_vital.name']: 'TBD',
+      ['browser.web_vital.value']: Math.round(this._totalBlockingDuration),
+      ['browser.web_vital.rating']: rating,
       [ATTR_TBD_LOAF_TOTAL_DURATION]: Math.round(this._totalDuration),
       [ATTR_TBD_LOAF_WORK_DURATION]: Math.round(this._workDuration),
       [ATTR_TBD_LOAF_STYLE_AND_LAYOUT_DURATION]: Math.round(
--- a/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
@@ -14,10 +14,12 @@ import type { PageManager } from '../../../api-page/index.ts';
 import { page } from '../../../api-page/index.ts';
 import type { URLDocument } from '../../../common/index.ts';
 import {
+  EMB_TYPES,
   KEY_APP_SURFACE_LABEL,
   KEY_BROWSER_URL_FULL,
   KEY_EMB_PAGE_ID,
   KEY_EMB_PAGE_PATH,
+  KEY_EMB_TYPE,
 } from '../../../constants/index.ts';
 import { EmbraceInstrumentationBase } from '../../EmbraceInstrumentationBase/index.ts';
 import {
@@ -325,6 +327,7 @@ export class WebVitalsInstrumentation extends EmbraceInstrumentationBase {
       eventName: WEB_VITAL_EVENT_NAME,
       severityNumber: SeverityNumber.INFO,
       attributes: {
+        [KEY_EMB_TYPE]: EMB_TYPES.WebVital,
         [KEY_BROWSER_WEB_VITAL_NAME]: metric.name.toLowerCase(),
         [KEY_BROWSER_WEB_VITAL_VALUE]: metric.value,
         [KEY_BROWSER_WEB_VITAL_DELTA]: metric.delta,
--- a/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
@@ -52,6 +52,13 @@ type AttributedPage = {
 
 const roundClamp = (value: number): number => Math.round(Math.max(0, value));
 
+const isPrimitiveValue = (
+  value: unknown,
+): value is string | number | boolean => {
+  const type = typeof value;
+  return type === 'number' || type === 'string' || type === 'boolean';
+};
+
 const loafScriptsAttribution = (
   metric: MetricWithAttribution,
   diag: DiagLogger,
@@ -349,9 +356,16 @@ export class WebVitalsInstrumentation extends EmbraceInstrumentationBase {
           ? ttfbSubPartsAttribution(metric, this._diag)
           : {}),
       },
-      body: this._includeRawAttribution
-        ? JSON.stringify(metric.attribution)
-        : undefined,
+      body:
+        this._includeRawAttribution && metric.attribution != null
+          ? JSON.stringify(
+              Object.fromEntries(
+                Object.entries(metric.attribution).filter(([, v]) =>
+                  isPrimitiveValue(v),
+                ),
+              ),
+            )
+          : undefined,
       timestamp: millisToHrTime(this._getTimeForMetric(metric)),
     };
 
--- a/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
@@ -1,8 +1,8 @@
-import type {
-  Attributes,
-  AttributeValue,
-  DiagLogger,
-} from '@opentelemetry/api';
+import type { Attributes, DiagLogger } from '@opentelemetry/api';
+import type { LogRecord } from '@opentelemetry/api-logs';
+import { SeverityNumber } from '@opentelemetry/api-logs';
+import { millisToHrTime } from '@opentelemetry/core';
+import { safeExecuteInTheMiddle } from '@opentelemetry/instrumentation';
 import { ATTR_URL_FULL } from '@opentelemetry/semantic-conventions';
 import type {
   CLSMetricWithAttribution,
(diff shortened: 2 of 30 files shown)


packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -222,32 +196,35 @@ export class WebVitalsInstrumentation extends EmbraceInstrumentationBase {
...
     urlAttribution = true,
     pageManager,
-  }: WebVitalsInstrumentationArgs = {}) {
+    applyCustomLogRecordData,
+    ...config
+  }: WebVitalsInstrumentationConfig = {}) {
     super({
-      instrumentationName: 'WebVitalsInstrumentation',
+      instrumentationName: '@opentelemetry/browser-instrumentation/web-vitals',
       instrumentationVersion: '1.0.0',
```
overbalance, line 226: We should probably change this or the instrumentationName so it's apparent when we change over to OTel for real

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -397,53 +317,50 @@ export class WebVitalsInstrumentation extends EmbraceInstrumentationBase {
...
+        [KEY_BROWSER_WEB_VITAL_NAME]: metric.name.toLowerCase(),
+        [KEY_BROWSER_WEB_VITAL_VALUE]: metric.value,
+        [KEY_BROWSER_WEB_VITAL_DELTA]: metric.delta,
+        [KEY_BROWSER_WEB_VITAL_RATING]: metric.rating,
+        [KEY_BROWSER_WEB_VITAL_ID]: metric.id,
+        [KEY_BROWSER_WEB_VITAL_NAVIGATION_TYPE]: metric.navigationType,
+        ...(attributedPage
+          ? {
+              [ATTR_URL_FULL]: attributedPage.fullURL,
+              [KEY_BROWSER_URL_FULL]: attributedPage.fullURL,
```
overbalance, line 346: Should we switch to the newest version from OTel: `browser.document.url.full`?

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/types.ts
```diff
@@ -16,11 +17,18 @@ export type WebVitalListeners = Record<
...
   urlDocument?: URLDocument;
   urlAttribution?: boolean;
   pageManager?: PageManager;
-} & Pick<EmbraceInstrumentationBaseArgs, 'diag' | 'perf'>;
+  diag?: DiagLogger;
+  perf?: PerformanceManager;
+}
+
+// Backward-compat alias
+export type WebVitalsInstrumentationArgs = WebVitalsInstrumentationConfig;
```
overbalance: Why did you decide to keep this? Does it need `@deprecated `?

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
joaquin-diaz: ~Yeah, let's do it~. This is used across other loaders so if we do it we need to update it all at once and first update the loaders to support it. We should do it in a different PR

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/types.ts
joaquin-diaz: No, the API is backward compatible. This is not needed, I removed it

overbalance (approved):

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/types.ts
```diff
@@ -1,11 +1,14 @@
...
   Metric,
   MetricWithAttribution,
   ReportOpts,
 } from 'web-vitals/attribution';
 import type { PageManager } from '../../../api-page/index.ts';
 import type { URLDocument } from '../../../common/index.ts';
-import type { EmbraceInstrumentationBaseArgs } from '../../EmbraceInstrumentationBase/index.ts';
+import type { PerformanceManager } from '../../../utils/index.ts';
 
 export type TrackingLevel = 'core' | 'all';
```
overbalance, line 14:
```suggestion
/** @deprecated All vitals metrics are tracked by default */
export type TrackingLevel = 'core' | 'all';
```

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/attributes.ts
```diff
@@ -1,9 +1,10 @@
-export const KEY_EMB_WEB_VITAL_NAME = 'emb.web_vital.name';
-export const KEY_EMB_WEB_VITAL_VALUE = 'emb.web_vital.value';
-export const KEY_EMB_WEB_VITAL_RATING = 'emb.web_vital.rating';
-export const KEY_EMB_WEB_VITAL_DELTA = 'emb.web_vital.delta';
-export const KEY_EMB_WEB_VITAL_ID = 'emb.web_vital.id';
-export const KEY_EMB_WEB_VITAL_NAVIGATION_TYPE =
-  'emb.web_vital.navigation_type';
+export const KEY_BROWSER_WEB_VITAL_NAME = 'browser.web_vital.name';
```
jpmunz, line 1: as a follow-up we should look at using these updated names in LoafInstrumentation.ts as well

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/types.ts
jpmunz: I think this still implies it can be configured, maybe "All vitals metrics are always tracked"

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/types.ts
```diff
@@ -16,11 +19,27 @@ export type WebVitalListeners = Record<
   ((onReport: WebVitalOnReport, opts?: ReportOpts) => void) | undefined
 >;
 
-export type WebVitalsInstrumentationArgs = {
-  /** @deprecated All vitals metrics are tracked by default */
-  trackingLevel?: TrackingLevel;
+export interface WebVitalsInstrumentationConfig extends InstrumentationConfig {
+  // OTel upstream options
+  /**
+   * @experimental
```
jpmunz, line 26: why experimental?

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -397,53 +320,52 @@ export class WebVitalsInstrumentation extends EmbraceInstrumentationBase {
...
+      attributes: {
+        [KEY_BROWSER_WEB_VITAL_NAME]: metric.name.toLowerCase(),
+        [KEY_BROWSER_WEB_VITAL_VALUE]: metric.value,
+        [KEY_BROWSER_WEB_VITAL_DELTA]: metric.delta,
+        [KEY_BROWSER_WEB_VITAL_RATING]: metric.rating,
+        [KEY_BROWSER_WEB_VITAL_ID]: metric.id,
+        [KEY_BROWSER_WEB_VITAL_NAVIGATION_TYPE]: metric.navigationType,
+        ...(attributedPage
+          ? {
+              [ATTR_URL_FULL]: attributedPage.fullURL,
```
jpmunz: do we need to set both `ATTR_URL_FULL` and `KEY_BROWSER_URL_FULL`?

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -397,53 +320,52 @@ export class WebVitalsInstrumentation extends EmbraceInstrumentationBase {
...
+            }
+          : {}),
+        ...(metric.name === 'INP'
+          ? loafScriptsAttribution(metric, this._diag)
+          : {}),
+        ...(metric.name === 'TTFB'
+          ? ttfbSubPartsAttribution(metric, this._diag)
+          : {}),
+      },
+      body: this._includeRawAttribution
```
jpmunz: I'd like to discuss this change a bit more, the previous `webVitalAttributionToReport` was only including primitive values whereas this would include the entire attribution object which could make the payload quite large for some vital reports. It also seems preferable to include each attribution value as its own attribute on the log, what is the reason for having it all be done on the body?

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/types.ts
joaquin-diaz: I'm carrying it over from the upstream instrumentation. Once we switch for upstream maybe it's no longer experimental. Not sure why it was marked as such in the first place. @overbalance do you know?

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
joaquin-diaz: Not anymore

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/attributes.ts
joaquin-diaz: I added it here, it's the same loader. We'll have to support both for a while

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
joaquin-diaz: Main reason is to reduce cardinality and sending a lot of different attributes depending on what's the web vital.  This is another discussion we had on the SIG there's consensus that this is a good way of doing it. Although we haven't started talking about payload size in there yet. This is also how it's done in the upstreamed version so but I guess we can always use the `_includeRawAttribution` flag if we have concerns about this. Let's discuss it in tomorrow's stand up

packages/web-sdk/src/instrumentations/loaf/LoafInstrumentation/LoafInstrumentation.ts
```diff
@@ -211,10 +211,10 @@ export class LoafInstrumentation extends EmbraceInstrumentationBase {
 
     const attrs: Record<string, string | number> = {
       [KEY_EMB_TYPE]: EMB_TYPES.WebVital,
-      ['emb.web_vital.id']: generateWebVitalID(),
-      ['emb.web_vital.name']: 'TBD',
-      ['emb.web_vital.value']: Math.round(this._totalBlockingDuration),
-      ['emb.web_vital.rating']: rating,
+      ['browser.web_vital.id']: generateWebVitalID(),
+      ['browser.web_vital.name']: 'TBD',
```
overbalance: This is lowercase in WebVitals. Want to change it now?
```suggestion
      ['browser.web_vital.name']: 'tbd',
```

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/types.ts
overbalance: yeah, I would drop "by default"

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/types.ts
overbalance: We weren't sure if people would use it (for your stated concern that it could be large/hard to parse) so marking as experimental lets people know it might have a short life

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/types.ts
```diff
@@ -16,11 +20,27 @@ export type WebVitalListeners = Record<
...
+  // Embrace-specific (testability + SPA page attribution)
   listeners?: WebVitalListeners;
   urlDocument?: URLDocument;
   urlAttribution?: boolean;
   pageManager?: PageManager;
-} & Pick<EmbraceInstrumentationBaseArgs, 'diag' | 'perf'>;
+  diag?: DiagLogger;
+  perf?: PerformanceManager;
+
+  /** @deprecated All vitals metrics are tracked by default */
```
overbalance:
```suggestion
  /** @deprecated All vitals metrics are tracked */
```

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
overbalance: To expand a bit, semconv for every key would need to be explicit. Looping through as we did before would be unusual in OTel.

packages/web-sdk/src/sdk/types.ts
```diff
@@ -363,7 +363,7 @@ export interface DefaultInstrumentationConfig {
   omit?: Set<OptionalInstrumentations>;
   exception?: GlobalExceptionInstrumentationArgs;
   click?: ClicksInstrumentationArgs;
-  'web-vital'?: WebVitalsInstrumentationArgs;
+  'web-vital'?: WebVitalsInstrumentationConfig;
```
overbalance, line 396: Since this is a breaking change, can we rename to `web-vitals` to align with the actual name?

PR #1316, opened 2026-05-09
refactor(storage): make NamespacedStorage handle errors and disable writes safely

overbalance:

`NamespacedStorage` was a thin key-prefix wrapper that let underlying storage exceptions propagate. Quota errors, revoked permissions, and similar environmental failures surfaced to every caller and kept throwing on every subsequent write attempt.

- All underlying storage calls are now guarded internally; reads that fail return `null` and writes that fail are swallowed.
- First write failure flips a sticky disabled flag and emits one error log. Subsequent `setItem` calls become no-ops so a single failure does not keep throwing.
- Public API and constructor signature unchanged. `implements Storage` preserved, so no consumers need to update.

## Testing

- `npm run validate` clean.
- `npm run test` for `NamespacedStorage`: 10 passed, including new coverage for swallowed read/remove failures and the sticky disable-after-first-write-failure behavior.

### DIFF
--- a/packages/web-sdk/src/utils/NamespacedStorage/NamespacedStorage.ts
+++ b/packages/web-sdk/src/utils/NamespacedStorage/NamespacedStorage.ts
@@ -1,21 +1,67 @@
+import type { DiagLogger } from '@opentelemetry/api';
+import { diag } from '@opentelemetry/api';
+
+export interface NamespacedStorageArgs {
+  storage: Storage;
+  namespace?: string;
+  diag?: DiagLogger;
+}
+
+/**
+ * `Storage` wrapper that isolates the SDK from storage errors and
+ * optionally namespaces keys with a prefix.
+ *
+ * Read failures (`getItem`, `key`, `length`) log at warn level.
+ *
+ * The first `setItem` failure puts the wrapper into a sticky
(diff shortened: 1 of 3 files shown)


joaquin-diaz (approved):

packages/web-sdk/src/utils/NamespacedStorage/NamespacedStorage.ts
```diff
@@ -1,58 +1,141 @@
...
+
+interface NamespacedStorageArgs {
+  storage: Storage;
+  diag?: DiagLogger;
+  namespace?: string;
+}
+
+/**
+ * `Storage`-shaped wrapper with safe error handling, an optional key
+ * namespace, and cross-tab helpers. Mutation methods return `true` on
```
jpmunz:
```suggestion
 * namespace. Mutation methods return `true` on
```

jpmunz (approved):

packages/web-sdk/src/utils/NamespacedStorage/NamespacedStorage.ts
overbalance: good catch

jpmunz (approved):

PR #1326, opened 2026-05-13
refactor(instrumentations): extend OTel InstrumentationBase, remove our fork

overbalance:
**Needs more discussion, moving back to draft**

The SDK kept a 203-line vendored copy of OTel's `InstrumentationAbstract` to avoid coupling to "internal and experimental code." That copy hand-rolled constructor-time auto-enable in every subclass and left the enable/disable contract inconsistent (some subclasses tracked live state in `_config.enabled`, others in a private `_isEnabled` field), and `enable()` could run before the per-instance session manager was wired. No customer-visible bug shipped. This swaps the fork for OTel's public `InstrumentationBase` and gives the base one consistent enable/disable contract.

This is the **mechanical half** of a larger change. The opinionated lifecycle policy (capability vetoes, fail-safe rollback, `setConfig` semantics) was split out and is intentionally **not** included here.

- Delete the vendored `InstrumentationAbstract` (203 lines); `EmbraceInstrumentationBase` extends `@opentelemetry/instrumentation`'s `InstrumentationBase` directly.
- Force `enabled: false` into `super()` so OTel's constructor-time auto-enable never fires; defer `enable()` to OTel's `registerInstrumentations` autoLoader, which runs after `setupDefaultInstrumentations` wires the managers — so manager-before-enable ordering is now guaranteed.
- Base owns `enable()`/`disable()`: they guard idempotency, flip `_config.enabled` (the single flag `isEnabled()` reads), and call subclass `_onEnable()` / `_onDisable()` hooks. All 12 subclasses are converted to those hooks; the per-subclass auto-enable tails and ad-hoc `_isEnabled` fields are removed.
- `NavigationInstrumentation` is the one instrumentation not registered through the autoLoader (it's a public singleton), so `getNavigationInstrumentation` enables it explicitly, guarded.
- Capability gaps (LoAF / ElementTiming entry-type support, WebVitals `PerformanceObserver`, UserTiming mark/measure, EmptyRoot root node) stay as plain early-returns inside `_onEnable` and never throw, so one unsupported feature can't disrupt enabling the others.

## Testing
- [x] `npm run check` (typecheck + ESLint baseline) and `npm run lint` (Biome) clean from the repo root.
- [x] Unit suite green, except one pre-existing, order-dependent `UserTimingInstrumentation` flake that also fails on clean `main` (passes in isolation).
- [x] New e2e test in `initSDK.test.ts` flows the default `LoafInstrumentation` through `initSDK`, mocks `PerformanceObserver` to emit a fake LoAF entry, and asserts `emb-loaf-report` appears — defends that the autoLoader-deferred `enable()` lands after manager wiring.
- [x] Per-subclass tests updated for the no-auto-enable lifecycle (explicit `.enable()` after construction).

Note: a capability-gated instrumentation's `isEnabled()` can now report `true` even where its observer was unavailable (the flag flips before the hook's early-return). Harmless in practice; making it honest is exactly the kind of lifecycle policy deliberately kept out of this PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "tests/integration"
       ],
       "devDependencies": {
-        "@arethetypeswrong/core": "0.18.2",
+        "@arethetypeswrong/core": "0.18.3",
         "@biomejs/biome": "2.4.15",
         "@commitlint/cli": "21.0.1",
         "@commitlint/config-conventional": "21.0.1",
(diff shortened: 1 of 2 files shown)


packages/web-sdk/src/instrumentations/clicks/ClicksInstrumentation/ClicksInstrumentation.ts
```diff
@@ -70,17 +70,15 @@ export class ClicksInstrumentation extends EmbraceInstrumentationBase {
...
-  public disable(): void {
+  public override disable(): void {
+    this.setConfig({ ...this._config, enabled: false });
     document.removeEventListener('click', this._onClickHandler);
   }
 
-  public enable(): void {
+  public override enable(): void {
     document.addEventListener('click', this._onClickHandler);
+    this.setConfig({ ...this._config, enabled: true });
```
overbalance: I considered a helper for this repeating pattern but I think we should stay aligned to the style of OTel source code instead

joaquin-diaz (approved):

PR #1328, opened 2026-05-15
chore(deps): split tests/performance out of root workspace

overbalance:

Pulling `tests/performance` into the root npm workspace dragged Lighthouse and other large transitive deps (plus an OpenTelemetry tree) into the root `node_modules`, bloating installs for everyone who never runs the perf suite.

- Remove `tests/performance` from the root `workspaces` array; it now installs independently with its own lockfile.
- Switch perf-suite SDK deps from workspace `*` to `file:` references and add a `pretest` hook that rebuilds the parent and reinstalls.
- Set `enableErrorReporting: false` explicitly in the Lighthouse run so the stub is provably unreachable.

## Testing

- [x] `npm install` at root no longer pulls Lighthouse
- [x] `npm run test:performance` still passes end-to-end

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,8 +8,7 @@
       "workspaces": [
         "packages/*",
         "demo/*",
-        "tests/integration",
-        "tests/performance"
+        "tests/integration"
       ],
       "devDependencies": {
         "@arethetypeswrong/core": "0.18.2",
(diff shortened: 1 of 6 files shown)


jpmunz (approved):
lgtm! Is this something that should have been caught by the automated checks that look at bundle size? Or was the increase not large enough to have triggered it?

overbalance:
@jpmunz these node OTel packages are all dev dependencies imported with lighthouse. They are not included in our output so they are not subject to validation. It's not a bad idea to measure the node_modules folder size over time though.

PR #1334, opened 2026-05-19
EMBR-12221 feat(rage-click): add rage click instrumentation

joseph-embrace:


packages/web-sdk/src/instrumentations/rage-clicks/RageClicksInstrumentation/RageClicksInstrumentation.ts
```diff
@@ -0,0 +1,150 @@
...
+      return;
+    }
+
+    const sessionSpan = this.sessionManager.getSessionSpan();
+    if (!sessionSpan) {
+      this._diag.debug('no active session for rage-click event');
+      return;
+    }
+
+    sessionSpan.addEvent(
```
joaquin-diaz: We should emit a log instead of a span event, is similar in structure but is not attached to the session span

packages/web-sdk/src/instrumentations/rage-clicks/RageClicksInstrumentation/RageClicksInstrumentation.ts
```diff
@@ -0,0 +1,150 @@
...
+    const withinRadius =
+      dx <= RAGE_CLICK_RADIUS_PX && dy <= RAGE_CLICK_RADIUS_PX;
+
+    if (sameTarget || withinRadius) {
+      this._cluster.count += 1;
+    }
+  }
+
+  private _flushCluster(): void {
+    const cluster = this._cluster;
```
joaquin-diaz: Maybe doing:
```
if (this._cluster === null) {
  return;
}

.. emit log

this._cancelCluster();
```

is more readable? and you need the extra const. Also by setting the cluster to null you're not clearing the timeout

packages/web-sdk/src/instrumentations/rage-clicks/RageClicksInstrumentation/RageClicksInstrumentation.ts
joseph-embrace: Thanks! Changed to a log.

packages/web-sdk/src/instrumentations/rage-clicks/RageClicksInstrumentation/RageClicksInstrumentation.ts
joseph-embrace: We don't need to clear the timeout in here because this function only fires when the timeout is complete, but I am doing a refactor anyway and it probably will be clearer to put all the nullify/clearTimeout in one place.

We need to reassign to the local scope because we want to nullify the cluster regardless of the code path. This code wouldn't work:

```ts
if (this._cluster === null) {
  return;
}

this._cluster = null;

// ⚠️ Fails because this._cluster is now null
if (this._cluster.count < RAGE_CLICK_THRESHOLD) {
  return;
}
```

and we can't nullify `this._cluster` after that `count` check because we might return early and have a dangling cluster.

jpmunz (approved):
lgtm, one thing I wanted to raise is that we have these extra config attributes in the existing ClickInstrumentation:
https://github.com/embrace-io/embrace-web-sdk/blob/70c310af2a9d6d9d5a26b8ea45ff859f0d2e228e/packages/web-sdk/src/instrumentations/clicks/ClicksInstrumentation/types.ts#L4

these were added because the method of getting a friendly name for the element there has the possibility of pulling PII from the text element. But I don't believe that would be the case with `getSelector`? We may want a follow-up that switches ClickInstrumentation to use `getSelector` as well for consistency

joseph-embrace:
> lgtm, one thing I wanted to raise is that we have these extra config attributes in the existing ClickInstrumentation:
> 
> https://github.com/embrace-io/embrace-web-sdk/blob/70c310af2a9d6d9d5a26b8ea45ff859f0d2e228e/packages/web-sdk/src/instrumentations/clicks/ClicksInstrumentation/types.ts#L4
> 
> these were added because the method of getting a friendly name for the element there has the possibility of pulling PII from the text element. But I don't believe that would be the case with `getSelector`? We may want a follow-up that switches ClickInstrumentation to use `getSelector` as well for consistency

Interesting! We won't have that problem with `getSelector` as it doesn't use `innerText` at all. I will raise this with Andy because if we want to standardise on this I think there's a case to go further than what `getSelector` does. For example reading from `data-track`, `data-trackid`, `elementtiming`, etc.

PR #1335, opened 2026-05-20
EMBR-12223 feat: instrument page load abandonment detection

joaquin-diaz:
When a user closes/navigates away before the page finishes loading, we have no visibility into that abandonment. This impacts understanding of startup performance and drop-off rates.

- Add `emb.page_load` attribute to cold-start session part spans, stamped at session part end:
  - `true` if `document.readyState === 'complete'` at the time the part ends
  - `false` if the page had not finished loading
- `readyState` is checked via the existing `visibilityDoc` reference in `EmbraceUserSessionManager`, avoiding any race condition with instrumentation enabling order
- Only cold-start parts get the attribute (`emb.cold_start === true`)
- Flush the log pipeline at session part end via `loggerProvider.forceFlush()`


jpmunz (approved):

overbalance:
What do you think about a name like "early exit" or "exit before load"? The phrase incomplete startup makes it sound to me like the SDK had an error trying to get started, not describing that it's the user's choice.

EDIT: thinking about it more, maybe we should set "load" to false by default, then flip it to true using the existing logic. Then you don't have to add the attribute as you're exiting. There's already a lot of work happening on page exit in general. One less racy item might be good.

overbalance (changes requested):
pagehide ~~fires~~ can fire before visibilitychange so the sessionSpan will be null sometimes. I'll review locally to confirm.

packages/web-sdk/src/sdk/initSDK.ts
```diff
@@ -316,6 +316,12 @@ export const initSDK = (
       });
     }
 
+    // Make sure we flush all logs triggered by any instrumentation
```
jpmunz: we can omit this now with the updated approach right?

packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
```diff
@@ -541,6 +550,10 @@ export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<Docu
 
   public enable(): void {
     window.removeEventListener('load', this._onDocumentLoaded);
+    const span = this.sessionManager.getSessionSpan();
+    if (span?.attributes[KEY_EMB_COLD_START] === true) {
+      span.setAttribute(KEY_EMB_PAGE_LOAD, false);
```
jpmunz: is this needed? Will we have a distinction in metric calculation between `cold_start=true AND page_load=false` vs. `cold_start=true` and a missing page_load attribute?

packages/web-sdk/src/sdk/initSDK.ts
joaquin-diaz: Yeah we could but I don't see any harm in keeping it. It makes sure we don't miss-match logs and sessions, even if we set session id on emit, in case we ever try changing anything and getting the current session too late. We would also need it in case we do want to send a log on exit in the future

packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
joaquin-diaz: Technically not but it feels cleaner to avoid sending this extra attribute when is not needed. It only makes sense paired with cold_start=true

packages/web-sdk/src/sdk/initSDK.ts
jpmunz: one potential downside is because we're forcing the flush we'd bypass any logic we have around log batching, or is that separate from flushing on the provider and we'd still have the exporter able to batch?

may need to discuss in more detail I'm not sure I'm following the potential for mismatching, the session id / timestamp of the log is sent when it is emitted and flushing here is just controlling when it is sent out isn't it?

packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
jpmunz: ah, I was referring to this full block not just the cold_start=true check, like do we need to ever send `page_load=false`?

joaquin-diaz: It's a useful way of knowing that this check run correctly. But no we don't need it to flag a session part as abandoned, we only need both to be true. I think I'd rather be explicit and send it always when cold start is true. I also moved this logic to the `EmbraceUserSessionManager` to avoid any race condition of the session not being there when the instrumentation runs

packages/web-sdk/src/sdk/initSDK.ts
joaquin-diaz: It will export all the logs in the pending batch. 

I'll remove this, worst case is that the next time we're adding a log on session ending it won't show up until we add this back, but we'll see that during testing

overbalance (approved):

jpmunz (approved):

PR #1337, opened 2026-05-22
EMBR-12322 chore: removing webpack4 integration testing

jpmunz:
Removing webpack4 integration testing given security scares with this platform, want to minimize potential impact on developers running these tests locally and pulling in outdated dependencies


packages/web-sdk/README.md
```diff
@@ -612,8 +612,6 @@ module.exports = {
 };
 ```
```
overbalance: I'm not sure we should even provide this guidance. As more packages use `exports`, this list will grow and grow. The management doesn't seem worth it.
```suggestion
```

jpmunz: ya that's fair, removed

overbalance (approved):

PR #1339, opened 2026-05-22
EMBR-8265 chore: resolve cache poisoning alerts

jpmunz:
Resolving the zizmor cache poisoning audit alerts from https://github.com/embrace-io/embrace-web-sdk/security/code-scanning

Audits flagged a bunch of occurrences, these fell into 3 buckets:

1. #### Use of npm cache w/ `actions/setup-node` in publish / preview steps
    - these were the actual vulnerabilities where if someone was able to poison the cache they could get malicious code into the published artifacts, some details on how this could happen in https://docs.zizmor.sh/audits/#cache-poisoning
      - worth noting that this would be difficult in practice since the package lock file should make sure that we aren't installing mismatched dependencies
    - fix here was to remove caching on these two steps, which plugs the vulnerability but does mean these jobs will take a bit longer going forward since they can't leverage the cache
2.  #### Use of npm cache w/ `actions/setup-node` in other steps (e.g. test-sdk)
    - these should be safe to ignore since there are no artifacts published from these steps
3.  #### Use of `actions/cache`
    - we're only using this action to check for the existence of a previously passed check and not reading any data from cache so safe to ignore


jpmunz:
@embrace-io/opensource-devops if you're able to confirm my understanding of how to address these audits is correct

overbalance (approved):

vitaliyf (approved):

PR #1347, opened 2026-05-26
EMBR-12416 refactor(sessions): remove pagehide/pageshow listeners and web_hard_navigation reason

overbalance:
The `pagehide` and `pageshow` listeners on `EmbraceUserSessionManager` were dead code in shipping engines. Per the HTML "unload a document" algorithm and verified empirical timings in Chrome, Gecko, and WebKit, `blur` (active-tab unload) or an earlier `visibilitychange`-to-`hidden` (backgrounded-tab unload) always ends the part before `pagehide` fires; on BFCache restore, `focus` and `visibilitychange`-to-`visible` fire before `pageshow`. The `web_hard_navigation` end reason was therefore unreachable in production.

Independent of that, `getVisibilityState` was treating non-visible states (`prerender`, `unloaded`, missing) as `Foreground`, which is wrong for the `emb.state` "actively viewing" contract.

- Remove `_onPageHide`/`_onPageShow` handlers and their listener registrations.
- Drop `pagehide`/`pageshow` from `addActivityListeners`/`removeActivityListeners`.
- Drop `web_hard_navigation` from the `SessionPartEndReason` union. All disengagement classifies as `web_background`.
- Invert `getVisibilityState` so anything other than `'visible'` maps to `Background`, fixing `emb.state` stamping for `prerender`/`unloaded`/missing-value cases and tightening the part-start engagement gate.
- Update tests and README to reflect the engagement-transition-only model.

## Testing
- `npm run check` (tsc + eslint, repo root) passes.
- `npm test` from `packages/web-sdk/`: 789 passed, 1 skipped.
- `npm run lint:fix` (repo root) reports no changes.
- No integration golden changes required (no golden pinned `web_hard_navigation`).

### DIFF
--- a/packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
+++ b/packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
@@ -782,16 +782,12 @@ export class EmbraceUserSessionManager implements UserSessionManagerInternal {
     this._handleEngagementTransition('focus');
   };
 
-  // Window focus lost (alt-tab away, clicked into DevTools or another window).
-  // hasFocus is now false; engagement transition ends the active part.
-  //
-  // pagehide and pageshow are intentionally not listened to. In shipping
-  // engines, blur fires first on active-tab unload, and the earlier
-  // visibilitychange-to-hidden already ended the part for backgrounded-tab
-  // unload. pagehide always short-circuits. On BFCache restore, focus and
-  // visibilitychange-to-visible fire before pageshow, so pageshow finds
-  // an already-active part. Initial-load part-start is driven directly by
-  // initSDK via the 'init' SessionPartStartReason.
+  // Window focus lost (alt-tab, DevTools, another window). Engagement
+  // transition ends the active part as web_background. pagehide/pageshow
+  // are not listened to: blur or visibilitychange-to-hidden end the part
+  // first (see README "Unload and BFCache handling" for the verified
+  // cross-engine ordering). The part-end doubles as the unload flush via
+  // EmbraceSessionPartBatchedSpanProcessor.onEnd's synchronous keepalive export.
   // https://developer.chrome.com/docs/web-platform/page-lifecycle-api
   private readonly _onBlur = (): void => {
     this._handleEngagementTransition('blur');
--- a/packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
+++ b/packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
@@ -88,8 +88,8 @@ import {
 
 /**
  * Parts are engagement-gated (visible AND focused), so only one part can
- * be active at a time. That mutex lets us treat storage as a
- * plain shared row read on engagement and written on changes.
+ * be active at a time. Engagement exclusivity lets us treat storage as
+ * a plain shared row read on engagement and written on changes.
  */
 export class EmbraceUserSessionManager implements UserSessionManagerInternal {
   private _state: UserSessionState | null = null;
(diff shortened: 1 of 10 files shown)


packages/web-sdk/src/utils/getVisibilityState.ts
```diff
@@ -1,7 +1,11 @@
 import type { VisibilityStateDocument } from '../common/index.ts';
 import { EMB_STATES } from '../constants/index.ts';
 
+// emb.state describes whether the user is actively viewing the page, not
+// whether visibilityState is literally 'hidden'. Non-visible states
+// (prerender, unloaded, hidden) are not "actively viewing", so they all
+// map to Background.
 export const getVisibilityState = (visibilityDoc: VisibilityStateDocument) =>
-  visibilityDoc.visibilityState === 'hidden'
```
jpmunz, line 5: were you seeing other values besides 'visible' or 'hidden' being reported here? According to https://www.w3.org/TR/page-visibility-2/#visibilitystate-enum those should be the only possibilities which would make this equivalent to the previous block

packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
```diff
@@ -787,45 +784,20 @@ export class EmbraceUserSessionManager implements UserSessionManagerInternal {
 
   // Window focus lost (alt-tab away, clicked into DevTools or another window).
   // hasFocus is now false; engagement transition ends the active part.
+  //
+  // pagehide and pageshow are intentionally not listened to. Per the HTML
+  // spec's "unload a document" algorithm (§7.5.9), pagehide fires before
```
jpmunz: can you include a link to this reference?

overall does this observed behaviour have to be the case given the specification or does it just happen to so atm given current browser versions tested? If the order could change in the future is there a downside to keeping the logic in place?

packages/web-sdk/src/managers/EmbraceUserSessionManager/README.md
```diff
@@ -291,18 +290,22 @@ Several specific edge cases are handled:
...
   runs benignly against the already-cleared storage row.
 
-## BFCache handling
+## Unload and BFCache handling
 
-There is no BFCache-specific code in the manager. Behavior is absorbed by the
-existing engagement events:
+There is no BFCache-specific code in the manager and no `pagehide`/`pageshow`
+listener. Behavior is absorbed by the engagement-transition events because, per
+the HTML "unload a document" algorithm and the empirically verified ordering in
```
jpmunz: can we link to a reference for the algorithm, also let's list off the specific browsers + versions tested

packages/web-sdk/src/utils/getVisibilityState.ts
overbalance: I wasn't able to replicate but checked the spec to make sure and saw these extra states. Looks like it was deprecated in favor of document.prerendering though. I'll remove.

packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
overbalance: Thanks, I just noticed that citation too and added it.

The original spec had a certain order but browsers had already shipped a different order and it was considered too large of a breaking change, so this description reflects reality and the revised spec. I doubt they would ever change this given the reluctance to modify it in the past.

jpmunz (approved):
change looks to me with the added context in the comments

similar question as before though, is there a downside to keeping the pageShow/pageHide handlers in place in case we encounter a browser where this behaviour changes? They would just be no-ops if the visibility handlers fire first correct?

overbalance:
@jpmunz part start/end is tightly coupled to visbilitychange so adding an additional listener would require deduping logic to make sure we construct the right number of parts. very old safari fires pagehide without visibilitychange but it's an outlier. for our current browser support we should be good with a single listener. if we see data loss in the future, we can address then

PR #1352, opened 2026-05-28
EMBR-12222 feat(first-interaction): add first interaction instrumentation

joseph-embrace:
This instrumentation is pretty simple:

- When enabled, attach listeners for pointerdown, keydown, and scroll
- When any of those fires, emit a log and detach the listeners
- On session part start, attach new listeners and do it all again
- 💃

### DIFF
--- a/packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
@@ -1,0 +1,167 @@
+import { SeverityNumber } from '@opentelemetry/api-logs';
+import { EMB_TYPES, KEY_EMB_TYPE } from '../../../constants/attributes.ts';
+import { getSelector } from '../../../utils/index.ts';
+import { EmbraceInstrumentationBase } from '../../EmbraceInstrumentationBase/index.ts';
+import type { FirstInteractionType } from './constants.ts';
+import {
+  ATTR_FIRST_INTERACTION_ELEMENT_SELECTOR,
+  ATTR_FIRST_INTERACTION_ELEMENT_TYPE,
+  ATTR_FIRST_INTERACTION_INTERACTION_TYPE,
+  ATTR_FIRST_INTERACTION_X,
+  ATTR_FIRST_INTERACTION_Y,
+  FIRST_INTERACTION_EVENT_NAME,
+} from './constants.ts';
+import type { FirstInteractionInstrumentationArgs } from './types.ts';
+
+interface InteractionData {
+  interactionType: FirstInteractionType;
+  target: Element | null;
+  timestamp: number;
+  x?: number;
+  y?: number;
+}
+
+/*
+  Captures the first user input event in each session part and emits one log
+  per part. Listeners are detached after the first event is logged.
+*/
+export class FirstInteractionInstrumentation extends EmbraceInstrumentationBase {
+  private readonly _onClick: (event: PointerEvent) => void;
+  private readonly _onKeyDown: (event: KeyboardEvent) => void;
+  private readonly _onScroll: (event: Event) => void;
+  private _listenersAttached = false;
+  private _removeSessionPartStartedFn: (() => void) | null = null;
+
+  public constructor({ diag, perf }: FirstInteractionInstrumentationArgs = {}) {
+    super({
+      instrumentationName: 'FirstInteractionInstrumentation',
+      instrumentationVersion: '1.0.0',
+      diag,
+      perf,
+      config: {},
+    });
+
+    this._onClick = (event: PointerEvent): void => {
+      this._handle({
+        interactionType: event.pointerType === 'touch' ? 'tap' : 'click',
+        target: event.target instanceof Element ? event.target : null,
+        timestamp: event.timeStamp,
+        x: event.clientX,
+        y: event.clientY,
+      });
+    };
+
+    this._onKeyDown = (event: KeyboardEvent): void => {
+      this._handle({
+        interactionType: 'keypress',
+        target: event.target instanceof Element ? event.target : null,
+        timestamp: event.timeStamp,
+      });
+    };
+
+    this._onScroll = (event: Event): void => {
+      this._handle({
+        interactionType: 'scroll',
+        target: null,
+        timestamp: event.timeStamp,
+      });
+    };
+
+    if (this._config.enabled) {
+      this.enable();
+    }
+  }
+
+  public enable(): void {
+    this._attachListeners();
+
+    if (!this._removeSessionPartStartedFn) {
+      this._removeSessionPartStartedFn =
+        this.userSessionManager.addSessionPartStartedListener(() => {
+          this._attachListeners();
+        });
+    }
+  }
+
+  public disable(): void {
+    this._detachListeners();
+
+    if (this._removeSessionPartStartedFn) {
+      this._removeSessionPartStartedFn();
+      this._removeSessionPartStartedFn = null;
+    }
+  }
+
+  private _handle(data: InteractionData): void {
+    if (!this._listenersAttached) {
+      // It's possible two handlers fire in quick succession before the listeners
+      // are detached. We only want to process the first one.
+      return;
+    }
+
+    this._detachListeners();
+
+    try {
+      this._emit(data);
+    } catch (e) {
+      this._diag.error('failed to process first-interaction candidate', e);
+    }
+  }
+
+  private _attachListeners(): void {
+    if (this._listenersAttached) {
+      return;
+    }
+
+    const opts: AddEventListenerOptions = { capture: true, passive: true };
+
+    document.addEventListener('click', this._onClick, opts);
+    document.addEventListener('keydown', this._onKeyDown, opts);
+    document.addEventListener('scroll', this._onScroll, opts);
+
+    this._listenersAttached = true;
+  }
+
+  private _detachListeners(): void {
+    if (!this._listenersAttached) {
+      return;
+    }
+
+    const opts: EventListenerOptions = { capture: true };
+
+    document.removeEventListener('click', this._onClick, opts);
+    document.removeEventListener('keydown', this._onKeyDown, opts);
+    document.removeEventListener('scroll', this._onScroll, opts);
+
+    this._listenersAttached = false;
+  }
+
+  private _emit(data: InteractionData): void {
+    let elementType = 'document';
+    let elementSelector = '';
+
+    if (data.interactionType !== 'scroll' && data.target) {
+      elementType = data.target.tagName.toLowerCase();
+      elementSelector = getSelector(data.target);
+    }
+
+    const attributes: Record<string, string | number> = {
+      [KEY_EMB_TYPE]: EMB_TYPES.OTelLog,
+      [ATTR_FIRST_INTERACTION_INTERACTION_TYPE]: data.interactionType,
+      [ATTR_FIRST_INTERACTION_ELEMENT_TYPE]: elementType,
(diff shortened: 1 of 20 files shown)


packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/constants.ts
```diff
@@ -0,0 +1,12 @@
+export const FIRST_INTERACTION_EVENT_NAME = 'first_interaction';
```
jpmunz: not sure if this is a specific otel convention but we do use a '-' for event names in other instrumentations, e.g. https://github.com/embrace-io/embrace-web-sdk/blob/b73b3973c9d05e05497c2c2088ba33234aea5405/packages/web-sdk/src/instrumentations/rage-click/RageClickInstrumentation/constants.ts#L12

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    this._listenersAttached = true;
+  }
+
+  private _detachListeners(): void {
+    if (!this._listenersAttached) {
+      return;
+    }
+
+    const opts: EventListenerOptions = { capture: true };
```
jpmunz, line 130: does this need to match the `opts` set when attaching?

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  private _attachListeners(): void {
+    if (this._listenersAttached) {
+      return;
+    }
+
+    const opts: AddEventListenerOptions = { capture: true, passive: true };
+
+    document.addEventListener('pointerdown', this._onPointerDown, opts);
```
jpmunz: in https://github.com/embrace-io/embrace-web-sdk/blob/b73b3973c9d05e05497c2c2088ba33234aea5405/packages/web-sdk/src/instrumentations/rage-click/RageClickInstrumentation/RageClickInstrumentation.ts#L70 we're listening on 'click', is there a particular reason to use 'pointerdown' here instead?

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
```diff
@@ -0,0 +1,164 @@
...
+    super({
+      instrumentationName: 'FirstInteractionInstrumentation',
+      instrumentationVersion: '1.0.0',
+      diag,
+      perf,
+      config: {},
+    });
+
+    this._onPointerDown = (event: PointerEvent): void => {
+      this._handle(() => ({
```
jpmunz: curious about the extra call here, could the event handlers call `this._handle` directly with these objects rather than supplying a function that returns them?

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
```diff
@@ -0,0 +1,164 @@
...
+  private _handle(getInteractionData: () => InteractionData): void {
+    if (!this._listenersAttached) {
+      // It's possible two handlers fire in quick succession before the listeners
+      // are detached. We only want to process the first one.
+      return;
+    }
+
+    try {
+      this._emit(getInteractionData());
+      this._detachListeners();
```
jpmunz: I'm wondering if `this._detachListeners()` should be in a `finally` instead, in case we have some situation where the emit keeps failing

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
```diff
@@ -0,0 +1,164 @@
...
+      [ATTR_FIRST_INTERACTION_ELEMENT_TYPE]: isScroll
+        ? 'document'
+        : (data.target?.tagName.toLowerCase() ?? ''),
+      [ATTR_FIRST_INTERACTION_ELEMENT_SELECTOR]: isScroll
+        ? ''
+        : getSelector(data.target),
+    };
+
+    if (data.x !== undefined && data.y !== undefined) {
+      attributes[ATTR_FIRST_INTERACTION_X] = String(data.x);
```
jpmunz: is the String cast needed here?

jpmunz (commented):
some minor comments, lgtm overall

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
joseph-embrace: Yeah so I went back and forth on this. The intention is that `_handle()` can call this in a try-catch to avoid any errors bubbling up. In hindsight it's probably too defensive - I'll back it out and just pass plain objects in to `_handle()`.

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
joseph-embrace: Good catch! Although this try-catch will not be necessary if I remove the callback pattern (https://github.com/embrace-io/embrace-web-sdk/pull/1352#discussion_r3320645009)

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
joseph-embrace: @joaquin-diaz in the spec you wrote that this should be a "string-encoded float". It seems like most other instrumentations use `Math.round()` to truncate numbers to int-like precision. Should we do that here as well rather than casting to a string?

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
joseph-embrace: @jpmunz Great question! It's more of a product question that I will raise with Andy. The click event fires when a pointer is both pressed (`pointerdown`) _and_ released (`pointerup`) on the same element. There are some user behaviours that `click` doesn't capture, e.g. selecting text with a mouse or moving your finger around the screen with a touch device.

Historically web analytics have treated pointerdown as interactions, which is why I've chosen that. But we may want to do something different, so I'll double check. Actually @joaquin-diaz the spec does say to use `pointerdown` - do you remember if that was an intentional product decision?

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
joseph-embrace: Only `capture` is [a valid option](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#options) for removeEventListener

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
joseph-embrace: Actually, [`clientX` is a `long`](https://w3c.github.io/pointerevents/#ref-for-dom-mouseevent-clientx-1) (int) so I will just pass the raw `number` through.

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/constants.ts
joseph-embrace: @joaquin-diaz the spec says `first_interaction` - can you confirm it's okay to hyphenate instead? It does seem like all the other event names are hyphenated except for `browser.web_vital`:

```
λ ack 'EVENT_NAME =' packages/web-sdk/src
packages/web-sdk/src/instrumentations/loaf/LoafInstrumentation/constants.ts
13:export const LOAF_EVENT_NAME = 'emb-loaf-report';
14:export const LOAF_SCRIPTS_EVENT_NAME = 'emb-loaf-scripts';

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/constants.ts
3:export const WEB_VITAL_EVENT_NAME = 'browser.web_vital';

packages/web-sdk/src/instrumentations/server-timing/ServerTimingInstrumentation/constants.ts
1:export const SERVER_TIMING_EVENT_NAME = 'emb-server-timing';

packages/web-sdk/src/instrumentations/rage-click/RageClickInstrumentation/constants.ts
12:export const RAGE_CLICK_EVENT_NAME = 'rage-click';

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/constants.ts
1:export const FIRST_INTERACTION_EVENT_NAME = 'first_interaction';

packages/web-sdk/src/instrumentations/user-timing/UserTimingInstrumentation/constants.ts
1:export const USER_TIMING_EVENT_NAME = 'emb-user-timing';
```

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
joseph-embrace: [Confirmed with Andy](https://embrace-io.slack.com/archives/C08494GKTRV/p1780058854549599?thread_ts=1780003976.874979&cid=C08494GKTRV) that we should use `click`.

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/constants.ts
joseph-embrace: I changed this to `first-interaction` for consistency. Easy enough to change back if that's preferred, but I feel like the spec was a typo.

jpmunz (approved):

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
```diff
@@ -0,0 +1,167 @@
...
+      this._diag.error('failed to process first-interaction candidate', e);
+    }
+  }
+
+  private _attachListeners(): void {
+    if (this._listenersAttached) {
+      return;
+    }
+
+    const opts: AddEventListenerOptions = { capture: true, passive: true };
```
overbalance, line 116: Is `once` worth adding here?

joaquin-diaz (approved):

packages/web-sdk/src/instrumentations/first-interaction/FirstInteractionInstrumentation/FirstInteractionInstrumentation.ts
joseph-embrace: I thought about this a bit more and decided against `once`. It doesn't add anything or solve any existing problems, since we need to keep track of and remove the other event listeners anyway.

PR #1353, opened 2026-05-28
feat(sessions): source user-session durations from remote config

overbalance:

The user-session max-duration and inactivity-timeout were local `initSDK` options. They now come from remote config and refresh per user session, so the backend can tune them without an SDK change.

- Remove the local `initSDK` options and the `UserSessionConfig` plumbing.
- Parse `max_duration_seconds` / `inactivity_timeout_seconds` from the remote `user_session` payload into `DynamicSDKConfig`.
- Resolve the durations (clamp + inactivity-<=-max) from the `DynamicConfigManager` at each user-session creation and freeze them into that session, so a mid-session config change only affects the next session.
- Refresh remote config on each new user session, skipping cold start (already refreshed by `initSDK`). The async fetch lands for the following session.
- Name the resolved fields entity-first (`userSessionMaxDurationSeconds`, `userSessionInactivityTimeoutSeconds`) to match the `USER_SESSION_*` constants and `emb.user_session_*` attribute keys.

## Testing

- `npm run check` (tsc + eslint) passes.
- Unit tests pass, including new coverage for remote `user_session` parsing, the per-session refresh and cold-start skip (including the `setTracerProvider` eager-create path), and duration freezing across config changes.

---

> Stacked on #1355 (fake-fetch test-helper refactor). Review/merge that first.

### DIFF
--- a/packages/web-sdk/src/managers/EmbraceConfigManager/types.ts
+++ b/packages/web-sdk/src/managers/EmbraceConfigManager/types.ts
@@ -9,12 +9,15 @@ export type RemoteConfigURLParams = {
   deviceId: string;
 };
 
-// All numbers from remote config are ranges from 0.0 to 100.0 that represent the % of devices which should be enabled
-// for that particular feature / option
+// Percentage fields hold a 0.0-100.0 value: the % of devices a feature / option is enabled for.
 export type RemoteConfig = {
   threshold: number; // Main traffic control %, devices that fall outside of this should not emit any telemetry at all
   network_span_forwarding?: {
     pct_enabled: number;
+  };
+  user_session?: {
+    max_duration_seconds?: number;
+    inactivity_timeout_seconds?: number;
   };
 };
 
--- a/packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
+++ b/packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
@@ -168,38 +168,42 @@ export class EmbraceUserSessionManager implements UserSessionManagerInternal {
    * state.
    */
   private _resolveUserSessionDurations(): {
-    maxUserSessionDurationSeconds: number;
-    inactivityTimeoutSeconds: number;
+    userSessionMaxDurationSeconds: number;
+    userSessionInactivityTimeoutSeconds: number;
   } {
     const config = this._dynamicConfigManager.getConfig();
 
-    const maxUserSessionDurationSeconds = clampNumber({
+    const userSessionMaxDurationSeconds = clampNumber({
       diag: this._diag,
-      value: config.maxUserSessionDurationSeconds,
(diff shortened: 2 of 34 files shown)


packages/web-sdk/src/managers/EmbraceConfigManager/types.ts
```diff
@@ -9,13 +9,16 @@ export type RemoteConfigURLParams = {
...
 
-// All numbers from remote config are ranges from 0.0 to 100.0 that represent the % of devices which should be enabled
-// for that particular feature / option
+// Percentage fields hold a 0.0-100.0 value: the % of devices a feature / option is enabled for.
 export type RemoteConfig = {
   threshold: number; // Main traffic control %, devices that fall outside of this should not emit any telemetry at all
   network_span_forwarding?: {
     pct_enabled: number;
   };
+  user_session?: {
```
jpmunz, line 18: are these placeholders or has the remote config service already been updated to send these particular keys?

packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
```diff
@@ -610,11 +629,22 @@ export class EmbraceUserSessionManager implements UserSessionManagerInternal {
         EMBRACE_USER_SESSION_NUMBER_KEY,
         this._diag,
       );
+      // Refresh remote config for the next user session. The fetch is async,
+      // so it lands in the cache after this session's durations are frozen
+      // below; this session uses the currently cached config. Skipped on cold
+      // start, where initSDK already refreshes at startup.
+      if (!this._coldStart) {
+        void this._dynamicConfigManager.refreshRemoteConfig();
```
jpmunz, line 637: I'm wondering if we need a guard here against this being called too frequently, though I guess the only way that would really happen would be through multiple manual API calls to end a user session

alternatively we could enforce a cooldown in EmbraceDynamicConfigManager itself, maybe better since that would protect against different callers separately asking for refreshes, that would also mean we wouldn't need the ` if (!this._coldStart) {` check here

packages/web-sdk/src/managers/EmbraceConfigManager/types.ts
overbalance: The service is already sending this format.

packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
overbalance: This guard exists so remote config doesn't double request on cold start: init() does a remote fetch of its own. Mobile has a one hour cooldown but that complicates QA enough that I consider it out of scope for the next release.

jpmunz: sounds good, we can consider as a follow-up

jpmunz (approved):

PR #1357, opened 2026-05-29
EMBR-12298 feat(sessions): add web foreground inactivity timeout

overbalance:

The live foreground part-inactivity timer was driven by `userSessionInactivityTimeoutSeconds`, the same value used for the lazy post-backgrounding deadline, so the two could not be tuned independently. This adds a distinct, remote-config-driven foreground inactivity timeout so the live in-tab timer and the lazy cross-visit deadline can be configured separately, and aligns the part end reason it produces with that "foreground inactivity" naming.

- Parse `web_foreground_inactivity_timeout_seconds` from remote config into `userSessionForegroundInactivityTimeoutSeconds` on `DynamicSDKConfig`.
- Clamp to [30s, 24h] (default 1800s). When it exceeds the max session duration it falls back to its own default, independently of the inactivity timeout, and emits a `diag.warn`.
- Freeze the resolved value into the user-session state at creation and emit `emb.user_session_foreground_inactivity_timeout_seconds`.
- Drive the live part-inactivity timer from the foreground value; the lazy post-part-end deadline (`inactivityDeadlineTs`) still uses `userSessionInactivityTimeoutSeconds`.
- Rename the part-level `SessionPartEndReason` value `web_inactivity` to `web_foreground_inactivity` so the emitted end reason matches the timeout that produces it. This changes the wire value of `emb.session_part_end_reason`: the backend should expect `web_foreground_inactivity`.
- Extract a `FINAL_SESSION_PART_END_REASONS` set, replacing the inline reason comparison that decides whether a part is stamped `emb.is_final_session_part`.
- Update the `EmbraceUserSessionManager` README (timers table, part-end finalization, storage, and attribute sections) to match.

Timeout defaults match the existing inactivity timeout, so timer behavior is unchanged until the backend sends the new field.

## Testing

- Unit tests cover remote-config parsing (present/absent), clamp and independent fallback, frozen-into-state, attribute stamping, and the live-vs-lazy timer split (live timer fires on the foreground value, lazy deadline on the inactivity value).
- `npm run lint` and `npm run check` clean; web-sdk unit suite green.
- Integration golden files regenerated to capture the renamed `web_foreground_inactivity` end reason.


packages/web-sdk/src/managers/EmbraceUserSessionManager/README.md
```diff
@@ -102,7 +102,7 @@ engagement condition holds, the call is a debug-level no-op.
 | Value | Trigger |
 | --- | --- |
 | `web_background` | `visibilitychange` to hidden or `blur`. Also covers hard-nav unload and BFCache freeze, since blur or an earlier `visibilitychange` to hidden ends the part before `pagehide` fires. |
-| `web_inactivity` | The 30 minute part-inactivity timer fires without any user input event resetting it. |
+| `web_inactivity` | The foreground part-inactivity timer (`userSessionForegroundInactivityTimeoutSeconds`, default 30 minutes) fires without any user input event resetting it. |
```
jpmunz: given the naming of `userSessionForegroundInactivityTimeoutSeconds` would it make sense to also rename the reason now to "web_foreground_inactivity" ?

packages/web-sdk/src/managers/EmbraceUserSessionManager/README.md
```diff
@@ -118,15 +118,16 @@ On end, the manager:
    prevent the span from ending).
 3. Inside `finally`: ends the span and clears `_sessionPartSpan`,
    `_activeSessionPartId`, and `_activeSessionPartCounts`.
-4. If `reason !== 'user_session_ended'`, calls
+4. If the end reason is not final (anything other than `user_session_ended` or
```
jpmunz, line 121: don't need to make it configurable but a think a small readable improvement in EmbraceUserSessionManager.ts could be to have some top-level FINAL_SESSION_PART_END_REASONS constant to call these ones out, could then refer to that generically here rather than listing the reasons explicitly

jpmunz (commented):
just want to confirm my understanding of the changes here:
1. no behavioural changes assuming `user_session_foreground_inactivity==user_session_inactivity`, a foreground session part ending due to user inactivity still ends the entire user session
1. this just allows that those values **could** be configured differently, if they were it would mean we could either:
    1. (`user_session_foreground_inactivity < user_session_inactivity`) choose to end a user session earlier than it normally would have if we detected foreground inactivity as oppose to no active session parts
    2. (`user_session_foreground_inactivity > user_session_inactivity`) choose to end a user session later than it normally would have if the deadline for user session inactivity passed but there was still a foreground session part with no user activity

I think the use case of 2ii) makes sense to me, we're giving a bit of an extra grace period and extending the potential length of the user session in the case that there is a focused foreground tab but just no active user interaction (e.g. potentially the user is reading an article)

I'm not sure I understand the use case for 2i), if the user had a foreground tab open and they stopped interacting then their user session would actually end quicker than if they had closed that tab correct?

packages/web-sdk/src/managers/EmbraceUserSessionManager/README.md
overbalance: I like having a Set for these, I'll add that. The README should remain consistent throughout though.

overbalance:
> just want to confirm my understanding of the changes here:
> 
> 1. no behavioural changes assuming `user_session_foreground_inactivity==user_session_inactivity`, a foreground session part ending due to user inactivity still ends the entire user session
> 2. this just allows that those values **could** be configured differently, if they were it would mean we could either:
>    
>    1. (`user_session_foreground_inactivity < user_session_inactivity`) choose to end a user session earlier than it normally would have if we detected foreground inactivity as oppose to no active session parts
>    2. (`user_session_foreground_inactivity > user_session_inactivity`) choose to end a user session later than it normally would have if the deadline for user session inactivity passed but there was still a foreground session part with no user activity

You got it, correct in all three cases.

jpmunz:
> > just want to confirm my understanding of the changes here:
> > 
> > 1. no behavioural changes assuming `user_session_foreground_inactivity==user_session_inactivity`, a foreground session part ending due to user inactivity still ends the entire user session
> > 2. this just allows that those values **could** be configured differently, if they were it would mean we could either:
> >    
> >    1. (`user_session_foreground_inactivity < user_session_inactivity`) choose to end a user session earlier than it normally would have if we detected foreground inactivity as oppose to no active session parts
> >    2. (`user_session_foreground_inactivity > user_session_inactivity`) choose to end a user session later than it normally would have if the deadline for user session inactivity passed but there was still a foreground session part with no user activity
> 
> You got it, correct in all three cases.

is `user_session_foreground_inactivity < user_session_inactivity` a valid case though? Should we be restricting that timeout from being smaller?

overbalance:
@jpmunz it may be niche, but we should give people the option to set a smaller value. Consider a customer support chat in tab A. The agent suggests they open a new tab (B), reset their password, and then return to the conversation in tab A. Backgrounding the tab is part of the process and they want to allow time for that. If the user stays on tab A for some period of time without doing anything, the agent could consider that an abandonment and move on to the next person that needs help. If that's how they want to model it, I think we should give them the option.

jpmunz (approved):

PR #1358, opened 2026-05-29
docs: correct and verify CLAUDE.md against the codebase

overbalance:

CLAUDE.md had stale/inaccurate guidance (command labels, example test path, instrumentation-registration path, bundler list, transport and session-id claims) and was missing conventions the codebase already enforces.

- Corrected commands, paths, and platform/transport/session-id claims, each verified against the code.
- Added conventions (terminology, naming, comments, testing, synchronous unload), a monorepo orientation line, and a Gotchas section. Trimmed low-value content.

## How has this been tested?

Every claim verified against the code. `npm run lint` and `npm run check` pass.

## Checklist

- [x] Documentation added
- [ ] Tests added


.claude/CLAUDE.md
```diff
@@ -138,6 +164,11 @@ npm run test:integration:update-golden    # Update golden files
 3. Export from `packages/web-sdk/src/processors/index.ts`
 4. Wire into processor chain in `initSDK.ts`
 
+## Gotchas
+
+- **Session-id keys are stamped on every log** (by `UserSessionLogRecordProcessor`). An empty-string value (no active user session) is the contract, not a bug
+- **Logs render on the session timeline.** When you need something visible on the timeline, emit a log rather than a span event
```
jpmunz: span events on the session part span are also rendered on the timeline

overbalance: Do we want to encourage new span events?

jpmunz: I guess a separate question but the statement seems to imply that span events don't render on the timeline which isn't correct

overbalance: Understood. In rewriting it, should we omit "span event" language here and in future docs?

jpmunz: not sure enough that we wouldn't have a case to refer to them in the future, I think this specific line could probably be omitted altogether as a gotcha

jpmunz (approved):

PR #1364, opened 2026-06-05
EMBR-12299 docs(sessions): document 2.21.0 user-session API migration and expose getSessionPartSpan publicly

overbalance:

The 2.21.0 user-session API renames (user session vs session part) had no migration documentation, and `getSessionPartSpan()` was only available on the internal manager interface even though the deprecated `getSessionSpan()` forwarder points users to it.

- Add a "2.x to 2.21.0" section to `UPGRADING.md` covering the renamed methods, the semantics of `endUserSession()`, and the deprecated methods whose behavior changed (no-ops / always-null returns)
- Move `getSessionPartSpan()` from `UserSessionManagerInternal` to the public `UserSessionManager` interface with TSDoc warning that the SDK owns the span lifecycle and direct access is for backwards compatibility only

## How has this been tested?

`npm run check` and `npm run lint` pass; docs-and-interface-only change with no runtime behavior modified.

## Checklist

- [x] Documentation added
- [ ] Tests added

### DIFF
--- a/packages/web-sdk/UPGRADING.md
+++ b/packages/web-sdk/UPGRADING.md
@@ -28,9 +28,8 @@ session.endSessionSpan();
 session.endUserSession();
 ```
 
-A new user session begins with
-the next session part: immediately if the tab is still foregrounded, otherwise the next time it is foregrounded. The
-call is subject to a 5-second cooldown.
+A new user session begins with the next session part: immediately if the tab is still foregrounded, otherwise
+the next time it is foregrounded. The call is subject to a 5-second cooldown.
 
 `getSessionPartSpan()` is provided for backwards compatibility only: direct access to the span will be removed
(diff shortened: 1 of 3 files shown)


packages/web-sdk/UPGRADING.md
```diff
@@ -1,5 +1,50 @@
...
+import { session } from '@embrace-io/web-sdk';
+
+// Before
+session.endSessionSpan();
+
+// After
+session.endUserSession();
+```
+
+Note that `endUserSession()` ends the entire user session, not just the current span. A new user session begins with
```
jpmunz: Not sure if we need to include the detail about the session part span? What about:

```suggestion
A new user session begins with
```

jpmunz (approved):

PR #1365, opened 2026-06-05
build: require npm 11.16.0, default node to 26.3.0, disable npm cache in publishing workflows

overbalance:

Aligns the development toolchain with open-telemetry/opentelemetry-browser#306, which requires npm 11.16.0.

- Require npm `^11.16.0` via `packageManager` and `devEngines`
- Default node to 26.3.0 (`.nvmrc`) and allow `^26.3.0` alongside `^24` in `devEngines`
- Allow the lefthook install script via `allowScripts`
- Disable npm caching in the package-publishing workflows (`package-manager-cache: false`)

## How has this been tested?

`npm ci`, lint, and typecheck pass locally on the required toolchain; CI on this PR exercises the workflow changes.

## Checklist

- [ ] Documentation added
- [ ] Tests added


.github/workflows/pkg-pr-new.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version-file: .nvmrc
-          cache: npm
+          package-manager-cache: false
```
jpmunz, line 24: hmm interesting that the zizmor checks didn't flag these

package.json
```diff
@@ -78,16 +78,19 @@
...
     "packageManager": {
       "name": "npm",
-      "version": "^11.9.0",
+      "version": "^11.16.0",
       "onFail": "error"
     },
     "runtime": {
       "name": "node",
-      "version": "^24"
+      "version": "^24 || ^26.3.0"
```
jpmunz, line 90: should this just be `^26.3.0` ? Is it possible to satisfy the `npm@11.16.0` constraint on node 24?

package.json
```diff
@@ -78,16 +78,19 @@
...
     }
+  },
+  "allowScripts": {
```
jpmunz, line 93: what does this attribute end up doing? I don't see it defined in https://docs.npmjs.com/cli/v11/configuring-npm/package-json

package.json
```diff
@@ -78,16 +78,19 @@
...
+    "lefthook@2.1.8": true
```
jpmunz, line 94: we had discussed briefly maybe just not having lefthook at all, what is it used for currently is it just for installing git hooks?

.github/workflows/pkg-pr-new.yml
overbalance: Yeah I thought so too.

package.json
overbalance: Technically yes, because you're free to update npm separately. I included 24 because it's LTS but it's not strictly necessary.

package.json
overbalance: It signals to npm that package scripts are allowed to run when installed. The default is essentially `npm i --ignore-scripts` with this list as exceptions.

package.json
overbalance: Yes. It has some features githooks doesn't have like parallelism and the ability to prevent committing.

package.json
overbalance: By the way, this is generated when you run https://docs.npmjs.com/cli/v11/commands/npm-approve-scripts

jpmunz (approved):

PR #1369, opened 2026-06-08
build(npm): set ignore-scripts to prevent install lifecycle scripts

overbalance:

Pinning npm to 11.16.0 alone does not prevent dependency install lifecycle scripts from running: it only emits a warning. To actually block arbitrary `preinstall`/`install`/`postinstall` scripts during `npm install` (supply-chain hardening), `ignore-scripts` must be set explicitly.

While enabling `ignore-scripts` in the performance-tests workspace, the `get-port` dependency had to go: with scripts ignored it is no longer reliable, and its only use was picking an ephemeral debug port, which can be a fixed value here.

- Add `ignore-scripts=true` to the root `.npmrc` and to `tests/performance/.npmrc`.
- Drop the `get-port` dependency in the performance tests; hardcode the remote debugging ports (`60061` for CDP, `60062` for Lighthouse).
- Fold the perf-tests `pretest` step into `test` (so the build/install/playwright-install run as one script).
- Remove the deprecated `enableErrorReporting` flag from the Lighthouse config.

## How has this been tested?

- `npm install` no longer runs dependency lifecycle scripts.
- Pre-commit hooks (lint, typecheck, build) pass on the change.

## Checklist

- [ ] Documentation added
- [ ] Tests added

### DIFF
--- a/tests/performance/tests/lighthouse-startup-performance-tests.spec.ts
+++ b/tests/performance/tests/lighthouse-startup-performance-tests.spec.ts
@@ -2,7 +2,6 @@ import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
 import { test } from '@playwright/test';
-import getPort from 'get-port';
 import lighthouse from 'lighthouse';
 import { chromium } from 'playwright';
 import { resultsToMarkdownTable } from '../../utils/jsonToMarkdownTable.ts';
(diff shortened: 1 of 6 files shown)


joaquin-diaz (approved):

tests/performance/tests/lighthouse-startup-performance-tests.spec.ts
```diff
@@ -111,7 +110,6 @@ test.describe('Lighthouse Performance Tests', () => {
         output: ['json', 'html'],
         onlyCategories: ['performance'],
         pauseAfterLoadMs: 5000,
-        enableErrorReporting: false,
```
joaquin-diaz, line 114: Was this a default value?

overbalance: yes, `false` is the default in CI. We don't typecheck this folder so I just noticed it and removed.

PR #1370, opened 2026-06-11
EMBR-12585 chore: add React Router SPA platform and e2e navigation tests

joaquin-diaz:
## What problem does this solve

Lays the groundwork for instrumenting SPA soft-navigation. Before adding the instrumentation itself, we need a test platform that actually does soft navigation so we can write and validate the e2e tests against it.

## Short description

- Adds `tests/integration/platforms/vite-react-router/` — a minimal React 19 + React Router v7 SPA (Home, Products, Product Detail pages) built with Vite 7. Uses `createBrowserRouter` so navigations go through `history.pushState`, not hash changes.
- Adds `tests/integration/tests/e2e/vite-react-router-tests.spec.ts` — three skipped e2e tests (forward clicks, browser back, browser forward). Each asserts that a session part ends after every soft navigation, with cumulative counts (`validateThatSessionPartEnded(n)`). Tests are skipped until soft-nav instrumentation is implemented.
- Adds `tests/integration/playwright.config.spa.ts` — a focused config that only spins up the react-router platform (via `vite preview` on port 3016) and the API server, for running SPA tests in isolation.
- Moves `getCurrentUserSessionId` and `validateThatSessionPartEnded` (previously `validateThatUserSessionEnded`) into the `TestWithMockApi` base fixture so all test suites inherit them without duplication.
- Changes `validateThatSessionPartEnded` to take `(expectedCount: number, userSessionId?: string)` and polls until the server has received at least `expectedCount` session-part spans for the current user session. The server now tracks counts instead of booleans (`ReceivedSpans` changed from `Record<string, boolean>` to `Record<string, number>`).

## Checklist

- [x] Tests added / updated
- [x] No new lint errors
- [x] Types check


tests/integration/platforms/vite-react-router/package-lock.json
overbalance, line 1: The other integrations don't have lockfiles so each CI run imports the latest packages. Copy `.npmrc` from another folder if needed

tests/integration/utils/test-with-mock-api.ts
```diff
@@ -237,6 +240,46 @@ const testWithMockApi = base.extend<TestWithMockApi>({
...
+      );
+
+      if (!userSessionId) {
+        throw new Error('Session ID is not available on the page');
+      }
+
+      return userSessionId;
+    });
+  },
+  validateThatUserSessionEnded: async ({ getCurrentUserSessionId }, use) => {
```
jpmunz: this looks to be validating that the session part ended not the user session correct?

overbalance (approved):
After addressing the lockfile

tests/integration/tests/e2e/vite-react-router-tests.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+          value: 'hidden',
+          writable: true,
+        });
+        document.dispatchEvent(new Event('visibilitychange'));
+      });
+    });
+  },
+});
+
+test.describe('Vite React Router SPA Navigation', () => {
```
jpmunz, line 33: maybe worth a TODO here, these aren't the final state of these cases correct? We'll be validating session parts ending after navigations as well later? Or alternatively maybe we should write these as we expect them for the final state but marked them as skipped for now

tests/integration/utils/test-with-mock-api.ts
joaquin-diaz: Yeah right, the name was wrong. I'll fix it

tests/integration/tests/e2e/vite-react-router-tests.spec.ts
joaquin-diaz: I didn't want to assume too much of the implementation just yet, but I think it's safe to assume that the session part would end after navigation. I'll add a `validateThatSessionPartEnded();` after each navigation and skip the tests

tests/integration/utils/test-with-mock-api.ts
```diff
@@ -237,6 +240,46 @@ const testWithMockApi = base.extend<TestWithMockApi>({
...
+      if (!userSessionId) {
+        throw new Error('Session ID is not available on the page');
+      }
+
+      return userSessionId;
+    });
+  },
+  validateThatSessionPartEnded: async ({ getCurrentUserSessionId }, use) => {
+    await use(async (userSessionId?: string) => {
+      const currentUserSessionId =
```
jpmunz, line 259: since we're validating that the session part ended is it correct that this is the user session id? Won't we generally have the same user session id for the full integration test run?

joaquin-diaz: We don't expose the session part id and I didn't want to add it so I changed the tests to just count the amount of expected session parts

PR #1371, opened 2026-06-11
build(deps): replace rimraf with native node remove script

overbalance:
## What problem is this PR solving?

`rimraf` is only used for deleting directories in `clean` scripts, but it requires Node 20+ and pulls in several transitive dependencies (`glob`, `minipass`, `path-scurry`, `package-json-from-dist`) for that one job. Since the repo already pins Node >= 26, native `fs.globSync` and `fs.rmSync` cover everything rimraf did here, with no dependency at all.

- Add `scripts/emb-rm.js`, a small helper that expands its arguments with `fs.globSync` (literal paths and glob patterns alike) and removes each match with `fs.rmSync(path, {recursive: true, force: true})`.
- The helper refuses to remove anything at or outside the working directory (exit 1, nothing deleted), so a mistyped path in a clean script cannot remove the package itself or its siblings.
- Register `scripts/` as a workspace exposing the helper as an `emb-rm` bin, so package-level `clean` / `build:clean` scripts are just `emb-rm <targets>` with no relative paths. Glob arguments stay quoted so the pattern reaches the helper instead of being half-expanded by `sh`.
- The root `clean` invokes `node scripts/emb-rm.js` directly instead of the bin, so it also works when `node_modules` is already gone (e.g. running clean twice in a row).
- Drop `rimraf` from devDependencies and regenerate the lockfile.

## How was this tested?

- Full cycle matrix in a clean room: root clean with nothing installed, `npm ci` (verifies the bin links from the lockfile), `npm run build`, per-workspace cleans against real build output, root clean after build, then `npm i` and build again.
- All 14 converted clean scripts run in place via `npm run`, including the non-workspace integration platform directories (bin resolves via ancestor `node_modules/.bin` on PATH).
- Reproduced the Next.js 16 case where `.next/node_modules/` contains symlinks to directories: `build:clean` removes `.next` cleanly and the symlink targets are left untouched (`fs.rmSync` removes the link itself, matching rimraf behavior).
- Verified the containment guard refuses parent traversal, absolute paths outside the package, and `/`, and that a mixed valid/invalid invocation deletes nothing.
- `npm run lint` and `npm run check` pass (via pre-commit hooks).

Also adds a zero-argument usage guard to emb-rm (prints usage and exits 1 instead of silently succeeding) and clarifies the escape-guard refusal message to say "at or outside" the working directory.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "type": "module",
   "scripts": {
     "prepare": "npx -y lefthook install -f",
-    "clean": "node scripts/emb-rm.js \"**/.turbo\" \"**/.next\" \"**/dist\" \"**/node_modules\"",
+    "clean": "node scripts/emb-rm.js '**/.turbo' '**/.next' '**/dist' '**/node_modules'",
     "build": "turbo run build",
     "test": "turbo run test",
     "lint": "biome check .",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "type": "module",
   "scripts": {
     "prepare": "npx -y lefthook install -f",
-    "clean": "npx -y rimraf -g \"**/.turbo\" \"**/.next\" \"**/dist\" \"**/node_modules\"",
+    "clean": "node scripts/emb-rm.js \"**/.turbo\" \"**/.next\" \"**/dist\" \"**/node_modules\"",
     "build": "turbo run build",
     "test": "turbo run test",
     "lint": "biome check .",
@@ -30,6 +30,7 @@
   "workspaces": [
     "packages/*",
     "demo/*",
+    "scripts",
     "server",
     "tests/integration"
   ],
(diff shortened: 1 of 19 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "type": "module",
   "scripts": {
     "prepare": "npx -y lefthook install -f",
-    "clean": "npx -y rimraf -g \"**/.turbo\" \"**/.next\" \"**/dist\" \"**/node_modules\"",
+    "clean": "node scripts/emb-rm.js '**/.turbo' '**/.next' '**/dist' '**/node_modules'",
```
jpmunz, line 7: this can't be run directly as `emb-rm` as the other examples?

jpmunz (approved):

package.json
overbalance: No because it's the command that wipes out the `node_modules/.bin` folder that holds the installed script file.

PR #1374, opened 2026-06-12
feat(sessions): remove artificial gaps between session part spans at user-session rollover

overbalance:

When a user-session rollover ends one part and starts the next, each span lifecycle call sampled its own clock, so consecutive parts could show a small artificial gap (or overlap) at the boundary.

The manager README had also drifted from the code in this area: it referred to `_terminateUserSession` in six places (the method is named `_rolloverUserSession`), claimed that method is the only path that ends a user session (the part-inactivity timer ends one directly through `endSessionPartInternal`), and described the persisted inactivity deadline as computed from the part end timestamp.

- `StartSessionPartOptions` / `EndSessionPartOptions` gain an optional `timestamp` (epoch millis) anchoring the part span start/end; defaults to now.
- The user-session rollover path samples the clock once and passes the same boundary timestamp to the paired end/start calls, so consecutive parts tile exactly. Real gaps in the part timeline (page load, hidden tab) keep their own timestamps.
- The timestamp anchors the part span only. User-session state and timers always stamp from now (documented on `UserSessionManagerInternal`): `endSessionPartInternal` captures now alongside the anchored stamp and uses it for the persisted inactivity deadline, so an anchored part end can never pull the lazy-expiry window backward.
- Manager README: documents the shared boundary timestamp in the termination flow, renames all `_terminateUserSession` references to `_rolloverUserSession`, replaces the "only path" claim with an accurate description of the part-inactivity path, clarifies when the user-session state storage row is cleared, and describes the inactivity deadline as computed from the actual call time rather than the anchored span end.

## How has this been tested?

Three unit tests, all mutation-verified (removing the anchoring or the now-based deadline routing makes them fail): one passes explicit boundary timestamps with the clock advanced past the boundary and asserts the span end/start times, one triggers a real `endUserSession()` rollover with a part-ended listener that advances the clock mid-end, asserting the rolled part spans still tile, and one ends a part with an anchored timestamp behind now and asserts the persisted inactivity deadline stamps from now. README statements verified against EmbraceUserSessionManager.ts.

## Checklist

- [x] Documentation added
- [x] Tests added

### DIFF
--- a/packages/web-sdk/src/managers/EmbraceUserSessionManager/README.md
+++ b/packages/web-sdk/src/managers/EmbraceUserSessionManager/README.md
@@ -3,6 +3,12 @@
 `EmbraceUserSessionManager` is the source of truth for user-session and
 session-part state across all tabs in this browser.
 
+This document describes the model, the observable behavior, and the stable
+contracts: public API, emitted attributes, storage keys, and configuration.
+It deliberately avoids naming private methods or walking through internal
+algorithms, which drift as the implementation evolves. Read the source for
+those.
+
 ## Two-level model
 
 The manager tracks two distinct things:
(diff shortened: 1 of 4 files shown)


packages/web-sdk/src/managers/EmbraceUserSessionManager/types.ts
```diff
@@ -118,7 +122,15 @@ export interface UserSessionManagerInternal extends UserSessionManager {
 
   getSessionPartId: () => string | null;
 
+  /**
```
jpmunz: not sure we need this comment here, it's referring to implementation details in EmbraceUserSessionManager

packages/web-sdk/src/managers/EmbraceUserSessionManager/README.md
jpmunz, line 1: given the tendency to drift is it useful having this go into specific function names and implementations? Should we rewrite as more of a higher level overview?

overbalance: This doc was created for the development of the session manager, not necessarily for users. Yes, it's likely better to generalize it if we want to keep it for them too.

packages/web-sdk/src/managers/EmbraceUserSessionManager/types.ts
overbalance: I moved it to the source. The shared value info is still important to know. Check out the latest commit.

packages/web-sdk/src/managers/EmbraceUserSessionManager/README.md
overbalance: It's now revised to keep contractual information without using internal names. This doc has been really helpful for keeping agent changes accurate and on task so I definitely think we should keep it.

jpmunz: yep that looks good now

jpmunz (approved):

PR #1392, opened 2026-06-17
EMBR-12526 chore: move session part listener management to EmbraceInstrumentationBase

jpmunz:
Continuing from https://github.com/embrace-io/embrace-web-sdk/pull/1390 moving some duplicated management of session part listeners to the base class so that I can re-use this for the max scroll depth instrumentation

Purposely left `NavigationInstrumentation` out of the refactor since it behaves a bit differently and may end up changing as part of the SPA work anyway

### DIFF
--- a/packages/web-sdk/src/instrumentations/EmbraceInstrumentationBase/EmbraceInstrumentationBase.ts
+++ b/packages/web-sdk/src/instrumentations/EmbraceInstrumentationBase/EmbraceInstrumentationBase.ts
@@ -25,6 +25,7 @@ export abstract class EmbraceInstrumentationBase<
   private _logManager: LogManager;
   private readonly _perf: PerformanceManager;
   private _limitManager: LimitManagerInternal | undefined;
+  protected _isEnabled = false;
 
   protected constructor({
     instrumentationName,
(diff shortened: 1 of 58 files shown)


packages/web-sdk/src/instrumentations/EmbraceInstrumentationBase/EmbraceInstrumentationBase.ts
```diff
@@ -15,6 +15,10 @@ import { OTelPerformanceManager } from '../../utils/index.ts';
 import { InstrumentationAbstract } from '../InstrumentationAbstract/index.ts';
 import type { EmbraceInstrumentationBaseArgs } from './types.ts';
 
+type sessionPartListeners = {
```
joaquin-diaz: Do we have other instances of types staring in lower case? We should probably set up a linter rule for this

packages/web-sdk/src/instrumentations/EmbraceInstrumentationBase/EmbraceInstrumentationBase.ts
```diff
@@ -15,6 +15,10 @@ import { OTelPerformanceManager } from '../../utils/index.ts';
...
+  start?: (() => void) | null;
```
joaquin-diaz: Why is this optional and nullable at the same time?

packages/web-sdk/src/instrumentations/EmbraceInstrumentationBase/EmbraceInstrumentationBase.ts
jpmunz: updated to capitalized, ya isn't linter enforced but seems to be the correct style

packages/web-sdk/src/instrumentations/EmbraceInstrumentationBase/EmbraceInstrumentationBase.ts
jpmunz: ah good call I was copying this from the typing in the instrumentations that were doing this individually but ya doesn't seem needed, removed

joaquin-diaz (approved):

PR #1398, opened 2026-06-29
EMBR-12558 feat: roll over session part on soft navigation

joaquin-diaz:

SPA soft navigations (React Router link clicks, browser back/forward) were not triggering session part rollovers. The Navigation API's `currententrychange` event is the right hook but was unwired.

- Wire `_rolloverSessionPart` to `navigation.currententrychange` via a new `onSoftNavigation` activity listener in `activity.ts`
- Add `Navigation` and `SoftNavigationEvent` interfaces in `common/types.ts` to type the Navigation API without casting; `NavigationHost` now types `navigation` as `Navigation` instead of `EventTarget`
- Add `location: { href: string }` to `NavigationHost` so the handler can detect same-URL replacements without accessing `window.location` directly
- Skip rollover when `event.from.url === navigationHost.location.href` to avoid spurious rollovers from framework hydration (e.g. Next.js App Router fires `currententrychange` via `history.replaceState` during hydration with the same URL)
- Add `navigationHost?: NavigationHost` to `EmbraceUserSessionManagerArgs`; defaults to `window`
- Add 5 unit tests covering: rollover on soft-nav, no-op when no active part, listener cleanup on shutdown, graceful handling of absent Navigation API, no-op on same-URL navigation
- Increase bundle size limit from 55 to 56 KB
- Restructure integration test `ReceivedSpans` from `Record<string, boolean>` to `Record<string, Record<string, { endReason }>>`, keyed by session part ID — enables server-side count validation
- Consolidate `validateThatSessionPartEnded` / `validateThatSessionPartsEnded` into a single `validateThatSessionPartsEnded(expectedCount = 1, userSessionId?)` that polls `/received-spans` for accuracy
- Remove `test.skip` from all three Vite React Router SPA navigation e2e tests

## How has this been tested?

- Unit tests: all 27 tests in `EmbraceSessionPartActivity.test.ts` pass (22 existing + 5 new)
- Integration tests: SPA navigation e2e tests in `vite-react-router-tests.spec.ts` are unskipped and exercise forward links, browser back, and browser forward scenarios; Next.js integration tests all pass (same-URL skip fix prevents spurious rollovers during hydration)

## Checklist

- [x] Tests added
- [ ] Documentation added

### DIFF
--- a/packages/web-sdk/src/common/types.ts
+++ b/packages/web-sdk/src/common/types.ts
@@ -19,9 +19,28 @@ export interface VisibilityStateDocument {
   ) => void;
 }
 
+// Minimal shape of NavigationCurrentEntryChangeEvent (Navigation API, not yet in TypeScript's DOM lib).
+// `from` is optional so test fakes that dispatch a plain Event remain valid.
+export interface SoftNavigationEvent extends Event {
+  readonly from: { url: string };
+}
+
+// Minimal shape of window.navigation (Navigation API, not yet in TypeScript's DOM lib).
+export interface Navigation {
+  addEventListener(
+    type: 'currententrychange',
+    listener: (event: SoftNavigationEvent) => void,
+  ): void;
+  removeEventListener(
+    type: 'currententrychange',
+    listener: (event: SoftNavigationEvent) => void,
+  ): void;
+}
+
 // Useful for testing so that we can pass in a window-like object for soft-navigation detection
 export interface NavigationHost {
-  navigation?: EventTarget;
+  navigation?: Navigation;
+  location: { href: string };
 }
 
 // Useful for testing so that we can pass in a document-like object and change its URL
--- a/packages/web-sdk/src/common/types.ts
+++ b/packages/web-sdk/src/common/types.ts
@@ -19,6 +19,11 @@ export interface VisibilityStateDocument {
   ) => void;
 }
 
+// Useful for testing so that we can pass in a window-like object for soft-navigation detection
+export interface NavigationHost {
+  navigation?: EventTarget;
+}
+
 // Useful for testing so that we can pass in a document-like object and change its URL
 export interface URLDocument {
   URL: string;
--- a/packages/web-sdk/src/common/types.ts
+++ b/packages/web-sdk/src/common/types.ts
@@ -19,24 +19,6 @@ export interface VisibilityStateDocument {
   ) => void;
 }
 
-// Minimal shape of NavigationCurrentEntryChangeEvent (Navigation API, not yet in TypeScript's DOM lib).
-// `from` is optional so test fakes that dispatch a plain Event remain valid.
-export interface SoftNavigationEvent extends Event {
-  readonly from: { url: string };
-}
-
-// Minimal shape of window.navigation (Navigation API, not yet in TypeScript's DOM lib).
-export interface Navigation {
-  addEventListener(
-    type: 'currententrychange',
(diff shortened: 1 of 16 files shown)


packages/web-sdk/src/common/types.ts
```diff
@@ -19,6 +19,30 @@ export interface VisibilityStateDocument {
...
+  addEventListener(
+    type: 'currententrychange',
+    listener: (event: SoftNavigationEvent) => void,
+  ): void;
+  removeEventListener(
+    type: 'currententrychange',
+    listener: (event: SoftNavigationEvent) => void,
+  ): void;
+}
+
```
overbalance: I see Navigation, NavigationCurrentEntryChangeEvent, NavigationHistoryEntry in TS 6.0.3. You can prob remove all this.

server/server.ts
```diff
@@ -152,8 +152,20 @@ const server = createServer((req, res) => {
...
+            receivedSpans[userSessionId] = {};
+          }
           if (sessionPartSpan) {
+            const sessionPartId = sessionPartSpan.attributes.find(
+              (attr) => attr.key === 'emb.session_part_id',
+            )?.value.stringValue;
+            const endReason =
+              sessionPartSpan.attributes.find(
+                (attr) => attr.key === 'emb.session_part_end_reason',
+              )?.value.stringValue ?? 'unknown';
```
overbalance:
```suggestion
            const endReason =
              sessionPartSpan.attributes.find(
                (attr) => attr.key === 'emb.session_part_end_reason',
              )?.value.stringValue ?? undefined;
            if (!endReason) {
              logWarn(
                'emb-session-part received without emb.session_part_end_reason; SDK contract broken?',
              );
            }
```

tests/integration/types.ts
```diff
@@ -4,6 +4,10 @@ declare global {
   }
 }
 
-type ReceivedSpans = Record<string, boolean>;
+type SessionPartInfo = {
+  endReason: string;
```
overbalance:
```suggestion
  endReason?: string;
```
Based on the previous `?? 'unknown'` in server.ts

packages/web-sdk/src/common/types.ts
joaquin-diaz: Thanks! My local was outdated

overbalance (approved):
I recommend expanding the reason value. Otherwise looks great!

packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
```diff
@@ -879,16 +885,33 @@ export class EmbraceUserSessionManager implements UserSessionManagerInternal {
...
+      onSoftNavigation: this._onSoftNavigation,
     });
     this._clearSessionPartInactivityTimer();
     this._clearMaxDurationTimer();
   }
 
+  private readonly _onSoftNavigation = (
+    event: NavigationCurrentEntryChangeEvent,
+  ): void => {
+    // Skip same-URL replacements (e.g. framework hydration via history.replaceState).
```
jpmunz, line 903: Won't we also need to ignore fragment and query string changes here?

joaquin-diaz: `event.from.url` and `this._navigationHost.location.href` include the entire query string so navigating to the same URL (including qs) would be ignored

PR #1401, opened 2026-06-30
feat: add soft navigation performance instrumentation

joaquin-diaz:

Chrome's [Soft Navigations API](https://developer.chrome.com/docs/web-platform/soft-navigations) exposes `soft-navigation` PerformanceObserver entries that identify SPA route transitions detected by the browser heuristic. The SDK had no way to capture these, leaving SPA navigation timing invisible in Embrace.

- Adds `SoftNavigationPerformanceInstrumentation`, a new always-on instrumentation (opt-out via `omit: 'soft-navigation-performance'`) that observes `soft-navigation` PerformanceObserver entries and emits a span per entry. Span name is the new URL; attributes capture `navigationId`, `interactionId`, `startTime`, `duration`, `paintTime`, and `presentationTime`.
- Follows the same `isEntryTypeSupported` guard pattern as `ElementTimingInstrumentation` — silently skips on browsers that don't support the entry type.
- Adds `soft_navigation` to `EmbraceLimitManager` (default cap: 100/session), consistent with `element_timing`/`server_timing`.
- Registers in `setupDefaultInstrumentations` and exposes an omit key in `DefaultInstrumentationConfig`.

### DIFF
--- a/packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
@@ -1,0 +1,106 @@
+import {
+  createPerformanceObserver,
+  isEntryTypeSupported,
+} from '../../../utils/index.ts';
+import { EmbraceInstrumentationBase } from '../../EmbraceInstrumentationBase/index.ts';
+import {
+  KEY_EMB_SOFT_NAVIGATION_DURATION,
+  KEY_EMB_SOFT_NAVIGATION_INTERACTION_ID,
+  KEY_EMB_SOFT_NAVIGATION_NAVIGATION_ID,
+  KEY_EMB_SOFT_NAVIGATION_PAINT_TIME,
+  KEY_EMB_SOFT_NAVIGATION_PRESENTATION_TIME,
+  KEY_EMB_SOFT_NAVIGATION_START_TIME,
+} from './constants.ts';
+import type {
+  PerformanceSoftNavigationTiming,
+  SoftNavigationPerformanceInstrumentationArgs,
+} from './types.ts';
+
+export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumentationBase {
+  private _observer: PerformanceObserver | null = null;
+
+  public constructor({
+    diag,
+    perf,
+    limitManager,
+  }: SoftNavigationPerformanceInstrumentationArgs = {}) {
+    super({
+      instrumentationName: 'SoftNavigationPerformanceInstrumentation',
+      instrumentationVersion: '1.0.0',
+      diag,
+      perf,
+      limitManager,
+      config: {},
+    });
+
+    if (this._config.enabled) {
+      this.enable();
+    }
+  }
+
+  public override enable(): void {
+    if (isEntryTypeSupported('soft-navigation')) {
+      super.enable();
+    } else {
+      this._diag.debug('soft-navigation not supported, skipping');
+    }
+  }
+
+  public override onEnable(): void {
+    if (this._observer) {
+      this._observer.disconnect();
+    }
+
+    this._observer = createPerformanceObserver<PerformanceSoftNavigationTiming>(
+      'soft-navigation',
+      (entry) => this._processEntry(entry),
+      { diag: this._diag },
+    );
+
+    if (!this._observer) {
+      this._isEnabled = false;
+      this._diag.error('failed to enable');
+      return;
+    }
+  }
+
+  public onDisable(): void {
+    if (this._observer) {
+      this._observer.disconnect();
+      this._observer = null;
+    }
+  }
+
+  private _processEntry(entry: PerformanceSoftNavigationTiming): void {
+    if (!this._isEnabled) {
+      return;
+    }
+
+    if (this.limitManager?.limitSoftNavigationEntry()) {
+      return;
+    }
+
+    const span = this.tracer.startSpan(entry.name, {
+      startTime: this.perf.epochMillisFromZeroTime(entry.startTime),
+      attributes: {
+        [KEY_EMB_SOFT_NAVIGATION_NAVIGATION_ID]: entry.navigationId,
+        [KEY_EMB_SOFT_NAVIGATION_INTERACTION_ID]: entry.interactionId,
+        [KEY_EMB_SOFT_NAVIGATION_START_TIME]: this.perf.millisFromZeroTime(
+          entry.startTime,
+        ),
+        [KEY_EMB_SOFT_NAVIGATION_DURATION]: entry.duration,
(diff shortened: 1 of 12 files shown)


overbalance:
@joaquin-diaz From the user's perspective, including `performance` in the instrumentation name might be confused with measuring the quality/speed of soft nav when that's not what it does. what do you think about dropping it?

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/types.ts
```diff
@@ -0,0 +1,13 @@
...
+export type SoftNavigationPerformanceInstrumentationArgs = Pick<
+  EmbraceInstrumentationBaseArgs,
+  'diag' | 'perf' | 'limitManager'
+>;
+
+export type PerformanceSoftNavigationTiming = PerformanceEntry & {
+  navigationId: string;
+  interactionId?: number;
+  paintTime?: number;
+  presentationTime?: number;
```
overbalance: Paint Timing declares `presentationTime` as nullable (`DOMHighResTimeStamp?`), so a not-yet-presented frame yields `null`, not `undefined`. The `!== undefined` guard in `_processEntry` then lets `null` through into a number-typed attribute. It should prob be `presentationTime?: number | null` here with `!= null` guards

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  public override enable(): void {
+    if (isEntryTypeSupported('soft-navigation')) {
+      super.enable();
+    } else {
+      this._diag.debug('soft-navigation not supported, skipping');
+    }
+  }
+
+  public override onEnable(): void {
```
overbalance, line 53: On a disable/enable cycle this rebuilds the observer with `buffered: true`, so entries processed before the pause are re-delivered: duplicate spans, each also burning the limit budget. Same behavior as element-timing. Is that accepted here, or should we dedupe by last-processed `startTime`?

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    this._observer = createPerformanceObserver<PerformanceSoftNavigationTiming>(
+      'soft-navigation',
+      (entry) => this._processEntry(entry),
+      { diag: this._diag },
+    );
+
+    if (!this._observer) {
+      this._isEnabled = false;
+      this._diag.error('failed to enable');
```
overbalance, line 66: `createPerformanceObserver` ends in `catch { return null }`, so this path only ever logs failed without a scope. Passing the caught error to `diag` in the util would fix all four consumers. Should prob fix in a follow up too

packages/web-sdk/src/sdk/setupDefaultInstrumentations.ts
```diff
@@ -81,6 +82,15 @@ export const setupDefaultInstrumentations = (
     );
   }
 
+  if (!config.omit?.has('soft-navigation-performance')) {
+    instrumentations.push(
+      new SoftNavigationPerformanceInstrumentation({
+        ...config['soft-navigation-performance'],
+        limitManager,
```
overbalance, line 88: `limitManager` placed after the spread discards a user-supplied one, while the public args type advertises it. Same as the other instrumentations but we should fix in a follow up

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
```diff
@@ -0,0 +1,106 @@
...
+  private _processEntry(entry: PerformanceSoftNavigationTiming): void {
+    if (!this._isEnabled) {
+      return;
+    }
+
+    if (this.limitManager?.limitSoftNavigationEntry()) {
+      return;
+    }
+
+    const span = this.tracer.startSpan(entry.name, {
```
overbalance: does this need an `emb.type`?

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
joaquin-diaz: I see your point but I don't know how often people would disable/enable. Given that is also an issue for other instrumentations we should probably fix it in `packages/web-sdk/src/utils/performanceObserver/performanceObserver.ts`. Do you think is fine doing it as a follow up?

packages/web-sdk/src/sdk/setupDefaultInstrumentations.ts
joaquin-diaz: Fixed

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
joaquin-diaz: For now we'll process them as performance spans until we figure out what we want to do with it

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/types.ts
joaquin-diaz: Fixed

joaquin-diaz:
> @joaquin-diaz From the user's perspective, including `performance` in the instrumentation name might be confused with measuring the quality/speed of soft nav when that's not what it does. what do you think about dropping it?

What do you mean? The span does represent the performance of a soft navigation. Given that it includes things like presentationTime and duration you get to know how much it took to go from interaction to painting the new page

overbalance:
@joaquin-diaz I'm suggesting that everything we do is technically measuring the performance of a webapp, so the word "performance" is redundant but I defer to you. Nice work!

overbalance (approved):

joaquin-diaz:
> @joaquin-diaz I'm suggesting that everything we do is technically measuring the performance of a webapp, so the word "performance" is redundant but I defer to you. Nice work!

Oh, I see what you're saying now. I think I'd like to keep the "Performance" suffix only to avoid confusing this instrumentation with something that signals soft navigations (which is what we have in activity)

PR #1411, opened 2026-07-06
feat: add soft navigation span polyfill

joaquin-diaz:

`SoftNavigationPerformanceInstrumentation` previously only worked when Chrome was launched with `--enable-features=SoftNavigationHeuristics`, which is not enabled by default. Without it, no soft navigation spans were emitted for SPA route changes.

- Added a polyfill path to `SoftNavigationPerformanceInstrumentation` that activates when the native `soft-navigation` PerformanceObserver entry type is unavailable. The polyfill correlates `PerformanceEventTiming` click entries with Navigation API `currententrychange` events to synthesize a span with `emb.soft_navigation.source = 'polyfill'`.
- Refactored the `navigation` constructor arg to a `navigationHost` pattern (matching `EmbraceUserSessionManager`) so tests can inject a mock without touching `globalThis`.
- Added unit tests for the polyfill path including `getNavigationEventTrigger` and the full click→navigate→span flow.
- Added two headed integration tests: `soft-navigation-polyfill.spec.ts` (standard Chrome, polyfill active) and `soft-navigation-native.spec.ts` (Chrome with `--enable-features=SoftNavigationHeuristics`, native path). Each project in `playwright.config.headed.ts` is scoped via `testMatch` to only its relevant test.
- Added a `test:headed` Turbo task and `test:integration:headed` root script.
- Extended the soft nav demo with query-string (`?tab=details`) and hash (`#section`) navigation items.

## How has this been tested?

- Unit tests: `npm run test` (headless, all pass)
- Headed polyfill test: `npx turbo run test:headed --filter=tests-integration` — verified polyfill span is emitted and `source === 'polyfill'`
- Headed native test: same command, `chromium-soft-nav-heuristics` project — verified `source === 'performance_observer'` and timing attrs (`duration`, `paint_time`, `presentation_time`) are positive

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "server": "npx tsx server/server.ts",
     "commitlint": "commitlint",
     "test:integration": "turbo run test --filter=tests-integration",
+    "test:integration:headed": "turbo run test:headed --filter=tests-integration",
     "test:integration:update-golden": "turbo run update-golden --filter=tests-integration",
     "test:integration:container": "bash scripts/test-integration-podman.sh",
     "test:integration:container:update-golden": "UPDATE_GOLDEN=1 bash scripts/test-integration-podman.sh",
--- a/packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
@@ -32,16 +32,15 @@ import type {
  */
 export function getNavigationEventTrigger(
   navigationTimestamp: number,
-  clickEntries: PerformanceEventTiming[],
+  entry: PerformanceEventTiming,
 ): PerformanceEventTiming | null {
-  return (
-    clickEntries.find(
-      (entry) =>
-        entry.startTime <= navigationTimestamp &&
-        navigationTimestamp <= entry.startTime + entry.duration,
-    ) ?? null
-  );
+  return entry.startTime <= navigationTimestamp &&
+    navigationTimestamp <= entry.startTime + entry.duration
+    ? entry
+    : null;
 }
+
+const PENDING_NAVIGATION_TTL_MS = 60_000;
 
 type PendingNavigation = { timestamp: number; url: string };
 
@@ -53,7 +52,12 @@ export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumenta
   private readonly _navigationHost: NavigationHost;
 
   private readonly _handleCurrentEntryChange = (): void => {
-    const url = this._navigationHost.navigation?.currentEntry?.url ?? '';
+    const url = this._navigationHost.navigation?.currentEntry?.url;
+    if (!url) {
+      this._diag.debug('currententrychange fired with no URL, skipping');
+      return;
+    }
+
     // Store origin-relative time (performance.now() equivalent) so it can be
     // compared directly against PerformanceEventTiming.startTime, which is also
     // origin-relative. getNowMillis() is epoch-based; subtracting getZeroTime()
@@ -89,7 +93,7 @@ export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumenta
       this._usePolyfill = false;
       super.enable();
     } else if (
-      this._navigationHost.navigation != null &&
+      this._navigationHost.navigation &&
       isEntryTypeSupported('event')
     ) {
       this._usePolyfill = true;
@@ -124,10 +128,12 @@ export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumenta
   }
 
   private _enablePolyfill(): void {
-    const nav = this._navigationHost.navigation;
-    if (!nav) {
-      return;
+    if (this._eventObserver) {
+      this._eventObserver.disconnect();
     }
+
+    // biome-ignore lint/style/noNonNullAssertion: enable() is only called when navigationHost.navigation is defined
+    const nav = this._navigationHost.navigation!;
 
     nav.addEventListener('currententrychange', this._handleCurrentEntryChange);
 
@@ -214,8 +220,13 @@ export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumenta
       return;
     }
 
+    this._pendingNavigations = this._pendingNavigations.filter(
+      ({ timestamp }) =>
+        entry.startTime - timestamp < PENDING_NAVIGATION_TTL_MS,
+    );
+
     const index = this._pendingNavigations.findIndex(
-      ({ timestamp }) => getNavigationEventTrigger(timestamp, [entry]) !== null,
+      ({ timestamp }) => getNavigationEventTrigger(timestamp, entry) !== null,
     );
 
     if (index === -1) {
--- a/packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
@@ -1,3 +1,5 @@
+/* eslint-disable baseline-js/use-baseline */
+import type { NavigationHost } from '../../../common/index.ts';
 import { KEY_BROWSER_URL_FULL } from '../../../constants/index.ts';
 import {
   createPerformanceObserver,
@@ -20,13 +22,51 @@ import type {
   SoftNavigationPerformanceInstrumentationArgs,
 } from './types.ts';
 
+/**
+ * Returns the click entry whose processing window contains the given navigation
+ * timestamp, or null if no match is found.
+ *
+ * For pushState navigations, currententrychange fires synchronously during the
+ * click handler, so the timestamp falls within the click entry's
+ * [startTime, startTime + duration] window.
+ */
+export function getNavigationEventTrigger(
+  navigationTimestamp: number,
+  clickEntries: PerformanceEventTiming[],
+): PerformanceEventTiming | null {
+  return (
+    clickEntries.find(
+      (entry) =>
+        entry.startTime <= navigationTimestamp &&
+        navigationTimestamp <= entry.startTime + entry.duration,
+    ) ?? null
+  );
+}
+
+type PendingNavigation = { timestamp: number; url: string };
+
 export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumentationBase {
   private _observer: PerformanceObserver | null = null;
+  private _eventObserver: PerformanceObserver | null = null;
+  private _pendingNavigations: PendingNavigation[] = [];
+  private _usePolyfill = false;
+  private readonly _navigationHost: NavigationHost;
+
+  private readonly _handleCurrentEntryChange = (): void => {
+    const url = this._navigationHost.navigation?.currentEntry?.url ?? '';
+    // Store origin-relative time (performance.now() equivalent) so it can be
+    // compared directly against PerformanceEventTiming.startTime, which is also
+    // origin-relative. getNowMillis() is epoch-based; subtracting getZeroTime()
+    // converts it back to origin-relative.
+    const timestamp = this.perf.getNowMillis() - this.perf.getZeroTime();
+    this._pendingNavigations.push({ timestamp, url });
+  };
 
   public constructor({
     diag,
     perf,
     limitManager,
+    navigationHost = window as NavigationHost,
   }: SoftNavigationPerformanceInstrumentationArgs = {}) {
     super({
       instrumentationName: 'SoftNavigationPerformanceInstrumentation',
@@ -37,6 +77,8 @@ export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumenta
       config: {},
     });
 
+    this._navigationHost = navigationHost;
+
     if (this._config.enabled) {
       this.enable();
     }
@@ -44,13 +86,26 @@ export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumenta
 
   public override enable(): void {
     if (isEntryTypeSupported('soft-navigation')) {
+      this._usePolyfill = false;
       super.enable();
-    } else {
-      this._diag.debug('soft-navigation not supported, skipping');
+    } else if (
+      this._navigationHost.navigation != null &&
+      isEntryTypeSupported('event')
+    ) {
+      this._usePolyfill = true;
+      super.enable();
     }
   }
 
   public override onEnable(): void {
+    if (this._usePolyfill) {
+      this._enablePolyfill();
+    } else {
+      this._enableNative();
+    }
+  }
+
+  private _enableNative(): void {
     if (this._observer) {
       this._observer.disconnect();
     }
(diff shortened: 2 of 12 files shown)


tests/integration/playwright.config.headed.ts
```diff
@@ -0,0 +1,47 @@
...
+    // Chrome with SoftNavigationHeuristics enabled: exercises the native
+    // soft-navigation PerformanceObserver entry type.
+    {
+      name: 'chromium-soft-nav-heuristics',
+      testMatch: '**/e2e-headed/soft-navigation-native.spec.ts',
+      use: {
+        ...devices['Desktop Chrome'],
+        headless: false,
+        launchOptions: {
+          args: ['--enable-features=SoftNavigationHeuristics'],
```
overbalance, line 42: Chrome 150 just shipped SoftNavigationHeuristics so you don't need this anymore

joaquin-diaz: It's actually 151 that ships without the flag
```
Update June 2026: The Chrome team has sent out an [Intent to Ship notification](https://developer.chrome.com/https://groups.google.com/a/chromium.org/g/blink-dev/c/BfSsaY2b_y8/m/ioItMWpYAgAJ) and we are aiming to launch this feature unflagged from Chrome 151
```

overbalance: My mistake. I saw the chrome mcp server version enabled locally but it is using "Chrome For Testing" that is a version ahead: 151 already. https://github.com/microsoft/playwright#cross-browser-support

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
```diff
@@ -20,13 +22,51 @@ import type {
...
+  private readonly _navigationHost: NavigationHost;
+
+  private readonly _handleCurrentEntryChange = (): void => {
+    const url = this._navigationHost.navigation?.currentEntry?.url ?? '';
+    // Store origin-relative time (performance.now() equivalent) so it can be
+    // compared directly against PerformanceEventTiming.startTime, which is also
+    // origin-relative. getNowMillis() is epoch-based; subtracting getZeroTime()
+    // converts it back to origin-relative.
+    const timestamp = this.perf.getNowMillis() - this.perf.getZeroTime();
+    this._pendingNavigations.push({ timestamp, url });
```
overbalance: From 🤖: 
**`_pendingNavigations` grows unbounded.** A `{timestamp, url}` is pushed on every `currententrychange`, but the only in-session removal is the `splice` in `_processClickEntry` when a delivered `click` window contains the timestamp. Wholesale clearing happens only in `onDisable`, which never runs mid-session. There is no cap, TTL, or eviction.

Any navigation not consummated by an observable click becomes permanent garbage:
- back/forward / `history.go()` (traverse) produce no click entry
- programmatic navigation (`router.push` in a `setTimeout`/`fetch` callback, bare `pushState`)
- keyboard-activated links (produce `keydown`/`keyup`, filtered by the `entry.name !== 'click'` guard)
- fast clicks under `durationThreshold: 16`, whose `event` entry is never delivered
- `pushState`+`replaceState` in one handler: two pushes, one `splice`

In a long-lived SPA the array grows monotonically for the tab's lifetime and each real click runs an O(n) `findIndex` over the backlog. Suggest bounding it: evict entries older than a small window (a genuine click->nav match commits within ~one frame) and/or cap length with drop-oldest, with a `diag.debug` on discard. (Stale entries do not cause wrong-URL spans since their timestamps precede any future click window, they are just inert garbage.)

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
```diff
@@ -20,13 +22,51 @@ import type {
...
+
 export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumentationBase {
   private _observer: PerformanceObserver | null = null;
+  private _eventObserver: PerformanceObserver | null = null;
+  private _pendingNavigations: PendingNavigation[] = [];
+  private _usePolyfill = false;
+  private readonly _navigationHost: NavigationHost;
+
+  private readonly _handleCurrentEntryChange = (): void => {
+    const url = this._navigationHost.navigation?.currentEntry?.url ?? '';
```
overbalance: if there is no url we may just want to warn/debug. without it the span is a lot less useful

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
```diff
@@ -68,11 +123,50 @@ export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumenta
     }
   }
 
+  private _enablePolyfill(): void {
```
overbalance, line 129: what do you think about using the guard from line 109 here?

```suggestion
  private _enablePolyfill(): void {
    if (this._eventObserver) {
      this._eventObserver.disconnect();
    }
```

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
```diff
@@ -68,11 +123,50 @@ export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumenta
...
+    const nav = this._navigationHost.navigation;
+    if (!nav) {
```
overbalance: this may be dead code because of the null check in enable()

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
```diff
@@ -20,13 +22,51 @@ import type {
...
+ * Returns the click entry whose processing window contains the given navigation
+ * timestamp, or null if no match is found.
+ *
+ * For pushState navigations, currententrychange fires synchronously during the
+ * click handler, so the timestamp falls within the click entry's
+ * [startTime, startTime + duration] window.
+ */
+export function getNavigationEventTrigger(
+  navigationTimestamp: number,
+  clickEntries: PerformanceEventTiming[],
```
overbalance: how do you get more than one clickEntry here?

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
joaquin-diaz: I'll add a cap to clean up things older than ~60s should be safe enough

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
joaquin-diaz: It's from an old approach. Good catch!

package.json
```diff
@@ -12,13 +12,14 @@
...
     "validate": "npm run validate:versions && npm run licenses:check && turbo run validate",
-    "licenses": "patch-package && turbo run licenses",
+    "licenses": "turbo run licenses",
     "licenses:check": "npm run licenses && git diff --exit-code -- packages/web-sdk/THIRD_PARTY_NOTICES.txt packages/web-cli/THIRD_PARTY_NOTICES.txt",
     "dev": "turbo run dev --filter=embrace-web-sdk-react-demo...",
     "demo": "npm i && npm run demo -w demo/frontend",
     "server": "npx tsx server/server.ts",
     "commitlint": "commitlint",
     "test:integration": "turbo run test --filter=tests-integration",
+    "test:integration:headed": "turbo run test:headed --filter=tests-integration",
```
jpmunz, line 22: would this only be for local runs? Can you file a backlog item for getting this working in CI?

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
```diff
@@ -37,20 +81,35 @@ export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumenta
...
+      this._usePolyfill = false;
+      super.enable();
+    } else if (
+      this._navigationHost.navigation &&
+      isEntryTypeSupported('event')
+    ) {
+      this._usePolyfill = true;
       super.enable();
-    } else {
-      this._diag.debug('soft-navigation not supported, skipping');
```
jpmunz, line 49: I think still worth the debug log in the case that neither of these are available

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
```diff
@@ -68,11 +127,52 @@ export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumenta
...
+    // biome-ignore lint/style/noNonNullAssertion: enable() is only called when navigationHost.navigation is defined
+    const nav = this._navigationHost.navigation!;
+
+    nav.addEventListener('currententrychange', this._handleCurrentEntryChange);
+
+    this._eventObserver = createPerformanceObserver<PerformanceEventTiming>(
+      'event',
+      (entry) => this._processClickEntry(entry),
+      // SPA routers commit navigations synchronously inside the click handler,
+      // so the entry duration is well under the 104ms default threshold. 16ms
```
jpmunz: where does the 104 number come from?

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
```diff
@@ -20,13 +22,55 @@ import type {
...
+  private readonly _navigationHost: NavigationHost;
+
+  private readonly _handleCurrentEntryChange = (): void => {
+    const url = this._navigationHost.navigation?.currentEntry?.url;
+    if (!url) {
+      this._diag.debug('currententrychange fired with no URL, skipping');
+      return;
+    }
+
+    // Store origin-relative time (performance.now() equivalent) so it can be
```
jpmunz: would going by the `timestamp` on the currententrychange event not work?

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
```diff
@@ -110,4 +210,56 @@ export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumenta
...
+    const [matched] = this._pendingNavigations.splice(index, 1);
+    this._emitPolyfillSpan(entry, matched.timestamp, matched.url);
+  }
+
+  private _emitPolyfillSpan(
+    clickEntry: PerformanceEventTiming,
+    navigationTimestamp: number,
+    url: string,
+  ): void {
+    if (this.limitManager?.limitSoftNavigationEntry()) {
```
jpmunz, line 254: should also check `if (!this._isEnabled) {`

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
```diff
@@ -68,11 +127,52 @@ export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumenta
...
+    if (this._eventObserver) {
+      this._eventObserver.disconnect();
+    }
+
+    // biome-ignore lint/style/noNonNullAssertion: enable() is only called when navigationHost.navigation is defined
+    const nav = this._navigationHost.navigation!;
+
+    nav.addEventListener('currententrychange', this._handleCurrentEntryChange);
+
+    this._eventObserver = createPerformanceObserver<PerformanceEventTiming>(
```
jpmunz, line 142: could you add an explanation as to why this is used over a document event listener on 'click'?

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
```diff
@@ -110,4 +210,56 @@ export class SoftNavigationPerformanceInstrumentation extends EmbraceInstrumenta
...
+  private _processClickEntry(entry: PerformanceEventTiming): void {
+    if (!this._isEnabled) {
+      return;
+    }
+
+    if (entry.name !== 'click') {
+      return;
+    }
+
+    this._pendingNavigations = this._pendingNavigations.filter(
```
jpmunz, line 228: I think worth a comment here that this isn't strictly needed for the logic but is in place to prevent `_pendingNavigations` growing unbounded

package.json
joaquin-diaz: Yes. Added https://app.notion.com/p/embraceio/Run-headed-e2e-tests-in-CI-39c7e3c99852807098e6e4e6b665783b?source=copy_link

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
joaquin-diaz: Added https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/observe#durationthreshold

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
joaquin-diaz: Definitively, good catch

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/SoftNavigationPerformanceInstrumentation.ts
joaquin-diaz: Added

jpmunz (approved):

PR #1412, opened 2026-07-07
feat: soft navigation support for Web Vitals

joseph-embrace:

We want to capture web vitals metrics in soft navigations to give greater visibility into the performance of SPAs.

- Upgraded to web-vitals@6.0 which contains the soft navigation changes
- Add `reportSoftNavs` option on WebVitalsInstrumentation (default: `true`)
- Update WebVitalsInstrumentation to handle soft navigations
  - Don't report TTFB for soft navs (soft navigation reporting is disabled on the TTFB listeners, since web-vitals only reports a dummy zero value there)
  - Use `navigationURL` when provided by web-vitals, rather than `attributedPage.fullURL`
  - Add `browser.web_vital.navigation_id` and `browser.web_vital.interaction_id` attributes
- Add a `vite-test-harness` platform app (`tests/integration/platforms/`) that the headed web vitals spec targets, so the tests are independent of the demo pages

## How has this been tested?

Unit tests, plus a headed integration spec (`soft-navigation-web-vitals.spec.ts`) asserting the emitted web vital values, URLs, and navigation IDs across hard and soft navigations.

## Checklist

- [ ] Documentation added
- [x] Tests added


packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -328,6 +339,12 @@ export class WebVitalsInstrumentation extends EmbraceInstrumentationBase {
   }
 
   private _emitWebVital(metric: MetricWithAttribution): void {
+    // Soft navigations have no server request, so the library reports a dummy zero TTFB.
```
joaquin-diaz: Is this a bug in their library? Weird they would do that on purpose. Shouldn't we not subscribe to `reportSoftNavs` for the TTFB listener instead?

demo/frontend/src/webVitalsStore.ts
```diff
@@ -0,0 +1,107 @@
...
+const ATTR_PAGE_PATH = 'emb.page.path';
+
+const asString = (value: unknown): string | undefined =>
+  typeof value === 'string' ? value : undefined;
+
+const asNumber = (value: unknown): number | undefined =>
+  typeof value === 'number' ? value : undefined;
+
+const isMetricName = (value: string | undefined): value is MetricName =>
+  value === 'cls' ||
```
jpmunz: would also want 'tbd'

though since we're checking for `WEB_VITAL_LOG_TYPE` do we need this at all?

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
jpmunz: Similar question for LCP, given product requirements around ICP should we not subscribe to `reportSoftNavs` for that one?

jpmunz (commented):
should the base branch here be `jwynn/fix/navigation-id-type` rather than `main`?

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
joseph-embrace: @joaquin-diaz it seems like a difference in opinion rather than a bug. they are saying "the equivalent of TTFB happened, and it was instant". we could do the same but IMO it's just noise.

@jpmunz as far as web-vitals is concerned, ICP is not a metric. I will need to set up another observer to capture that separately.

joaquin-diaz: Makes sense, I would not subscribe to TTFB for soft navs then?

demo/frontend/src/webVitalsStore.ts
joseph-embrace: Sorted, thanks!

tests/integration/playwright.config.headed.ts
```diff
@@ -20,6 +23,15 @@ export default defineConfig({
       url: 'http://localhost:3001/health-check',
       reuseExistingServer: true,
     },
+    // Port 4848 so the test instance (which must send telemetry to the mock
```
joaquin-diaz: For now I'd say that is fine to use the demo as a client but in the future we should have an independent page so we don't break this test by changing the demo

tests/integration/playwright.config.headed.ts
```diff
@@ -34,9 +46,16 @@ export default defineConfig({
...
+      testMatch: [
+        '**/e2e-headed/soft-navigation-native.spec.ts',
+        '**/e2e-headed/soft-navigation-web-vitals.spec.ts',
+      ],
       use: {
         ...devices['Desktop Chrome'],
+        // Playwright's bundled Chromium ships the older soft-navigation API
+        // that web-vitals does not support; installed Chrome has the current
+        // one (PerformanceSoftNavigation.getLargestInteractionContentfulPaint).
+        channel: 'chrome',
```
joaquin-diaz, line 59: Should we add a comment with the required installed chrome version?

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/types.ts
```diff
@@ -33,6 +33,16 @@ export interface WebVitalsInstrumentationConfig extends InstrumentationConfig {
   includeRawAttribution?: boolean;
   applyCustomLogRecordData?: (logRecord: LogRecord) => void;
 
+  /**
+   * @experimental
```
jpmunz: should we leave off the experimental now that the soft navigation logic has been merged into the main web-vitals branch?

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -231,6 +238,9 @@ export class WebVitalsInstrumentation extends EmbraceInstrumentationBase {
     this._urlDocument = urlDocument ?? window.document;
     this._urlAttribution = urlAttribution;
     this._includeRawAttribution = includeRawAttribution;
+    this._reportSoftNavs = reportSoftNavs;
```
jpmunz: does this need to be saved as an instance variable, should we instead be branching based on `_softNavsActive` in all cases?

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -353,6 +371,12 @@ export class WebVitalsInstrumentation extends EmbraceInstrumentationBase {
   }
 
   private _emitWebVital(metric: MetricWithAttribution): void {
+    // web-vitals reports a hardcoded dummy value of zero for TTFB on soft navigations. We don't
```
jpmunz: should we instead set `reportSoftNavs` to false on the listener for TTFB?

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -365,9 +389,26 @@ export class WebVitalsInstrumentation extends EmbraceInstrumentationBase {
...
+          ? {
+              [KEY_BROWSER_WEB_VITAL_INTERACTION_ID]:
+                metric.navigationInteractionId,
+            }
+          : {}),
         ...(attributedPage
           ? {
-              [KEY_BROWSER_URL_FULL]: attributedPage.fullURL,
+              // The navigationURL emitted by web-vitals is authoritative for which URL the metric
+              // belongs to, since it was captured when the metric occurred. This only applies to
```
jpmunz: given this should the check be `(metric.navigationType === 'soft-navigation' && this._softNavsActive)`? What would `metric.navigationURL` represent in the case that `metric.navigationType` was not 'soft-navigation'?

tests/integration/playwright.config.headed.ts
joseph-embrace: Done - thanks!

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
joseph-embrace: Yes, great find! I added `_softNavsActive` later to fix a bug that the integration tests found. `_reportSoftNavs` can go.

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
joseph-embrace: I'm on the fence about this. I chose this path because it's explicitly documenting the fact that we don't want zero TTFB values emitted. I will change it to `reportSoftNavs: false` for TTFB since you and @joaquin-diaz have both mentioned it now. We will just be relying on integration tests preventing regressions rather than the code itself.

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
joseph-embrace: Ah sorry the second part of that comment is out of date, again because the integration tests found a bug 🙃 we do want to use `metric.navigationURL` for hard navs because it's the source of truth for the URL state at the time the metric was recorded by the browser.

tests/integration/playwright.config.headed.ts
joseph-embrace: No this is a good point I hadn't considered. I will create a separate test-only page so the demo page is free to change.

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/types.ts
joseph-embrace: Done, thx!

tests/integration/platforms/vite-test-harness/src/main.tsx
```diff
@@ -0,0 +1,193 @@
+// Deterministic page for the headed integration specs. The texts, button
```
joseph-embrace: @joaquin-diaz I followed your pattern of having a very stripped-down test page here rather than using the demo app. Are you happy with that?

joseph-embrace: btw I've created a separate harness from the existing vite-react-router just to make it clear that the purpose of this harness is to generate deterministic outputs. Whereas the vite-react-router harness is more about exercising how the SDK interacts with React Router.

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
jpmunz: @joseph-embrace I think fine having both for redundancy if we want the extra guard, e.g. don't setup the report soft nav listener for TTFB in the first place and also avoid doing the emit here

jpmunz (approved):

overbalance (approved):

joaquin-diaz (approved):

tests/integration/platforms/vite-test-harness/src/main.tsx
joaquin-diaz: Yep, perfect. We can do a follow up to share some of the code

PR #1414, opened 2026-07-07
EMBR-12903 fix(licenses): sort discovered license files for deterministic cross-platform notices

overbalance:

`generate-license-file` picks whichever license file `glob` returns first when a package ships more than one (e.g. `import-in-the-middle` ships both `LICENSE` and a `LICENSE-3rdparty.csv` dependency manifest). `glob`'s order and casing differ across filesystems, so `THIRD_PARTY_NOTICES.txt` regenerated differently on macOS vs Linux CI and tripped the `licenses:check` drift gate. It also meant the committed notices listed the CSV manifest for `import-in-the-middle` instead of its actual license.

- Patch `generate-license-file` to sort candidate license files case-insensitively before picking, so the same file wins on every platform (upstream: TobyAndToby/generate-license-file#743). A plain `.sort()` isn't enough here: `glob` yields `LICENSE` on Linux but `license` on macOS, so code-unit ordering still diverges.
- Apply the patch via `patch-package`. Since `.npmrc` sets `ignore-scripts=true` (no `postinstall`), the patch runs from the root `licenses` script, which `licenses:check` / `validate` / CI all funnel through.
- Regenerate `packages/web-sdk/THIRD_PARTY_NOTICES.txt`: `import-in-the-middle` now correctly resolves to its Apache license.

Chosen over depending on the fork directly because its package lives in a subdirectory of a private bun monorepo and isn't published to npm.

## How was this tested?

- `npm run licenses:check` passes with no drift; macOS output matches the CI Linux blob byte-for-byte.
- `npm run lint` passes.

## Checklist

- [x] Covered by the existing `licenses:check` drift gate
- [x] No changes to public API


patches/generate-license-file+4.2.1.patch
```diff
@@ -0,0 +1,24 @@
...
+@@ -21,7 +21,18 @@ const licenseFile = (inputs) => tslib_1.__awaiter(void 0, void 0, void 0, functi
+         cwd: directory,
+         maxDepth: 1,
+     });
+-    const filteredLicenseFiles = licenseFiles.filter(file => !exports.extensionDenyList.includes((0, path_1.extname)(file)));
++    // glob returns matches in filesystem order and casing that differ between
++    // case-sensitive and case-insensitive filesystems (e.g. "LICENSE" on Linux
++    // vs "license" on macOS). Sort case-insensitively so the chosen file is
++    // deterministic across platforms; the base name (e.g. "license") sorts
++    // ahead of decorated variants (e.g. "license-3rdparty.csv").
```
jpmunz, line 14:
```suggestion
+    // ahead of decorated variants (e.g. "license-3rdparty.csv").
+    // remove once https://github.com/TobyAndToby/generate-license-file/pull/743 is merged and released
```

joaquin-diaz (approved):

PR #1416, opened 2026-07-07
feat: reset SDK zero time on soft navigation

joaquin-diaz:

The SDK's zero time ("when did the user start viewing what's on screen") only reset on hard navigations and bfcache restores — soft navigations left it pinned to the first page load, so timings after an SPA route change were measured from the wrong starting point.

Fixing that surfaced a pre-existing arithmetic bug in `OTelPerformanceManager`: raw Performance API values are always relative to `timeOrigin`, but the epoch conversion added them to `getZeroTime()` instead, double-counting the gap between the two whenever they differed (bfcache restores, prerendered pages). Example: 10 minutes in bfcache, then a click 2s after restore → timestamp reported ~10 minutes in the future. Meanwhile `millisFromZeroTime` (backing "time since view start" attributes) was a no-op stub that returned the raw offset unchanged.

- Zero time now also resets on soft navigations (`currententrychange`), skipping same-URL replacements such as framework hydration.
- Fixed the epoch conversion (`epochMillisFromZeroTime` → `epochMillisFromOrigin`, now `timeOrigin + offset`) and implemented `millisFromZeroTime` to actually rebase offsets onto zero time, clamped to 0. Renamed the reset state to `updateZeroTimeMillis`/`_zeroTimeMillis` accordingly.
- `ElementTimingInstrumentation` now anchors its span start and its `render_time`/`load_time`/`start_time` attributes to zero time instead of the original navigation.
- `DocumentLoadInstrumentation` converts span/paint-event timestamps explicitly through `perf` instead of relying on OTel's `TimeInput` magnitude heuristic.
- Removed unused `PerformanceManager` methods (`getNowHRTime`, `millisSinceHRTime`).
- Added `src/utils/PerformanceManager/README.md` documenting the two reference frames (time origin vs zero time), with a worked example and a per-instrumentation catalog; each instrumentation now carries a `// Timing frame:` comment, and CLAUDE.md points to the README.

## How has this been tested?

Unit tests added for: the zero-time reset on `pageshow`/`currententrychange` (including the same-URL no-op), `millisFromZeroTime` rebasing/clamping for prerender and bfcache cases, and element-timing span/attribute anchoring after a reset. Full suite (933 tests) and every instrumentation suite touching `perf` pass.

## Checklist

- [x] Documentation added
- [x] Tests added

### DIFF
--- a/packages/web-sdk/src/instrumentations/clicks/ClicksInstrumentation/ClicksInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/clicks/ClicksInstrumentation/ClicksInstrumentation.ts
@@ -16,7 +16,6 @@ import type { ClicksInstrumentationArgs } from './types.ts';
 
 import { getHTMLElementFriendlyName } from './utils.ts';
 
-// Timing frame: time-origin (see src/utils/PerformanceManager/README.md)
 export class ClicksInstrumentation extends EmbraceInstrumentationBase {
   private readonly _onClickHandler: (event: MouseEvent) => void;
 
--- a/packages/web-sdk/src/instrumentations/clicks/ClicksInstrumentation/ClicksInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/clicks/ClicksInstrumentation/ClicksInstrumentation.ts
@@ -16,6 +16,7 @@ import type { ClicksInstrumentationArgs } from './types.ts';
 
 import { getHTMLElementFriendlyName } from './utils.ts';
 
+// Timing frame: time-origin (see src/utils/PerformanceManager/README.md)
 export class ClicksInstrumentation extends EmbraceInstrumentationBase {
   private readonly _onClickHandler: (event: MouseEvent) => void;
 
--- a/packages/web-sdk/src/instrumentations/clicks/ClicksInstrumentation/ClicksInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/clicks/ClicksInstrumentation/ClicksInstrumentation.ts
@@ -64,7 +64,7 @@ export class ClicksInstrumentation extends EmbraceInstrumentationBase {
               ),
               'tap.coords': `${event.x.toString()},${event.y.toString()}`,
             },
-            this.perf.epochMillisFromZeroTime(event.timeStamp),
+            this.perf.epochMillisFromOrigin(event.timeStamp),
           );
         }
       } catch (e) {
--- a/packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
@@ -229,7 +229,7 @@ export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<Docu
       }
 
       if (!this.getConfig().ignorePerformancePaintEvents) {
-        addSpanPerformancePaintEvents(rootSpan);
+        addSpanPerformancePaintEvents(rootSpan, this.perf);
       }
 
       this._addCustomAttributesOnSpan(
@@ -257,7 +257,7 @@ export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<Docu
         hasKey(entries, performanceName) &&
         typeof entries[performanceName] === 'number'
       ) {
-        span.end(entries[performanceName]);
+        span.end(this.perf.epochMillisFromOrigin(entries[performanceName]));
       } else {
         span.end();
       }
@@ -370,7 +370,7 @@ export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<Docu
       const span = this.tracer.startSpan(
         spanName,
         {
-          startTime: entries[performanceName],
+          startTime: this.perf.epochMillisFromOrigin(entries[performanceName]),
         },
         parentSpan ? trace.setSpan(context.active(), parentSpan) : undefined,
       );
--- a/packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
@@ -70,7 +70,6 @@ const ATTR_HTTP_RESPONSE_CACHE_REVALIDATED = 'http.response.cache_revalidated'; 
 const ATTR_HTTP_REQUEST_INCOMPLETE = 'http.request.incomplete'; // Request started but didn't complete
 const ATTR_HTTP_REQUEST_PREVENTED = 'http.request.prevented'; // Request never started (blocked)
 
-// Timing frame: time-origin (see src/utils/PerformanceManager/README.md)
 export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<DocumentLoadInstrumentationConfig> {
   private readonly _onDocumentLoaded: () => void;
   private _performanceCollected = false;
--- a/packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
@@ -70,6 +70,7 @@ const ATTR_HTTP_RESPONSE_CACHE_REVALIDATED = 'http.response.cache_revalidated'; 
 const ATTR_HTTP_REQUEST_INCOMPLETE = 'http.request.incomplete'; // Request started but didn't complete
 const ATTR_HTTP_REQUEST_PREVENTED = 'http.request.prevented'; // Request never started (blocked)
 
+// Timing frame: time-origin (see src/utils/PerformanceManager/README.md)
 export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<DocumentLoadInstrumentationConfig> {
   private readonly _onDocumentLoaded: () => void;
   private _performanceCollected = false;
--- a/packages/web-sdk/src/instrumentations/element-timing/ElementTimingInstrumentation/ElementTimingInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/element-timing/ElementTimingInstrumentation/ElementTimingInstrumentation.ts
@@ -92,9 +92,15 @@ export class ElementTimingInstrumentation extends EmbraceInstrumentationBase {
         [KEY_EMB_TYPE]: EMB_TYPES.ElementTiming,
         [KEY_EMB_ELEMENT_TIMING_IDENTIFIER]: entry.identifier,
         [KEY_EMB_ELEMENT_TIMING_ELEMENT]: entry.element?.tagName?.toLowerCase(),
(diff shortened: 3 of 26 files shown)


packages/web-sdk/src/utils/PerformanceManager/types.ts
```diff
@@ -2,7 +2,7 @@ import type { HrTime } from '@opentelemetry/api';
 
 export interface PerformanceManager {
   getNowHRTime: () => HrTime;
```
jpmunz: maybe worth simplifying this interface so we have less concepts to think about together, it looks like `getNowHRTime` and `millisSinceHRTime` are unused. `getZeroTime` is only used in OTelPerformanceManager so could be marked private

packages/web-sdk/src/sdk/initSDK.ts
```diff
@@ -127,9 +127,14 @@ export const initSDK = (
...
-      updatePageShowMillis(window.performance.timeOrigin + event.timeStamp);
-    });
+    // Both events mark the start of a new "view" the user is looking at (a bfcache
+    // restore, or a client-side soft navigation), so both reset the SDK's zero time.
+    const resetZeroTime = (event: Event) =>
+      updateZeroTimeMillis(window.performance.timeOrigin + event.timeStamp);
+
+    window.addEventListener('pageshow', resetZeroTime);
+    // eslint-disable-next-line baseline-js/use-baseline
+    window.navigation?.addEventListener('currententrychange', resetZeroTime);
```
jpmunz: I think this should use the same condition as https://github.com/embrace-io/embrace-web-sdk/blob/1ac4df06d2da53d38bbec68dad2f153e421807b1/packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts#L905, otherwise we'd be resetting the zero time in cases where we don't consider a soft nav as having occurred

packages/web-sdk/src/instrumentations/element-timing/ElementTimingInstrumentation/ElementTimingInstrumentation.ts
```diff
@@ -86,7 +86,7 @@ export class ElementTimingInstrumentation extends EmbraceInstrumentationBase {
     const span = this.tracer.startSpan(entry.identifier, {
       // Span start anchors to navigation start so the span duration equals "time from navigation
       // until the element was rendered" — useful for analyzing page-load render timing.
-      startTime: this.perf.epochMillisFromZeroTime(0),
+      startTime: this.perf.epochMillisFromOrigin(0),
```
jpmunz: given the comment above should this bet `getZeroTime()` instead? And span.end below remain as `millisFromZeroTime(entry.startTime)` ?

packages/web-sdk/src/instrumentations/element-timing/ElementTimingInstrumentation/ElementTimingInstrumentation.ts
```diff
@@ -100,9 +100,9 @@ export class ElementTimingInstrumentation extends EmbraceInstrumentationBase {
       },
     });
     if (entry.loadTime > 0) {
-      span.addEvent('load', this.perf.epochMillisFromZeroTime(entry.loadTime));
+      span.addEvent('load', this.perf.epochMillisFromOrigin(entry.loadTime));
```
jpmunz, line 110: similarly, shouldn't this remain as `millisFromZeroTime` ?

packages/web-sdk/src/sdk/initSDK.ts
joaquin-diaz: Right. I'll be simplifying this later in NavigationInstrumentation so we don't have a lot of different hooks into `currententrychange` spread across the codebase.

packages/web-sdk/src/instrumentations/element-timing/ElementTimingInstrumentation/ElementTimingInstrumentation.ts
joaquin-diaz: Yeah, this being `getZeroTime` is correct if we want to use our own zero time as start time. But the rest of the `epochMillisFromOrigin` are fine given that they describe when something happened. `startTime`, `loadTime`, and all the other timestamps in performance entries are measured relative to performance.timeOrigin. Here's an example to help you visualize it. I find it really hard to do it without putting real values to these.

```
time origin = 13/07/26 12:00:00
our zero time = 13/07/26 12:01:00 (user soft navigated after 1 min)

-- page renders an element using elementiming. Element takes 1s to render. entry.renderTime = 1m 1s = entry.loadTime

span.startTime = 13/07/26 12:01:00 (our zero time)
span.endTIme = 13/07/26 12:01:01 (this.perf.epochMillisFromOrigin(entry.startTime))
```

jpmunz: hmm but then the span duration will be calculated on 2 different scales since startTime is based on zero time but endTime is based on time origin, is that fine?

packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
```diff
@@ -257,7 +258,7 @@ export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<Docu
         hasKey(entries, performanceName) &&
         typeof entries[performanceName] === 'number'
       ) {
-        span.end(entries[performanceName]);
+        span.end(this.perf.epochMillisFromOrigin(entries[performanceName]));
```
joaquin-diaz, line 260: This was working fine but relying in internal OTel logic that converts span.startTime and span.endTime to epoch:

```
if (time < performance.timeOrigin / 2) {
    return hrTime(time);          // treated as an offset → OTel adds performance.timeOrigin itself
} else {
    return millisToHrTime(time);  // treated as absolute epoch ms → taken as-is
}
```

Doing it explicitly here to be consistent with other instrumentations

packages/web-sdk/src/instrumentations/element-timing/ElementTimingInstrumentation/ElementTimingInstrumentation.ts
joaquin-diaz: No this needs to be an epoch

packages/web-sdk/src/instrumentations/element-timing/ElementTimingInstrumentation/ElementTimingInstrumentation.ts
joaquin-diaz: Span duration is always end - start. There aren't 2 different scales, we just say that startTime = our zero time and endTime = actual end time. We shorten the span in relation to the start time. I did change it so the attribute values are also relative to our zero time

packages/web-sdk/src/instrumentations/clicks/ClicksInstrumentation/ClicksInstrumentation.ts
```diff
@@ -16,6 +16,7 @@ import type { ClicksInstrumentationArgs } from './types.ts';
 
 import { getHTMLElementFriendlyName } from './utils.ts';
 
+// Timing frame: time-origin (see src/utils/PerformanceManager/README.md)
```
jpmunz: we could maybe leave off these top level ones since we have the summary in `src/utils/PerformanceManager/README.md`, also we can have cases where the instrumentation emits across multiple timing frames

packages/web-sdk/src/utils/PerformanceManager/README.md
```diff
@@ -0,0 +1,250 @@
...
+Every timing value the SDK handles lives in one of two frames:
+
+- **Time origin.** `performance.timeOrigin` is the epoch timestamp of the
+  original hard navigation. It is fixed for the entire life of the page: it
+  does not change on prerendering activation, on a bfcache restore (the
+  performance clock keeps counting through the freeze), or on a soft
+  navigation. Every raw value the browser hands out — `entry.startTime`,
+  `entry.renderTime`, `entry.loadTime`, `event.timeStamp`,
+  `performance.now()`, web-vitals attribution times — is defined as
+  *milliseconds since time origin*.
```
jpmunz: I think enough just to say it doesn't change, also I'd refer to the `DOMHighResTimeStamp` definition:

```suggestion
- **Time origin.** `performance.timeOrigin` is the epoch timestamp of the
  original hard navigation. It is fixed for the entire life of the page. Every time value from the Performance API is a [DOMHighResTimeStamp](https://developer.mozilla.org/en-US/docs/Web/API/Performance_API/High_precision_timing#domhighrestimestamp) defined as *milliseconds since time origin*.
```

packages/web-sdk/src/utils/PerformanceManager/README.md
```diff
@@ -0,0 +1,250 @@
...
+| 12:01:01.000 | Checkout hero image finishes rendering | the browser reports `entry.renderTime = 61,000` — still measured from time origin; the soft navigation does not reset the browser's clock |
+| 12:01:01.500 | SDK processes the entry and emits telemetry | — |
+
+Each method, applied to this scenario:
+
+| Call | Result | What it's for here |
+| --- | --- | --- |
+| `epochMillisFromOrigin(61_000)` | `1,700,000,061,000` (12:01:01.000) | the span **end** — the real instant the render happened |
+| `getZeroTime()` | `1,700,000,060,000` (12:01:00.000) | the span **start** — anchoring the span to the current view's start so its duration reads "time from view start until render" |
+| `millisFromZeroTime(61_000)` | `61,000 - 60,000 = 1,000` | the `render_time` **attribute** — "rendered 1s into the current view" |
```
jpmunz, line 72: in this example this would redundant right since the span's duration would already be equal to 1s?

packages/web-sdk/src/utils/PerformanceManager/README.md
```diff
@@ -0,0 +1,250 @@
...
+| Telemetry field | Source | Method |
+| --- | --- | --- |
+| log timestamp (CLS) | `attribution.largestShiftTime` | `epochMillisFromOrigin` |
+| log timestamp (INP) | `attribution.interactionTime` | `epochMillisFromOrigin` |
+| log timestamp (LCP / FCP / TTFB) | — (no per-event offset in attribution) | `getNowMillis()` at emission |
+| `browser.web_vital.value` / `.delta` | computed by the `web-vitals` library | none — see below |
+| TTFB sub-part attributes (`redirect`, `domainLookup`, `tcpConnection`, `tlsNegotiation`, `serverResponse`, `unattributed`) | differences of `PerformanceNavigationTiming` fields | none — pure durations |
+| raw attribution body (`includeRawAttribution`) | `metric.attribution` primitives, verbatim | none — intentionally a raw debug dump |
+
+`metric.value` for time-based vitals is "time since navigation start" as
```
jpmunz, line 180: this bit will likely be changing

jpmunz (approved):
lgtm, just some comments on the documentation

packages/web-sdk/src/utils/PerformanceManager/README.md
joaquin-diaz: Yes but generally we send the raw values as attribute anyways

packages/web-sdk/src/utils/PerformanceManager/README.md
joaquin-diaz: yep, we can update it once we work on web vitals for soft navs

PR #1418, opened 2026-07-08
feat: centralize soft navigation handling in EmbracePageManager

joaquin-diaz:

Route (`ux.surface`) span creation/ending was split across two independent, unordered actors: `EmbraceUserSessionManager` reacting to the Navigation API's `currententrychange` to roll over session parts, and `NavigationInstrumentation` reacting to react-router calling `setCurrentRoute` directly. Since neither ordering was guaranteed, a soft navigation could produce a spurious extra route span or incorrectly split a route span across a session-part boundary. React-router integrations also had a direct, unnecessary dependency on `NavigationInstrumentation`.

**Note:** this removes public API surface (`getNavigationInstrumentation`, `createReactRouterNavigationInstrumentation`, `NavigationInstrumentation.setCurrentRoute`/`setInstrumentationType`) — a breaking change for any manual integrator calling these directly.

- `EmbracePageManager` is now the single source of truth for the current route and the sole listener of the Navigation API. On soft navigation, it rolls over the session part (via a newly-public `rolloverSessionPartInternal` on `EmbraceUserSessionManager`) before setting a placeholder route, so the outgoing route span and outgoing session-part span end in the correct order/batch.
- React-router integrations (`withEmbraceRouting`, `withEmbraceRoutingLegacy`, `listenToRouterChanges`) now only call `page.setCurrentRoute(...)` — no more dependency on `NavigationInstrumentation`.
- `NavigationInstrumentation` is simplified to a pure mirror: it reacts to `page.addRouteChangedListener` (same url → rename span in place, different url → end old span + start new one) and to session-part-ended (so a route span never outlives the session part it started in). It's now wired up automatically as a default instrumentation via `setupDefaultInstrumentations`/`initSDK` instead of being manually constructed by consumers.
- `EmbraceUserSessionManager` no longer knows the Navigation API exists — that listening moved to `EmbracePageManager`.
- Dropped the `emb.instrumentation` attribute on route spans (which react-router flavor reported the route); removed the now-dead `EMB_NAVIGATION_INSTRUMENTATIONS` enum.


packages/web-sdk/src/instrumentations/navigation/NavigationInstrumentation/NavigationInstrumentation.ts
```diff
@@ -38,116 +36,101 @@ export class NavigationInstrumentation extends EmbraceInstrumentationBase {
...
+
+  private readonly _cleanPath = (path: string): string =>
+    this._shouldCleanupPathOptionsFromRouteName
+      ? path.replace(PATH_OPTIONS_RE, '')
+      : path;
+
+  // A route span must not outlive the session part it started in — e.g. the
+  // tab backgrounds, or the session ends, with no further navigation to
+  // trigger _onRouteChanged. Only the ending side is wired up: resuming a
+  // span on session-part-start was deliberately dropped in favor of relying
```
jpmunz: this would be a behaviour change from before then right?

If we have:

Hard load Page A -> background tab -> foreground tab -> soft navigate to Page B -> background tab


Previously we would end up with:
* SurfaceSpan1: Hard load Page A -> background tab
* SurfaceSpan2: foreground tab -> soft navigate to Page B
* SurfaceSpan3: soft navigate to Page B -> background tab

Now is the behaviour this?:
* SurfaceSpan1: Hard load Page A -> background tab
* SurfaceSpan2: soft navigate to Page B -> background tab

packages/web-sdk/src/sdk/setupDefaultInstrumentations.ts
```diff
@@ -97,6 +98,12 @@ export const setupDefaultInstrumentations = (
     );
   }
 
+  if (!config.omit?.has('navigation')) {
```
jpmunz: should we be allowing this to be omittable?

packages/web-sdk/src/managers/EmbracePageManager/README.md
```diff
@@ -0,0 +1,94 @@
+# EmbracePageManager
+
+`EmbracePageManager` is the single source of truth for the current route,
+and the sole listener of the Navigation API for soft-navigation detection.
+Centralizing both in one place is what avoids the race that used to exist
+between the session-part lifecycle and route-span creation: previously two
+independent actors (the session manager reacting to the Navigation API, and
+the navigation instrumentation reacting to route reports) could observe the
+same soft navigation in either order. Now there's one actor driving both.
```
jpmunz: I think this context is not needed once we complete the refactor 

```suggestion
```

packages/web-sdk/src/managers/EmbracePageManager/README.md
```diff
@@ -0,0 +1,94 @@
...
+- **`NavigationInstrumentation`** — subscribes to `addRouteChangedListener`
+  (mirrors route changes into `ux.surface` spans: same url → rename in place,
+  different url → end the old span and start a new one) and to the
+  session-part-ended listener (ends the open route span so it never outlives
+  the session part it started in — e.g. the tab backgrounds with no further
+  navigation). It never listens for session-part-*start*: resuming a span is
+  handled purely by the next route report, not by the session-part
+  lifecycle.
+- **`EmbraceUserSessionManager`** — owns session-part start/end/rollover, but
+  no longer knows the Navigation API exists.
```
jpmunz:
```suggestion
- **`EmbraceUserSessionManager`** — owns session-part start/end/rollover, no interaction with Navigation API
```

packages/web-sdk/src/managers/EmbracePageManager/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+- a new route is reported (`_onRouteChanged` sees a different url), or
+- the session part it started in ends, for *any* reason (soft nav,
+  backgrounding, inactivity, manual `endUserSession()`).
+
+Resuming a span on the next session-part-*start* was deliberately dropped:
+if the same route is still current when a new part starts, the next
+`setCurrentRoute` report (or, for frameworks that only report on navigation,
+none at all) is what determines whether a span reopens — not the session
+part boundary itself.
```
jpmunz:
```suggestion
Route spans are not started on the next session-part-*start*:
if the same route is still current when a new part starts, the next
`setCurrentRoute` report (or, for frameworks that only report on navigation,
none at all) is what determines whether a span reopens — not the session
part boundary itself.
```

packages/web-sdk/src/managers/EmbracePageManager/EmbracePageManager.ts
```diff
@@ -47,11 +67,62 @@ export class EmbracePageManager implements PageManager {
...
+    // NavigationInstrumentation ends its route span from the session-part-
+    // ended listener, which EmbraceUserSessionManager fires before the
+    // outgoing session-part span itself ends — so the route span is already
+    // queued by the time EmbraceSessionPartBatchedSpanProcessor flushes the
+    // old part's batch. Setting the new route first would end the outgoing
+    // route span too late (after that flush) and start the new placeholder
+    // span before the rollover, so the session-part-ended listener would
+    // incorrectly close the placeholder instead of the outgoing span.
+    this._userSessionManager?.rolloverSessionPartInternal({
+      endReason: 'web_soft_navigation',
```
jpmunz, line 124: since we're rolling over wouldn't there only ever be a single reason for both the end and start?

jpmunz (commented):
For:

> Note: this removes public API surface (getNavigationInstrumentation, createReactRouterNavigationInstrumentation, NavigationInstrumentation.setCurrentRoute/setInstrumentationType) — a breaking change for any manual integrator calling these directly.

Since we're not doing a major version bump can we make these non-breaking? Since Navigation Instrumentation is being setup automatically now if we switched these to no-ops would we preserve the same behaviour as today?

For:
> React-router integrations (withEmbraceRouting, withEmbraceRoutingLegacy, listenToRouterChanges) now only call page.setCurrentRoute(...) — no more dependency on NavigationInstrumentation.

Does this collide with the new soft nav detection? E.g. would we create 1 route span from the soft nav detection and then end up creating a 2nd from the React-router?

packages/web-sdk/src/instrumentations/navigation/NavigationInstrumentation/NavigationInstrumentation.ts
joaquin-diaz: That's updated now. It should behave the same as it was doing when the react instrumentations where in charge of ending the route spans

packages/web-sdk/src/sdk/setupDefaultInstrumentations.ts
joaquin-diaz: Nope, not anymore given that it's doing session part rollover. We can wait for user feedback for another ways of disabling or changing how soft nav works

packages/web-sdk/src/managers/EmbracePageManager/EmbracePageManager.ts
joaquin-diaz: Yeah but I think it doesn't hurt to keep the option open for different reasons just in case

joaquin-diaz:
@jpmunz Updated the readme so it represents current state instead of how we got there or other changes

packages/web-sdk/src/managers/EmbracePageManager/EmbracePageManager.ts
```diff
@@ -47,11 +69,59 @@ export class EmbracePageManager implements PageManager {
...
+  private readonly _onSoftNavigation = (
+    event: NavigationCurrentEntryChangeEvent,
+  ): void => {
+    // Skip same-URL replacements (e.g. framework hydration via history.replaceState).
+    // eslint-disable-next-line baseline-js/use-baseline
+    if (event.from.url === this._navigationHost.location.href) {
+      return;
+    }
+
+    updateZeroTimeMillis(window.performance.timeOrigin + event.timeStamp);
```
jpmunz: I think maybe for consistency we should have:

```suggestion
    updateZeroTimeMillis(epochMillisFromOrigin(event.timeStamp));
```

just so that nothing outside of the PerformanceManager is referencing timeOrigin directly

packages/web-sdk/src/managers/EmbracePageManager/README.md
```diff
@@ -0,0 +1,95 @@
...
+  `setCurrentRoute` call notifies subscribers via `addRouteChangedListener`.
+  On construction, it sets the initial route from the current location,
+  since `currententrychange` never fires for the page's own initial entry.
+  On a soft navigation, it rolls over the active session part
+  (`EmbraceUserSessionManager.rolloverSessionPartInternal`) before reporting
+  the new route, and resets the SDK's zero time (`updateZeroTimeMillis`) —
+  the same reset `initSDK` performs on `pageshow` for bfcache restores.
+- **React-router integrations** (`withEmbraceRouting`,
+  `withEmbraceRoutingLegacy`, `listenToRouterChanges`) — call
+  `page.setCurrentRoute({ path, url })` once they've resolved the templated
```
jpmunz, line 19: Does this have a possibility of colliding with the soft nav detection? E.g. if this is set from the router instrumentation first and then later the soft nav detection doesn't trigger because the url hasn't updated from the page managers perspective? If this is triggered by the router *without* a soft nav being detected at all would it end up creating a 2nd route span in the same session part?

jpmunz (approved):
A couple follow-up comments but overall lgtm

packages/web-sdk/src/managers/EmbracePageManager/README.md
joaquin-diaz: Good thinking, I had to validate this. Calling `setCurrentRoute` does not create a route span directly. We have two guards preventing this:

1. Prevents creating a new route id in EmbracePageManager
```
public setCurrentRoute = (route: Route) => {
  if (!this._currentRoute || this._currentRoute.url !== route.url) {
    this._currentPageId = generateUUID();
  }
  
  ..
```

2. Prevents creating a route span on a same URL
```
private readonly _onRouteChanged = (route: Route): void => {
  if (!this._config.enabled) {
    return;
  }

  if (this._currentRouteSpan && this._currentRouteSpanUrl === route.url) {
    // Same url as the currently open span: either the templated path just
    // resolved (rename in place), or a redundant re-render — no-op if the
    // path hasn't actually changed.
    if (this._currentRouteSpanPath !== route.path) {
      this._renameCurrentRouteSpan(route.path);
    }
    return;
  }
  
  this._endRouteSpan();
  this._startRouteSpan(route);
};
```

PR #1420, opened 2026-07-09
fix(web-vitals): use `PerformanceEntry.startTime` for web vitals log timestamps where possible

joseph-embrace:

We have an inconsistency with how the log timestamps are set:

- CLS uses `metric.attribution.largestShiftTime`
- INP uses `metric.attribution.interactionTime`
- Everything else uses the time that web-vitals emitted the metric (`this.perf.getNowMillis()`)

Metrics can be emitted very long after they occur, so we are not capturing the correct timestamp for other metrics.

I believe the log timestamp should be the time when the metric actually started. For INP this is still `interactionTime`. For CLS it's the start of the first layout shift (which is also the start of the session window). For everything else it's the `PerformanceEntry.startTime` for the most recent entry*.

*<sup>in practice, most metrics only have one entry</sup>

⚠️ I have also bumped the bundle size limit from 56 -> 60 KB.

## How has this been tested?

Unit tests

## Checklist

- [ ] Documentation added
- [x] Tests added

### DIFF
--- a/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.test.ts
+++ b/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.test.ts
@@ -23,6 +23,10 @@ import {
   KEY_EMB_TYPE,
 } from '../../../constants/index.ts';
 import { EmbracePageManager } from '../../../managers/index.ts';
+import {
+  _resetZeroTimeMillisForTesting,
+  updateZeroTimeMillis,
+} from '../../../utils/PerformanceManager/index.ts';
 import type { WebVitalListeners, WebVitalOnReport } from './types.ts';
 import { WebVitalsInstrumentation } from './WebVitalsInstrumentation.ts';
 
@@ -66,6 +70,7 @@ describe('WebVitalsInstrumentation', () => {
   afterEach(() => {
     clock.restore();
     instrumentation.disable();
+    _resetZeroTimeMillisForTesting();
   });
 
   it('should report CLS metrics as a log record', () => {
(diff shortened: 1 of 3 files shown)


packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -312,18 +312,32 @@ export class WebVitalsInstrumentation extends EmbraceInstrumentationBase {
...
+      const windowStart = metric.entries.length
+        ? Math.min(...metric.entries.map((entry) => entry.startTime))
+        : metric.attribution.largestShiftTime;
+      if (windowStart !== undefined) {
+        return this.perf.epochMillisFromZeroTime(windowStart);
+      }
+    }
+
+    // For other metrics, use the startTime of the last entry.
+    const entryStartTime = metric.entries[metric.entries.length - 1]?.startTime;
```
jpmunz: in this case since we're going by the last entry would we want to take `startTime` + `duration` to indicate when the last entry that was relevant to the vital calculation completed?

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -312,18 +312,32 @@ export class WebVitalsInstrumentation extends EmbraceInstrumentationBase {
...
-    }
-
+    // For INP use interactionTime, which is the start time of the user's interaction
     if (metric.name === 'INP' && metric.attribution.interactionTime) {
       return this.perf.epochMillisFromZeroTime(
         metric.attribution.interactionTime,
       );
     }
 
+    // For CLS use the first layout shift's startTime, which is the beginning of
```
jpmunz, line 322: Wouldn't we want to pick the time of the largest shift as representative of when the vital occurred? E.g. if a minor shift occurred at T=1s that on its own would've given a Good CLS rating but then was followed by a major shift at T=10s that caused the vital to have a Poor rating

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.test.ts
```diff
@@ -159,6 +159,69 @@ describe('WebVitalsInstrumentation', () => {
...
+    instrumentation = new WebVitalsInstrumentation({
+      diag,
+      perf,
+      listeners: mockWebVitalListeners,
+      urlAttribution: false,
+    });
+
+    const metricReportFunc = lcpStub.getCall(0).args[0] as WebVitalOnReport;
+
+    // Tick the click forward so the event startTime and report time are different.
```
overbalance:
```suggestion
    // Tick the clock forward so the event startTime and report time are different.
```

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
overbalance: Seeing it again, I agree with the previous logic too. If we were to attach a screenshot to the timeline showing the biggest contributor to CLS, I would expect to see the log event at the start of the largest shift.

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
overbalance: I think LCP makes sense as startTime and not completion. Also, it looks like we need a branch for TTFB because startTime (and +duration) are both inaccurate. We want ~~responseEnd~~ responseStart for that one

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
joseph-embrace: The value of CLS is the sum of all `LayoutShift.value` that occurred during the largest session window. The largest shift within that window is sometimes interesting, but it's not important on its own. If you were investigating what elements/scripts contributed to CLS, you would want to investigate around the start of the window - not the start of a single shift in that window.

image

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
joseph-embrace: I will update TTFB's start time to be `activationStart`, since that's [the "zero" point](https://github.com/GoogleChrome/web-vitals/blob/main/src/onTTFB.ts#L78-L81). If you sum the start time and `responseStart`, you should get the value/duration.

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -312,18 +312,32 @@ export class WebVitalsInstrumentation extends EmbraceInstrumentationBase {
...
+    // the session window. Fall back to largestShiftTime if the entries aren't
+    // available for some reason.
+    if (metric.name === 'CLS') {
+      const windowStart = metric.entries.length
```
overbalance:
```suggestion
      const windowStart = metric.entries.length > 0
```

We have a biome rule to be explicit about length Boolean. I'm surprised it didn't catch it

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.test.ts
joseph-embrace: Thanks!

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
joseph-embrace: Thanks - fixed!

overbalance (approved):

PR #1421, opened 2026-07-09
fix(web-vitals): update `PerformanceSoftNavigationTiming.navigationId` to match the spec

joseph-embrace:

Using `number` to [match the spec](https://wicg.github.io/soft-navigations/#sec-pe-extension) means we have to do less type-juggling when we implement soft navigation support for web vitals.

Screenshot 2026-07-09 at 6 55 43 PM

- Changed the type of `PerformanceSoftNavigationTiming.navigationId` from `string` to `number`

## How has this been tested?

Type checks

## Checklist

- [ ] Documentation added
- [ ] Tests added

### DIFF
--- a/packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/types.ts
+++ b/packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/types.ts
@@ -9,7 +9,7 @@ export type SoftNavigationPerformanceInstrumentationArgs = Pick<
 };
 
 export type PerformanceSoftNavigationTiming = PerformanceEntry & {
-  navigationId: string;
+  navigationId: number;
   interactionId?: number;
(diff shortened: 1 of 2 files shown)


jpmunz (approved):

packages/web-sdk/src/instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/types.ts
```diff
@@ -6,7 +6,7 @@ export type SoftNavigationPerformanceInstrumentationArgs = Pick<
 >;
 
 export type PerformanceSoftNavigationTiming = PerformanceEntry & {
-  navigationId: string;
+  navigationId: number;
```
overbalance, line 12: we had discussed mapping navigationIds to uuids so they are unique across tabs (and possibly beneficial in the db?) was that settled?

joaquin-diaz (approved):

PR #1431, opened 2026-07-17
EMBR-12823 feat(soft-navigation): correlate spans and logs within a soft-navigation window

joaquin-diaz:

Soft-navigation spans record that a navigation happened, but not what else occurred during it (network requests, user interactions, log messages, etc.), making it hard to see the surrounding activity for a given soft navigation.

- Adds a `SignalBuffer`: a bounded, in-memory rolling buffer of span/log signals, recorded as they start.
- `SignalCorrelationSpanProcessor` and `SignalCorrelationLogRecordProcessor` record eligible spans/logs (along with their `emb.type`) into the shared buffer.
- `SoftNavigationPerformanceInstrumentation` looks up the buffer for its span's time window and stamps `emb.soft_navigation.span_ids`/`log_ids`, plus parallel `span_id_types`/`log_id_types` arrays so the backend knows which table each id lives in.


packages/web-sdk/src/processors/SignalCorrelationSpanProcessor/SignalCorrelationSpanProcessor.ts
```diff
@@ -0,0 +1,58 @@
...
+  ReadableSpan,
+  Span,
+  SpanProcessor,
+} from '@opentelemetry/sdk-trace';
+import { EMB_TYPES, KEY_EMB_TYPE } from '../../constants/index.ts';
+import { KEY_EMB_SOFT_NAVIGATION_SOURCE } from '../../instrumentations/soft-navigation-performance/SoftNavigationPerformanceInstrumentation/constants.ts';
+import type { SignalBuffer } from '../utils/SignalBuffer.ts';
+import type { SignalCorrelationSpanProcessorArgs } from './types.ts';
+
+const isSoftNavigationSpan = (span: ReadableSpan | Span): boolean =>
```
overbalance: This can go in the new utils folder

packages/web-sdk/src/processors/SignalCorrelationSpanProcessor/index.ts
```diff
@@ -0,0 +1,3 @@
+import { SignalCorrelationSpanProcessor } from './SignalCorrelationSpanProcessor.ts';
+
+export { SignalCorrelationSpanProcessor };
```
overbalance:
```suggestion
export { SignalCorrelationSpanProcessor } from './SignalCorrelationSpanProcessor.ts';
```

packages/web-sdk/src/processors/SignalCorrelationLogRecordProcessor/index.ts
```diff
@@ -0,0 +1,3 @@
+import { SignalCorrelationLogRecordProcessor } from './SignalCorrelationLogRecordProcessor.ts';
+
+export { SignalCorrelationLogRecordProcessor };
```
overbalance:
```suggestion
export { SignalCorrelationLogRecordProcessor } from './SignalCorrelationLogRecordProcessor.ts';
```

packages/web-sdk/src/processors/utils/SignalBuffer.ts
```diff
@@ -0,0 +1,86 @@
...
+  spanTypes: string[];
+  logIds: string[];
+  logTypes: string[];
+}
+
+// Best-effort correlation ceiling: a window still open 60s after the newest
+// recorded signal will under-report early signals, since those signals have
+// already been evicted by then.
+const DEFAULT_MAX_AGE_MILLIS = 60_000;
+const DEFAULT_MAX_ENTRIES = 4096;
```
overbalance, line 27: I can file a ticket to hook this into a real Limiter in the future

packages/web-sdk/src/processors/utils/index.ts
```diff
@@ -0,0 +1,3 @@
+import { SignalBuffer } from './SignalBuffer.ts';
+
+export { SignalBuffer };
```
overbalance:
```suggestion
export { SignalBuffer } from './SignalBuffer.ts';
```

packages/web-sdk/src/sdk/initSDK.ts
```diff
@@ -415,6 +430,16 @@ const setupTraces = ({
     new UserSpanProcessor({ userManager }),
     new PageSpanProcessor({ pageManager }),
     new SpanScrubProcessor({ attributeScrubbers }),
+    // WHY: placement in this chain does not affect the stamp: every
```
overbalance:
```suggestion
    // Placement in this chain does not affect the stamp: every
```

packages/web-sdk/src/sdk/initSDK.ts
```diff
@@ -471,12 +496,22 @@ const setupLogs = ({
...
   pageManager,
   visibilityDoc,
+  signalBuffer,
 }: SetupLogsArgs) => {
   const finalLogProcessors: LogRecordProcessor[] = [
     ...logProcessors,
     new UserSessionLogRecordProcessor({
       userSessionManager,
     }),
+    // WHY: must run after UserSessionLogRecordProcessor, which stamps the
```
overbalance:
```suggestion
    // Must run after UserSessionLogRecordProcessor, which stamps the
```

packages/web-sdk/src/processors/SignalCorrelationSpanProcessor/SignalCorrelationSpanProcessor.ts
```diff
@@ -0,0 +1,58 @@
...
+    this._buffer = buffer;
+  }
+
+  public forceFlush(): Promise<void> {
+    return Promise.resolve(undefined);
+  }
+
+  public onStart(span: Span): void {
+    try {
+      if (isSoftNavigationSpan(span) || isSessionPartSpan(span)) {
```
overbalance, line 27: Because this functionality has been renamed to be generic, you may consider moving this filter, so the buffer can be used by other parts of the system.

packages/web-sdk/src/processors/utils/SignalBuffer.ts
joaquin-diaz: You mean using the `EmbraceLimitManager`? I think that's more for what we emit to what we capture. It would still be one attribute with X amount of entries. I think it's fine to keep it in here

packages/web-sdk/src/processors/SignalCorrelationSpanProcessor/SignalCorrelationSpanProcessor.ts
joaquin-diaz: I agree but this is the only place where we have enough context for filtering unless we start storing more info which I would like to avoid unless we need it. For now I think it's fine to leave it in here and we can see if at some point we need to correlate the soft nav span with something. Session parts are always correlated via session part id so we'll always want this filter in here

packages/web-sdk/src/processors/utils/SignalBuffer.ts
```diff
@@ -0,0 +1,86 @@
...
+  // point it's recorded (e.g. some spans settle their type only later).
+  readonly type?: string;
+}
+
+export interface SignalBufferArgs {
+  maxAgeMillis?: number;
+  maxEntries?: number;
+}
+
+export interface SignalWindowResult {
```
jpmunz, line 16: any specific reason to split these into separate arrays? Could it work as just a `SignalEntry[]`?

packages/web-sdk/src/processors/SignalCorrelationLogRecordProcessor/SignalCorrelationLogRecordProcessor.ts
```diff
@@ -0,0 +1,46 @@
...
+  private readonly _buffer: SignalBuffer;
+
+  public constructor({ buffer }: SignalCorrelationLogRecordProcessorArgs) {
+    this._buffer = buffer;
+  }
+
+  public onEmit(logRecord: SdkLogRecord): void {
+    try {
+      const uid = logRecord.attributes[ATTR_LOG_RECORD_UID];
+      // Stamped upstream by UserSessionLogRecordProcessor; if that ordering
```
jpmunz, line 24: probably cleaner overall to have this as a separate processor but the need to rely on the ordering is a bit unfortunate, we could consider having `UserSessionLogRecordProcessor` take an optional buffer argument and run this block in there when it is set instead

packages/web-sdk/src/processors/SignalCorrelationLogRecordProcessor/SignalCorrelationLogRecordProcessor.ts
```diff
@@ -0,0 +1,46 @@
...
+      if (typeof uid === 'string' && uid.length > 0) {
+        this._buffer.record({
+          kind: 'log',
+          id: uid,
+          startTime: hrTimeToMilliseconds(logRecord.hrTime),
+          type: logRecord.attributes[KEY_EMB_TYPE] as string | undefined,
+        });
+      }
+    } catch (e) {
+      diag.error('failed to record log for soft-navigation correlation', e);
```
jpmunz:
```suggestion
      diag.error('failed to record log for signal correlation', e);
```

packages/web-sdk/src/processors/SignalCorrelationSpanProcessor/SignalCorrelationSpanProcessor.ts
jpmunz: another option could be to have SignalBuffer on construction take a list of emb.types to ignore and have this check within `record` instead

joaquin-diaz: Soft nav span doesn't have a set emb.type since we use the generic performance span type, and it would make us ignore all those alongside the soft nav span

packages/web-sdk/src/processors/SignalCorrelationLogRecordProcessor/SignalCorrelationLogRecordProcessor.ts
joaquin-diaz: Yeah, I don't like it depending on ordering but I also don't want to not be a processor in case in the future we also depend on a processor other than `UserSessionLogRecordProcessor`. I'll move it to the bottom of the list so we don't have to think too much about ordering and we just keep it last for now

overbalance (approved):

jpmunz (approved):

packages/web-sdk/src/processors/utils/SignalBuffer.ts
joaquin-diaz: That's the output that gets jammed into the soft nav span. I think it's easier to query it later if is a plain array of ids. Internally it is stored as SignalEntry[]:

```
private readonly _entries: SignalEntry[] = [];
```

PR #1441, opened 2026-07-23
EMBR-13244 fix: web vital session_part_id attribution race

joaquin-diaz:

`emb.session_part_id` on `browser.web_vital` log records (INP, CLS) could be stamped incorrectly — empty, or attributed to the wrong session part — because it was resolved from whichever session part happened to be live at the moment the log was emitted, not at the time the metric's underlying event actually occurred. `web-vitals` backdates INP/CLS timestamps to the real interaction/shift time but defers the actual report to its own internal listener, which can fire well after the fact, once the session part has already ended.

- `EmbraceUserSessionManager` now keeps a rollover history (part id + start time), cleared once the enclosing user session ends, and exposes `getSessionPartIdAt(timestampEpochMillis)` to resolve the part active at an arbitrary past timestamp. Eviction is tied to user-session end rather than tab visibility so the history covers as much ground as possible before being reset.
- `WebVitalsInstrumentation` resolves `emb.session_part_id` from the metric's own event timestamp via that new method, instead of relying on the currently active session part.
- `UserSessionLogRecordProcessor` honors a pre-stamped `emb.session_part_id` when present, falling back to the live read otherwise.

## How has this been tested?

- Added unit tests on `EmbraceUserSessionManager` covering: multi-rollover resolution, falling back when a timestamp predates any recorded rollover, falling back when no part has ever started, resolving to the most-recently-created part even after it has closed, history persisting across a mere visibility restore, and history clearing once the user session ends.
- Added `WebVitalsInstrumentation` tests reproducing the original race (metric event time before a rollover, log emitted after) and confirming the fix, independent of `urlAttribution`.
- `npm run check` and the full `npm run test` suite pass.

## Checklist

- [ ] Documentation added (not applicable — internal SDK behavior only)
- [x] Tests added

### DIFF
--- a/packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
+++ b/packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
@@ -112,6 +112,12 @@ export class EmbraceUserSessionManager implements UserSessionManagerInternal {
   private _hasStoredState = false;
 
   private _activeSessionPartId: string | null = null;
+  // Rollover history backing getSessionPartIdAt, cleared once the tab
+  // becomes visible again
+  private _sessionPartHistory: Array<{
+    id: string;
+    startTimeEpochMillis: number;
+  }> = [];
   // Held for the part's lifetime so getUserSessionAttributes() reads
   // during the part return its starting value, even if another tab bumps
   // the shared counter while this tab is mid-disengagement.
@@ -340,6 +346,17 @@ export class EmbraceUserSessionManager implements UserSessionManagerInternal {
     return this._activeSessionPartId;
   }
 
+  public getSessionPartIdAt(timestampEpochMillis: number): string | null {
+    for (let i = this._sessionPartHistory.length - 1; i >= 0; i--) {
+      if (
+        this._sessionPartHistory[i].startTimeEpochMillis <= timestampEpochMillis
+      ) {
+        return this._sessionPartHistory[i].id;
+      }
+    }
+    return this.getSessionPartId();
+  }
+
   public getSessionPartSpan(): ExtendedSpan | null {
     return this._sessionPartSpan;
   }
@@ -400,6 +417,10 @@ export class EmbraceUserSessionManager implements UserSessionManagerInternal {
     this._activeSessionPartCounts = {};
     this._nextSessionPartCounts = {};
     this._sessionPartSpan = span;
+    this._sessionPartHistory.push({
+      id: activeSessionPartId,
+      startTimeEpochMillis: timestamp ?? this._perf.getNowMillis(),
+    });
 
     this._startSessionPartInactivityTimer();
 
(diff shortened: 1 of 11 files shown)


overbalance (commented):
Talked offline about moving the part eviction to user session ended

packages/web-sdk/src/processors/UserSessionLogRecordProcessor/UserSessionLogRecordProcessor.ts
```diff
@@ -31,15 +31,19 @@ export class UserSessionLogRecordProcessor implements LogRecordProcessor {
     const userSessionId = this._userSessionManager.getUserSessionId() ?? '';
     const previousUserSessionId =
       this._userSessionManager.getPreviousUserSessionId() ?? '';
+    // If the log record already has a session part id, we use that
+    const sessionPartId =
+      logRecord.attributes[KEY_EMB_SESSION_PART_ID] ??
+      this._userSessionManager.getSessionPartId() ??
```
jpmunz, line 37: maybe this makes it too broad a change but could avoid having the special case in `WebVitalsInstrumentation` if instead this was `this._userSessionManager.getSessionPartId(logRecord.timestamp)`. It seems logical to always do but maybe there's some cases I'm not considering?

packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
```diff
@@ -112,6 +112,13 @@ export class EmbraceUserSessionManager implements UserSessionManagerInternal {
   private _hasStoredState = false;
 
   private _activeSessionPartId: string | null = null;
+  // Rollover history backing getSessionPartIdAt, cleared once the user
```
jpmunz, line 115: since we have it can we use `SignalBuffer` here? Not sure if it's cleaner but that would also let us avoid exposing `getSessionPartIdAt` as part of the interface if we instead passed the signal buffer around to the spots that need it

packages/web-sdk/src/processors/UserSessionLogRecordProcessor/UserSessionLogRecordProcessor.ts
joaquin-diaz: We just talked about this with Jared actually, at a glance it would make sense to do this with all logs but I think we should do it as a follow up with some deeper research just to validate that it always makes sense

packages/web-sdk/src/managers/EmbraceUserSessionManager/EmbraceUserSessionManager.ts
joaquin-diaz: It would require some extra work because I moved the clean up of the history to when the user session ends, instead of session part, and we don't currently have a listener for that. Having said that I think I'd like to keep the session part history closer to where sessions are defined in case we need to expand what we keep in the history instead of just ids like now.

~I do think is worth moving the getSessionPartIdAt to the internal user manager interface so it's not exposed to users (like the other part methods that are hidden to the user)~ Already there. Is not exposed to users

jpmunz (approved):

overbalance (approved):

PR #1442, opened 2026-07-23
EMBR-13084 feat(instrumentation-dom-state): emit one dom-state log at part end with the first-view fold capture

overbalance:
> First of a two-PR stack; #1487 (the part-end tree-walk cap) is based on this branch and retargets to `main` automatically when this merges.

We have no visibility into the structure and size of the DOM. Viewport size, above-the-fold image count, document dimensions, element count and tree depth let us correlate layout and rendering characteristics with user-experience issues.

Add `DOMStateInstrumentation` (enabled by default, removable via `defaultInstrumentationConfig.omit`). It emits one `dom-state` log per session part end, carrying element count, average depth, and the document box, all measured at the part end itself. Unmeasurable keys are dropped rather than zeroed.

The first part-end log after the page's fold capture also carries `dom_state.images_above_fold.count`, `.viewport_height`, `.viewport_width`, and `.timestamp` (capture-time epoch millis; the log's own timestamp is the part end). The fold is captured at most once per page, only while a session part is engaged: at the load event, at the first part start after a load nobody watched, or at attach to an already loaded page. Load detection reads the navigation entry's `loadEventStart`; `readyState` is the fallback where no entry exists.

Session part correlation is stamped by the emit-time processor; since parts end synchronously, the ending part is the part the fold was captured in. A capture held across a `disable()` is dropped outright, and a failed part-end emit spends the fold capture rather than carrying it into a later part. The whole capture path is guarded, so a capture failure costs at most the fold keys, never the caller. Also exposes `getNavigationEntry()` on `PerformanceManager` so the load-event read goes through the injected clock.

## Known issues

Accepted trade-offs from the review passes are tracked in EMBR-13714.

## How has this been tested?

Unit tests cover the capture-moment decisions, the one-attempt-per-page latch, the pending-capture lifecycle across part ends and disable, capture- and emit-failure containment, the capture-time timestamp attribute, and the fold geometry. Integration golden files regenerated (300 tests pass across Chromium, Firefox, WebKit). Lint and check pass.

## Checklist

- [ ] Documentation added
- [x] Tests added


packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
```diff
@@ -0,0 +1,200 @@
...
+      eventName: DOM_STATE_EVENT_NAME,
+      severityNumber: SeverityNumber.INFO,
+      attributes,
+    });
+  }
+
+  private _measure(): DOMMetrics {
+    const { body, documentElement } = document;
+
+    const documentHeight = Math.max(
```
joaquin-diaz: There's no way of knowing by definition which one of these would be the one that represents the document height?

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
```diff
@@ -0,0 +1,200 @@
...
+    if (document.readyState === 'complete') {
+      this._onLoadHandler();
+    } else {
+      window.addEventListener('load', this._onLoadHandler, { once: true });
+    }
+
+    this.setSessionPartListeners({
+      end: () => {
+        try {
+          this._sendLog(
```
joaquin-diaz: Should we have an attribute to differentiate when the snapshot was taken? You're currently sending one at load and one at part end but there's no way of knowing which one is each

overbalance: I initially had an attribute for that but between the timestamp, the context of other events, and the existence of the above the fold attr, there's a lot of implicit timing information.

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
overbalance: This was taken line for line from lux. It covers lots of old browsers and document modes. Good callout though. I'll upgrade to baseline widely available: document.scrollingElement?.scrollHeight ?? 0

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
```diff
@@ -0,0 +1,188 @@
...
+  } {
+    let elementCount = 0;
+    let totalDepth = 0;
+
+    if (!root) {
+      return { elementCount, totalDepth };
+    }
+
+    const stack: Array<{ element: Element; depth: number }> = [
+      { element: root, depth: 1 },
```
jpmunz, line 255: should clarify with Andy, or maybe this is answerable from the lux implementation, should the root be considered a depth of 0 or 1?

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
```diff
@@ -0,0 +1,188 @@
...
+      attributes,
+    });
+  }
+
+  private _measure(): DOMMetrics {
+    const documentHeight = document.scrollingElement?.scrollHeight ?? 0;
+    const documentWidth = document.scrollingElement?.scrollWidth ?? 0;
+
+    const { elementCount, totalDepth } = this._traverse(
+      document.scrollingElement,
```
jpmunz: should this be `document.documentElement`? What is the distinction with using `scrollingElement`?

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
```diff
@@ -0,0 +1,188 @@
...
+      for (let i = 0; i < children.length; i++) {
+        stack.push({ element: children[i], depth: entry.depth + 1 });
+      }
+    }
+
+    return { elementCount, totalDepth };
+  }
+
+  private _countImagesAboveFold(): number {
+    const viewportHeight = document.documentElement?.clientHeight ?? 0;
```
jpmunz: should this be consistent with what `_measure` grabs? What would be the reason for using `documentElement.clientWidth` vs. `document.scrollingElement?.scrollWidth`

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
jpmunz: I'd simplify the sending and just have the log sent on session ends, in `this._onLoadHandler` we can take the image above the fold measurement and just hold it until the normal send on session end

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
```diff
@@ -0,0 +1,188 @@
...
+    };
+
+    if (this._config.enabled) {
+      this.enable();
+    }
+  }
+
+  public override onEnable(): void {
+    // If the SDK is enabled after load has already fired, snapshot immediately;
+    // otherwise wait for the one load event. Either way, the load snapshot marks
```
jpmunz:
```suggestion
    // otherwise wait for the on load event. Either way, the load snapshot marks
```

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
```diff
@@ -0,0 +1,188 @@
...
+    });
+
+    this._onLoadHandler = (): void => {
+      // Capture the DOM at the load moment synchronously so the snapshot reflects
+      // the true at-load state, but defer the send: when instantiated as part of an
+      // isolated (registerGlobally: false) SDK instance, the logger provider is only
+      // wired onto this instrumentation after construction returns, so emitting here
+      // would go to a no-op logger. A microtask runs after that wiring completes in
+      // both registration modes, and the cold-start session part is still active by
+      // then, so log/part correlation is preserved.
```
jpmunz: not sure this is needed

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
overbalance: yes it should. layout is irrelevant to depth

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
overbalance: the fix, or the comment?

jpmunz: the comment, or could be simplified

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
overbalance: This came from lux. Root === 1

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
overbalance: yes, I just aligned it with the new measure util in the first PR of this stack

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
overbalance: Changed my mind, I'm adding `dom_state.phase: load` since the event is now deferred and will arrive in the same batch as the other measurement logs

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
```diff
@@ -0,0 +1,283 @@
...
+    }
+  }
+
+  private _hasLoadEventFired(): boolean {
+    const navigation = this.perf.getNavigationEntry();
+
+    // WebKit gives about:blank and srcdoc documents no navigation entry at all;
+    // readiness is the only signal left, pre-dispatch gap and all.
+    if (!navigation) {
+      return document.readyState === 'complete';
```
jpmunz, line 128: in DocumentLoadInstrumentation `document.readyState` is the only check done:
https://github.com/embrace-io/embrace-web-sdk/blob/d25ec02ccc7117c105fe5dc0d3230715ab7eb01f/packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts#L386

is the `navigation.loadEventStart` needed here? Or conversely should DocumentLoadInstrumentation also be checking that?

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
```diff
@@ -0,0 +1,283 @@
...
+      }
+
+      this._heldViewLog = {
+        attributes: {
+          [KEY_EMB_TYPE]: EMB_TYPES.OTelLog,
+          [KEY_EMB_SESSION_PART_ID]: sessionPartId,
+          [ATTR_DOM_STATE_PHASE]: phase,
+          [ATTR_DOM_STATE_VIEWPORT_HEIGHT]: measurement.viewportHeight,
+          [ATTR_DOM_STATE_VIEWPORT_WIDTH]: measurement.viewportWidth,
+          [ATTR_DOM_STATE_IMAGES_ABOVE_FOLD]:
```
jpmunz: I don't know that a full other log is needed here we could just grab the images above fold measurement and include that on the log sent on session part end, e.g. if we had `this._imagesAboveFold = null` on init and `this._imagesAboveFold = this._countImagesAboveFold(measurement)` here then in `_buildSessionPartEndAttributes`:

```
      ...(this._imagesAboveFold !== null
        ? {
            [ATTR_DOM_STATE_IMAGES_ABOVE_FOLD]: this._imagesAboveFold,
          }
        : {}),
...

     this._imagesAboveFold = null;
```

I think would then simplify the other logic needed for sending the 2nd log and wiring up the user session manager, etc.

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
overbalance: They serve two different purposes and I think each one earns its place. Here we want the DOM to be settled after JS has loaded and modified the page. In doc-load we want to check the entries quickly committed after DCL.

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
overbalance: I'm not sure I follow. When you say "full other log" do you mean the first log on load? 

The phase matters for fold image collection so moving the value to session part end would associate that measurement with an incorrect collection time, phase, and node count.

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
```diff
@@ -0,0 +1,268 @@
...
+          [ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_COUNT]:
+            this._countImagesAboveFold(measurement),
+          [ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_VIEWPORT_HEIGHT]:
+            measurement.viewportHeight,
+          [ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_VIEWPORT_WIDTH]:
+            measurement.viewportWidth,
+          [ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_TIMESTAMP]:
+            this.perf.getNowMillis(),
+        },
+        sessionPartId,
```
jpmunz: since we'll be sending this on the next session part end would it be possible for this to have a different part ID? If there isn't a way to get out of sync we could simplify by not needing `setUserSessionManager` etc.

jpmunz (approved):
lgtm, just the question about whether we need the session part id tracking

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
overbalance: They get out of sync when there are two instances on the page but I don't like having this kind of logic in every instrumentation. I'm going to remove it and the guard that goes with it for now, knowing that it doesnt do exactly what we want until we design a non-globally registered handler.

PR #1443, opened 2026-07-23
EMBR-13077 feat(web-vitals): capture LCP element type and bounding rect attribution

overbalance:

Part of the Web Vital Element Enrichment epic (EMBR-13077). LCP web-vital events carried no information about *which* element was the largest contentful paint. This adds the LCP element's type (backed by the new `LCPElementType` materialized column) and its bounding rectangle so element-level attribution is available in the backend.

- New `lcpElementAttribution` helper in `WebVitalsInstrumentation.ts`, mirroring the existing `ttfbSubPartsAttribution` / `loafScriptsAttribution` pattern. It reads `metric.attribution.lcpEntry?.element` once and, when present, emits:
  - `emb.web_vital.attribution.elementType` — the element's lowercased `tagName`.
  - `emb.web_vital.attribution.elementBoundingRect` — JSON-stringified `{x, y, width, height}` from `getBoundingClientRect()`, each `Math.round`ed. `x`/`y` are intentionally not clamped, since a scrolled-away element legitimately has a negative position.
- Wired into `_emitWebVital` conditioned on `metric.name === 'LCP'`.
- Both attributes are omitted (not empty/garbage) when the element is unavailable. The attribution read and geometry are wrapped in a `try/catch` that logs via `diag`, so a throwing/missing element never drops the whole LCP record.
- Integration harness fix: the vite-6/vite-7 test pages now render a fixed-size `<div>` instead of an `<h1>` as the LCP element. A text heading's box height drifts with each browser's default font (heights of 37/38/41/44 seen across chromium/firefox/webkit), which made the bounding-rect golden nondeterministic. A block with explicit `width`/`height` reports an identical rect in every browser.

## How was this tested?

New unit tests in `WebVitalsInstrumentation.test.ts`:
- Element present → both attributes set (tag lowercased, rounded rect JSON including a negative `x`).
- Element absent → both attributes omitted.
- `getBoundingClientRect()` throws → LCP record still emitted, both attributes omitted, error logged via `diag`.

Full `@embrace-io/web-sdk` unit suite passes; `npm run lint` and `npm run check` are green. Integration goldens regenerated and now match across all three browser engines.

## Checklist

- [x] Tests added/updated
- [x] Lint and typecheck pass
- [x] Golden files regenerated

### DIFF
--- a/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
@@ -206,10 +206,10 @@ const lcpElementAttribution = (
     if (element) {
       const prefix = KEY_EMB_WEB_VITAL_ATTRIBUTION_PREFIX;
       const rect = element.getBoundingClientRect();
-      attributes[`${prefix}LCPElementType`] = element.tagName.toLowerCase();
+      attributes[`${prefix}lcpElementType`] = element.tagName.toLowerCase();
       // x/y are intentionally not clamped: an element scrolled above the
       // viewport legitimately has a negative position.
-      attributes[`${prefix}LCPElementBoundingRect`] = JSON.stringify({
+      attributes[`${prefix}lcpElementBoundingRect`] = JSON.stringify({
         x: Math.round(rect.x),
         y: Math.round(rect.y),
         width: Math.round(rect.width),
--- a/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
@@ -6,6 +6,7 @@ import { safeExecuteInTheMiddle } from '@opentelemetry/instrumentation';
 import type {
   CLSMetricWithAttribution,
   INPAttribution,
+  LCPAttribution,
   Metric,
   MetricWithAttribution,
   TTFBAttribution,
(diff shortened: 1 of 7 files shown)


packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -192,6 +193,36 @@ const ttfbSubPartsAttribution = (
...
+  const attributes: Attributes = {};
+
+  try {
+    const attribution = metric.attribution as LCPAttribution;
+    const element = attribution.lcpEntry?.element;
+
+    if (element) {
+      const prefix = KEY_EMB_WEB_VITAL_ATTRIBUTION_PREFIX;
+      const rect = element.getBoundingClientRect();
+      attributes[`${prefix}lcpElementType`] = element.tagName.toLowerCase();
```
joaquin-diaz: I think this is probably my bad when I created the tickets but the backend already prefixes the web vital attributes based on their name. So the attribute name should be either `elementType` or `element_type`. I remember us talking about all being snake case at some point?

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -192,6 +193,36 @@ const ttfbSubPartsAttribution = (
...
+      // x/y are intentionally not clamped: an element scrolled above the
+      // viewport legitimately has a negative position.
+      attributes[`${prefix}lcpElementBoundingRect`] = JSON.stringify({
```
joaquin-diaz: Same here

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
joaquin-diaz: I updated the tickets, sorry about that!

overbalance: You're right there are currently a mix of cases. I have been treating attribution like `http.headers.*` which leaves some flexibility but was going to make all information consistent as a new project

jpmunz (approved):

joaquin-diaz (approved):

PR #1444, opened 2026-07-23
EMBR-13089 feat(web-vitals): capture CLS layout shifts attribution

overbalance:

The SDK reports the CLS web vital value but not which elements shifted, so CLS regressions can't be traced back to the responsible DOM nodes. This adds per-shift attribution to the CLS log record.

- New `clsLayoutShiftsAttribution`, emitted only for CLS: serializes each layout-shift entry to `emb.web_vital.attribution.clsLayoutShifts` as `{ selector, startRect, endRect }`, taking the first (most impactful) source per shift and reusing web-vitals' `getSelector` for selector parity.
- Caps output at `MAX_CLS_LAYOUT_SHIFTS` (50) and reports `emb.web_vital.attribution.clsLayoutShiftsDroppedCount` when the cap is exceeded.
- Rects rounded to integer CSS pixels; best-effort with any errors logged via `diag`.
- Extracted a shared module-level `roundRect` helper, reused by both the CLS shift rects and the LCP element bounding rect (dropping the duplicated inline rounding on each path).

## How has this been tested?

Unit tests in `WebVitalsInstrumentation.test.ts` cover under-cap serialization with rounding, the 50-cap plus dropped count, the empty-sources fallback, first-source selection on multi-source shifts, and absence of the attributes on non-CLS metrics. `npm run check` and the web-vitals test file pass.

## Checklist

- [ ] Documentation added
- [x] Tests added

### DIFF
--- a/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
@@ -278,11 +278,14 @@ const lcpElementAttribution = (
     const element = attribution.lcpEntry?.element;
 
     if (element) {
-      const prefix = KEY_EMB_WEB_VITAL_ATTRIBUTION_PREFIX;
-      attributes[`${prefix}elementType`] = element.tagName.toLowerCase();
-      attributes[`${prefix}elementBoundingRect`] = JSON.stringify(
+      const elementType = element.tagName.toLowerCase();
+      const elementBoundingRect = JSON.stringify(
         roundRect(element.getBoundingClientRect()),
       );
+
+      const prefix = KEY_EMB_WEB_VITAL_ATTRIBUTION_PREFIX;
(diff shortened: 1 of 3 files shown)


packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -131,6 +133,50 @@ const loafScriptsAttribution = (
...
+const clsLayoutShiftsAttribution = (
+  metric: MetricWithAttribution,
+  diag: DiagLogger,
+): Attributes => {
+  const attributes: Attributes = {};
+
+  try {
+    /* eslint-disable baseline-js/use-baseline */
+    const entries = metric.entries as LayoutShift[];
+    const roundRect = (rect?: DOMRectReadOnly) => ({
```
jpmunz: should move this outside so it doesn't need to be redefined each call, could re-use it for LCP bounding rect as well

jpmunz (approved):

PR #1445, opened 2026-07-24
EMBR-13078 feat(web-vitals): capture INP element_type attribution attribute

joaquin-diaz:

`INPAttribution` only exposes the interaction target as a CSS selector string (`interactionTarget`), not as a DOM element type. There's no way to know what kind of element (button, a, div, etc.) triggered a slow INP interaction, which limits how actionable INP data is for debugging.

- Add `inpAttribution` helper in `WebVitalsInstrumentation.ts` that finds the first INP `PerformanceEventTiming` entry with a resolvable `target` (mirroring the same null-guard `.find` that `web-vitals` itself uses internally for `interactionTarget`), and emits `emb.web_vital.attribution.element_type` with the lowercased tag name.
- Attribute is omitted when the target is missing/removed from the DOM by report time.
- Wired into `_emitWebVital`'s attributes for INP metrics only.

## How has this been tested?

Added unit tests in `WebVitalsInstrumentation.test.ts` covering:
- a single entry with a resolvable target
- multiple entries where only one has a resolvable target (null-guard case)
- no entry has a target (attribute omitted)

`npm run check` (tsc + eslint) and the full `WebVitalsInstrumentation` suite pass.

## Checklist

- [ ] Documentation added
- [x] Tests added

### DIFF
--- a/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
@@ -147,7 +147,7 @@ const inpAttribution = (
     const tagName = target?.tagName?.toLowerCase();
 
     if (tagName) {
-      attributes[`${KEY_EMB_WEB_VITAL_ATTRIBUTION_PREFIX}INPElementType`] =
+      attributes[`${KEY_EMB_WEB_VITAL_ATTRIBUTION_PREFIX}element_type`] =
         tagName;
     }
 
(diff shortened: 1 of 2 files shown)


packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -131,6 +131,33 @@ const loafScriptsAttribution = (
...
+    )?.target as Element | null | undefined;
+
+    const tagName = target?.tagName?.toLowerCase();
+
+    if (tagName) {
+      attributes[`${KEY_EMB_WEB_VITAL_ATTRIBUTION_PREFIX}element_type`] =
+        tagName;
+    }
+
+    /* eslint-enable baseline-js/use-baseline */
```
jpmunz, line 154: this is only needed for line 142 right?

jpmunz (approved):

packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
joaquin-diaz: Technically yes but keeping it consistent with other blocks, If we want to do more with the entry we'll probably have to use it as well

overbalance (approved):

PR #1450, opened 2026-07-27
feat(utils): add measureDocument util for document and viewport size

overbalance:
Bottom of stack #1453.

An instrumentation that needs the content size of the document or the size of the viewport has no shared way to read them. A direct read of `documentElement` or `<body>` is correct in one rendering mode only. A pair of `documentElement.scrollHeight` and `window.innerHeight` has a second problem. It mixes a scrollbar-inclusive box with a scrollbar-exclusive one.

- Add `measureDocument()` in `packages/web-sdk/src/utils/`. It reads the document size and the viewport size from `document.scrollingElement`, the one element that is correct in both rendering modes
- Return `null` when there is no scroll root, and when the scroll root has no layout box. A caller then omits the attribute instead of emitting a 0 that a consumer would read as a real measurement
- Export `measureDocument` and `DocumentMeasurement` from the utils barrel

Both boxes come back from one call so they share a layout state. The document size and the viewport size only mean something against each other, and the scrollable range is their difference.

This PR adds no consumers. #1452 moves `MaxScrollDepthInstrumentation` onto `measureDocument()`. #1442 uses it for the DOM state snapshot.

## How has this been tested?

Five unit tests cover the util. The first test reads every dimension from a stubbed scroll root. Each stub value is distinct, so the test pins the mapping. Two tests cover the two `null` paths. One test measures the live document. The last test checks the quirks-mode condition in an iframe without a doctype, so the test proves the condition instead of asserting it in prose.

The live-document test appends a spacer three viewports tall. The bare test document measures equal in pairs (600/600, 800/800), so a `> 0` assertion would pass even with the document size and the viewport size swapped.

A mutation test covers each guard: a broken implementation makes the tests fail.

`npm run test`, `npm run lint`, and `npm run check` pass.

## Checklist

- [ ] Documentation added
- [x] Tests added

### DIFF
--- a/packages/web-sdk/src/utils/measureDocument.ts
+++ b/packages/web-sdk/src/utils/measureDocument.ts
@@ -1,0 +1,50 @@
+/**
+ * A point-in-time reading of the document and viewport boxes, in integer CSS
+ * pixels. Resize, zoom, and content mutation all invalidate it.
+ *
+ * `viewportHeight`/`viewportWidth` are the scroll root's client box, which
+ * excludes any space a classic scrollbar takes, so they are `<=`
+ * `window.innerHeight`/`innerWidth` rather than equal to them. The document
+ * boxes are always `>=` their viewport counterparts, and the scrollable range
+ * is the difference between the two.
+ */
+export interface DocumentMeasurement {
+  documentHeight: number;
+  documentWidth: number;
+  viewportHeight: number;
+  viewportWidth: number;
+}
+
+/**
+ * Measures the document's content size and the viewport size off
+ * `document.scrollingElement`, which is the document root in standards mode,
+ * and in quirks mode is `<body>` only when `<body>` is not itself potentially
+ * scrollable. Reading `documentElement` or `<body>` directly is correct in only
+ * one of the two modes.
+ *
+ * Returns `null` in the remaining quirks-mode cases, and when the scroll root
+ * has no layout box. The document may still scroll in the quirks-mode case, but
+ * no element measures it correctly, so callers should report the absence rather
+ * than substitute a zero that would read as a real measurement.
+ * https://developer.mozilla.org/en-US/docs/Web/API/Document/scrollingElement
+ */
+export const measureDocument = (): DocumentMeasurement | null => {
+  const scrollRoot = document.scrollingElement;
+  if (!scrollRoot) {
+    return null;
+  }
+
+  const viewportHeight = scrollRoot.clientHeight;
+  const viewportWidth = scrollRoot.clientWidth;
+  // A non-rendered or zero-sized frame has a scroll root with no layout box.
+  if (!viewportHeight || !viewportWidth) {
+    return null;
+  }
+
+  return {
+    documentHeight: scrollRoot.scrollHeight,
+    documentWidth: scrollRoot.scrollWidth,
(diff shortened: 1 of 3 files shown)


packages/web-sdk/src/utils/measureDocument.ts
```diff
@@ -0,0 +1,50 @@
...
+  viewportHeight: number;
+  viewportWidth: number;
+}
+
+/**
+ * Measures the document's content size and the viewport size off
+ * `document.scrollingElement`, which is the document root in standards mode,
+ * and in quirks mode is `<body>` only when `<body>` is not itself potentially
+ * scrollable. Reading `documentElement` or `<body>` directly is correct in only
+ * one of the two modes.
```
jpmunz:
```suggestion
 * scrollable. Reading `documentElement` or `<body>` directly is correct in only
 * standards mode.
```

packages/web-sdk/src/utils/measureDocument.ts
```diff
@@ -0,0 +1,50 @@
...
+ * than substitute a zero that would read as a real measurement.
+ * https://developer.mozilla.org/en-US/docs/Web/API/Document/scrollingElement
+ */
+export const measureDocument = (): DocumentMeasurement | null => {
+  const scrollRoot = document.scrollingElement;
+  if (!scrollRoot) {
+    return null;
+  }
+
+  const viewportHeight = scrollRoot.clientHeight;
```
jpmunz, line 40: previously in packages/web-sdk/src/instrumentations/max-scroll-depth/MaxScrollDepthInstrumentation/MaxScrollDepthInstrumentation.ts `window.innerHeight` was used here, is correct to switch to `clientHeight`?

packages/web-sdk/src/utils/measureDocument.ts
```diff
@@ -0,0 +1,50 @@
...
+  const viewportWidth = scrollRoot.clientWidth;
+  // A non-rendered or zero-sized frame has a scroll root with no layout box.
+  if (!viewportHeight || !viewportWidth) {
+    return null;
+  }
+
+  return {
+    documentHeight: scrollRoot.scrollHeight,
```
jpmunz, line 48: either here or in the comment documenting the function I think we should preserve the forced reflow comment from packages/web-sdk/src/instrumentations/max-scroll-depth/MaxScrollDepthInstrumentation/MaxScrollDepthInstrumentation.ts to warn callers against hitting this too often

packages/web-sdk/src/utils/measureDocument.ts
overbalance: This is a clunky sentence, it's saying one read is for standards mode and one read is for quirks mode. I'll rewrite.

packages/web-sdk/src/utils/measureDocument.ts
overbalance: Yes. innerheight includes the horizontal scrollbar height and scrollY uses the height excluding the scrollbar so to do a proper percentage you need to use clientHeight. On "invisible scrollbars" like touchpad on mac, they are the same value, but on other platforms the 15 or 20 pixel difference could lead to negative values which isn't right

jpmunz (approved):

PR #1452, opened 2026-07-27
EMBR-13293 fix(max-scroll-depth): measure via measureDocument and omit box-derived keys

overbalance:
Middle of stack #1453. Base is #1450.

`MaxScrollDepthInstrumentation` measured the scroll root by itself. It paired `documentElement.scrollHeight` with `window.innerHeight`. That pair is correct in standards mode only. It also mixes a scrollbar-exclusive box with a scrollbar-inclusive one, so the subtraction did not describe a real range.

`window.scrollY` stops at `scrollHeight - clientHeight`. `window.innerHeight` is never smaller than that client height, because it includes the horizontal scrollbar. The denominator was therefore narrower than the real scrollable range, and `max_scroll_depth.percent` came out too high. It reached 100 before the user reached the bottom. When the whole scrollable range was shorter than the scrollbar, the subtraction went negative and the instrumentation reported 0 for a page that does scroll.

When the page rendered nothing, the instrumentation emitted `percent: 0` and `document_height: 0`. A consumer reads those values as a real measurement at the top of a zero-height page. The correct meaning is no measurement.

The instrumentation also missed restored scroll offsets. The maximum came from scroll events only. A reload or a back navigation restores a deep offset, and the instrumentation then reported a depth of 0.

- Measure through `measureDocument()`. Do not read `documentElement` and `window` by hand. Both boxes now come from `document.scrollingElement`, so both are correct in both rendering modes and both exclude the scrollbar
- Omit `max_scroll_depth.percent` and `max_scroll_depth.document_height` when the scroll root is unmeasurable. A consumer then reads absence instead of a false 0. `pixels` and `did_scroll` come from the scroll position alone, so the instrumentation still emits them
- Add the current scroll position to the maximum at emit time. A document that loads at a restored offset has already reached that depth. `did_scroll` stays false in this case, which separates it from a real scroll
- Extract `_scrollPercent`. This method returns 0 for a zero scrollable range. It also clamps elastic overscroll. The range can no longer go negative, because both boxes come off the same element

Data continuity: `percent` now divides by the client height of the scroll root instead of `window.innerHeight`. The two agree unless the page renders a classic, non-overlay horizontal scrollbar. macOS and iOS use overlay scrollbars, so values there do not move. Where a classic scrollbar is present, reported percentages drop by up to the height of that scrollbar as a share of the scrollable range, because the earlier values were too high. `document_height` does not change in standards mode.

## How has this been tested?

The existing unit tests now use the scroll-root geometry. Each test stubs a detached element as `scrollingElement`, so the stubbed dimensions cannot match the real scroll root. Three tests are new. Two tests cover the unmeasurable paths: no scrolling element, and no layout box. Both assert that the two keys are absent. The third test restores an offset and asserts a depth with `did_scroll: false`.

All 9 unit tests pass. `npm run lint` and `npm run check` pass from the repo root.

The change does not affect the golden files. The test document measures 720 high in a 720 viewport. The scrollable range is therefore 0, and `percent` stays 0.

## Checklist

- [ ] Documentation added
- [x] Tests added

### DIFF
--- a/packages/web-sdk/src/instrumentations/max-scroll-depth/MaxScrollDepthInstrumentation/MaxScrollDepthInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/max-scroll-depth/MaxScrollDepthInstrumentation/MaxScrollDepthInstrumentation.ts
@@ -1,5 +1,7 @@
 import { SeverityNumber } from '@opentelemetry/api-logs';
 import { EMB_TYPES, KEY_EMB_TYPE } from '../../../constants/attributes.ts';
+import type { DocumentMeasurement } from '../../../utils/index.ts';
+import { measureDocument } from '../../../utils/index.ts';
 import { EmbraceInstrumentationBase } from '../../EmbraceInstrumentationBase/index.ts';
 import {
   ATTR_MAX_SCROLL_DEPTH_DID_SCROLL,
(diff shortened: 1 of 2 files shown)


packages/web-sdk/src/instrumentations/max-scroll-depth/MaxScrollDepthInstrumentation/MaxScrollDepthInstrumentation.ts
```diff
@@ -63,33 +65,60 @@ export class MaxScrollDepthInstrumentation extends EmbraceInstrumentationBase {
...
   }
+
+  private _scrollPercent({
+    documentHeight,
+    viewportHeight,
+  }: DocumentMeasurement): number {
+    // Both boxes are read off the same scroll root, so the document is never
+    // shorter than the viewport and the range is never negative.
+    const scrollBottom = documentHeight - viewportHeight;
+    if (scrollBottom === 0) {
```
jpmunz: rather than documenting the assumption it seems we could just enforce that in the check:

```suggestion
    const scrollBottom = documentHeight - viewportHeight;
    if (scrollBottom <= 0) {
```

jpmunz (approved):

packages/web-sdk/src/instrumentations/max-scroll-depth/MaxScrollDepthInstrumentation/MaxScrollDepthInstrumentation.ts
overbalance: I had that initially but thought it was confusing because it suggests that less than zero is possible when it's not.

EDIT: on second thought, the comment is temporal so I'm just going to drop it. 0 is what most readers would expect anyway

PR #1455, opened 2026-07-30
EMBR-13084 refactor(utils): measure the viewport off the window when there is no scroll root

overbalance:

`measureDocument` read both the document and the viewport off `document.scrollingElement`, which is `null` in quirks mode when `<body>` is potentially scrollable, so `MaxScrollDepthInstrumentation` dropped `max_scroll_depth.percent` and `max_scroll_depth.document_height` for those documents even though the depth was measurable. Extracted from #1442, which rebases on top as the util's second consumer.

Three commits, one concern each:

1. **fix(utils)**: measure the document off the scroll root, falling back to `<html>` for the document boxes and to the window for the viewport where there is no scroll root. `null` now means only "nothing to measure" (no browsing context, no root element, nothing rendered, or a zero box), never a measurable quirks document. The fallback range is approximate (quirks `<html>` undercounts; `innerHeight` includes scrollbar space): approximate-but-present beats omission. Tests rebuilt around real documents in iframes, since doctype-dependent behavior cannot be stubbed.
2. **refactor(max-scroll-depth)**: take `scrollableHeight` (floored at zero) from the measurement instead of subtracting boxes at the call site, where the subtraction can go negative.
3. **fix(max-scroll-depth)**: clamp the scroll position carried into the next session part; Safari reports it negative during rubber-band overscroll past the top.

## How has this been tested?

`npm run lint`, `npm run check`, and the web-sdk unit suite pass; both affected test files pass under `npm run test:multiBrowsers` (Chromium, Firefox, WebKit, Pixel 5 and iPhone 12 emulations).

### DIFF
--- a/packages/web-sdk/src/utils/measureDocument.ts
+++ b/packages/web-sdk/src/utils/measureDocument.ts
@@ -2,51 +2,69 @@
  * A point-in-time reading of the document and viewport boxes, in integer CSS
  * pixels. Resize, zoom, and content mutation all invalidate it.
  *
- * `viewportHeight`/`viewportWidth` are the scroll root's client box, which
- * excludes any space a classic scrollbar takes, so they are `<=`
- * `window.innerHeight`/`innerWidth` rather than equal to them. The document
- * boxes are always `>=` their viewport counterparts, and the scrollable range
- * is the difference between the two.
+ * `viewportHeight`/`viewportWidth` are the box the document scrolls inside,
+ * read from the scroll root's client box, or from the window where there is
+ * no scroll root.
+ *
+ * `scrollableHeight` is the scrollable range the measurement supports, floored
+ * at zero: exact off a scroll root, approximate where the viewport came from
+ * the window. Take it rather than subtracting the boxes, which can go negative.
  */
 export interface DocumentMeasurement {
   documentHeight: number;
   documentWidth: number;
+  scrollableHeight: number;
   viewportHeight: number;
   viewportWidth: number;
 }
 
 /**
- * Measures the document's content size and the viewport size off
- * `document.scrollingElement`, which is the document root in standards mode,
- * and in quirks mode is `<body>` only when `<body>` is not itself potentially
- * scrollable. `documentElement` is correct in standards mode only and `<body>`
- * in quirks mode only, so neither is safe to read directly.
- *
- * Returns `null` in the remaining quirks-mode cases, and when the scroll root
- * has no layout box. The document may still scroll in the quirks-mode case, but
- * no element measures it correctly, so callers should report the absence rather
- * than substitute a zero that would read as a real measurement.
+ * Measures the document off `document.scrollingElement`, the only element whose
+ * scroll box spans the viewport's scrolling area. In quirks mode `<html>` reports
+ * just its own, dropping its margin, border, and out-of-flow content, so it
+ * stands in only where there is no scroll root. The viewport cannot come from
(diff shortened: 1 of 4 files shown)


packages/web-sdk/src/utils/measureDocument.ts
```diff
@@ -1,53 +1,96 @@
...
+  // A non-null scroll root is the root element in standards mode or <body> in
+  // quirks mode, and both getters return the viewport for either of those, so the
+  // two sides cannot disagree. A zero on one means the scroll root cannot report
+  // the viewport, and the window box replaces the pair.
+  // https://drafts.csswg.org/cssom-view/#dom-element-clientwidth
+  if (scrollRoot?.clientHeight && scrollRoot.clientWidth) {
+    return measurement(scrollRoot.clientHeight, scrollRoot.clientWidth);
+  }
+
+  const viewportHeight = view.innerHeight;
```
overbalance: The added fallback

packages/web-sdk/src/utils/measureDocument.ts
```diff
@@ -1,53 +1,96 @@
...
 export interface DocumentMeasurement {
-  documentHeight: number;
-  documentWidth: number;
-  viewportHeight: number;
-  viewportWidth: number;
+  readonly documentHeight: number;
+  readonly documentWidth: number;
+  readonly viewportHeight: number;
+  readonly viewportWidth: number;
+  readonly scrollableHeight: number;
```
overbalance: Added the calculated value

packages/web-sdk/src/utils/measureDocument.ts
```diff
@@ -1,53 +1,96 @@
...
+  // between, so one forced layout serves them all and the values describe a
+  // single layout state.
   // https://developer.chrome.com/docs/performance/insights/forced-reflow
-  const viewportHeight = scrollRoot.clientHeight;
-  const viewportWidth = scrollRoot.clientWidth;
-  // A non-rendered or zero-sized frame has a scroll root with no layout box.
+  const documentHeight = doc.documentElement.scrollHeight;
+  const documentWidth = doc.documentElement.scrollWidth;
+
+  const measurement = (
```
jpmunz: seems odd to have an inline function here, I'd re-order things a bit to have viewportHeight/viewportWidth set based on what's available and then return the object with the final measurement directly at the end

packages/web-sdk/src/utils/measureDocument.ts
```diff
@@ -1,53 +1,96 @@
...
+    documentWidth,
+    viewportHeight,
+    viewportWidth,
+    // The document is measured off <html> and the viewport off <body> or the
+    // window. In standards mode the spec floors <html>'s scrollHeight at the
+    // viewport height, so the difference cannot go negative. Quirks mode is not
+    // floored, and there Chromium and Firefox report a short document's own
+    // height while WebKit floors it at the viewport, so the subtraction goes
+    // negative on two engines of the three.
+    // https://drafts.csswg.org/cssom-view/#dom-element-scrollheight
```
jpmunz: is this reference correct? That's referring to an element's scrollHeight property but we're computing something different in `scrollableHeight` right?

overbalance: The reference was to highlight the discrepancy in implementation but it's overkill to include. I'l. drop it.

jpmunz (approved):

PR #1456, opened 2026-08-02
docs(claude): trim and correct CLAUDE.md, extract SDK skills

overbalance:

`.claude/CLAUDE.md` restated what `package.json` and the source tree already show, and several of the claims it did make were wrong.

- Drop the command listings, source-layout tree, and distribution table. The build targets are declared in `tsdown.config.ts` and enforced by `scripts/validate.ts`.
- Extract the "Common Tasks" walkthroughs, plus golden-file and browser-detection guidance, into three skills.
- Correct seven wrong claims, notably a nonexistent `emb.page.*` attribute prefix (the real keys are `app.surface.*`), a backwards soft-navigation support claim, and two "build first" steps turbo already covers via `dependsOn`.
- Harden the golden-file diff script, which reported "nothing changed" on an unresolvable git ref, an added golden, or a deleted one.

The skills document the container regeneration path, which is broken on `main` and fixed in the stacked PR.

## How has this been tested?

Docs only. Every script name, path, port and attribute key checked against the repo; the diff script run across six cases including its failure paths.

## Checklist

- [x] Documentation added
- [ ] Tests added


.claude/skills/adding-sdk-components/SKILL.md
```diff
@@ -0,0 +1,30 @@
...
+# Adding SDK Components
+
+Both walkthroughs below have registration steps that are easy to miss: a component that is created but not exported and wired will silently never run.
+
+## Adding an Instrumentation
+
+1. Create in `packages/web-sdk/src/instrumentations/<name>/`
+2. Extend `EmbraceInstrumentationBase`
+3. Export from `packages/web-sdk/src/instrumentations/index.ts`
+4. Add its name to the `OptionalInstrumentations` union in `packages/web-sdk/src/sdk/types.ts`, and give it a key in `DefaultInstrumentationConfig` in the same file
```
jpmunz:
```suggestion
4. If optional add its name to the `OptionalInstrumentations` union in `packages/web-sdk/src/sdk/types.ts`, and give it a key in `DefaultInstrumentationConfig` in the same file
```

.claude/skills/regenerating-golden-files/SKILL.md
```diff
@@ -0,0 +1,100 @@
...
+
+   Confirm every added and removed key is one you meant to change, and that nothing was removed by accident.
+
+## Diffing attribute keys
+
+Run this from the repo root. It compares the working tree against `origin/main` and prints the attribute keys that appeared and disappeared per file.
+
+It always prints a trailing summary line, on purpose. Every check below exits nonzero with a message rather than printing nothing, because the failure that matters here is the one that looks like success: a comparison that comes back empty reads as "no attributes changed" whether that is true or the comparison never happened.
+
+```bash
```
jpmunz, line 38: can move this to its own python file under regenerating-golden-files/scripts/

PR #1461, opened 2026-08-03
EMBR-13419 refactor(sessions): stop setting session.id and session.previous_id

overbalance:

The SDK was stamping the OTel semantic-convention keys `session.id` and `session.previous_id` on spans and logs with its own user session ids. That key space belongs to the customer: anyone carrying their own notion of a session there had their values overwritten, and there was no way to tell the two apart downstream. Embrace already publishes its ids on `emb.user_session_id`, `emb.user_session_previous_id`, and `emb.session_part_id`, so the convention keys were redundant as well as destructive.

Matches the Android change in https://github.com/embrace-io/embrace-android-sdk/pull/3721.

- `EmbraceUserSessionManager` and `UserSessionLogRecordProcessor` no longer write `session.id` / `session.previous_id`; customer-set values on those keys now pass through untouched.
- Removed the plumbing that carried those keys: manager `types.ts`, exporter and `initSDK` test expectations, mock `ReadableSpan`, `server/utils.ts`, and the `IGNORED_ATTRIBUTES_LIST` entries.
- Regenerated the integration golden files, which drop the two attributes.
- Documented the ownership split in the manager README (full stamped-attribute table plus an explicit "never stamped" section), and corrected the integration README's description of golden comparison: only attribute *values* are ignored, key presence and count are still compared.

## How has this been tested?

- `npm run test` from `packages/web-sdk/`: 1037 passed, including a new `UserSessionLogRecordProcessor` case asserting customer-set `session.id` / `session.previous_id` survive while the `emb.*` ids are stamped alongside.
- `npm run test:integration` against the regenerated goldens: 300 passed.
- `npm run lint` and `npm run check` from the repo root, both clean.

## Checklist

- [x] Documentation added
- [x] Tests added


fractalwrench (approved):
LGTM

jpmunz (approved):

PR #1462, opened 2026-08-03
EMBR-13422 fix(web-vitals): attribute browser.url.full and page keys to one page

overbalance:

A single web vital log record could describe two different pages at once. `browser.url.full` was sourced from web-vitals' `navigationURL` when soft navigations were active, while `emb.page.path`, `emb.page.id` and `app.surface.label` always came from the page attributed when the metric was reported. When those disagreed, one record claimed the vital happened at one URL and identified a different page.

The case that triggers it, an SPA that redirects during bootstrap:

- Hard load of `https://app.com/`, so the navigation entry name is `https://app.com/`
- The app calls `pushState` to `/dashboard`. No user interaction preceded it, so this is not a soft navigation and web-vitals keeps reporting against the hard navigation
- The LCP element, the dashboard hero, renders while `document.URL` is `https://app.com/dashboard`

That emitted `browser.url.full = https://app.com/` next to `emb.page.path = /dashboard`. Grouping by URL and grouping by page gave different answers for the same record.

- `browser.url.full`, `emb.page.path`, `emb.page.id` and `app.surface.label` now all come from the attributed page. Four attributes, one page
- New `browser.web_vital.navigation_url` carries web-vitals' `navigationURL`, next to the existing `browser.web_vital.navigation_id` and `navigation_type`. Nothing is lost: the navigation a metric value is anchored to is still queryable, just no longer conflated with the page the content rendered on
- Restructured the soft-navigation URL tests around the consistency guarantee instead of the old per-branch URL assertions. The `PerformanceObserver.supportedEntryTypes` stubbing these rely on landed in #1460
- Added a headed integration test for the bootstrap-redirect case, driven by a new `?bootstrapRedirect=` flag on the vite test harness. This is the shape that shipped broken, and nothing in the suite covered it: the e2e matrix runs with soft navigations active but had no case where the navigation URL and the rendered page disagree

`_softNavsActive` still gates `reportSoftNavs` on listener registration, so it is not dead.

Alternative considered and rejected: resolve page identity *for* the navigation URL, so all four attributes describe the navigation rather than the report. At hard navigation start the SPA router usually has not resolved a route yet, so that yields a URL with no path or label.

## How has this been tested?

- Unit suite: 1037 passed, 0 failed
- Verified every new test is not vacuous by temporarily restoring the old URL selection: three unit tests fail, and so does the new integration test, which reports `browser.url.full = /?bootstrapRedirect=lcp` for content that rendered at `/lcp`. That is the production bug, reproduced end to end. This matters here specifically, because the behavior being fixed was itself protected by a test that passed only because its branch was unreachable
- Unit coverage now asserts all four page attributes together rather than two of them, and covers CLS, whose attributed page is re-captured when the largest shift moves
- Regenerated the 9 integration golden files that carry web vitals. The only semantic change is the added `browser.web_vital.navigation_url`; `browser.url.full` is unchanged in every record, so this is behavior-neutral for pages that do not redirect during bootstrap
- Integration suite: 300 passed, 72 skipped
- `npm run lint` and `npm run check` from the repo root: clean

## Checklist

- [ ] Documentation added
- [x] Tests added

Docs are left unticked deliberately: `browser.web_vital.navigation_url` still needs a home in the attribute reference, and the ticket tracks confirming the name and backend ingestion before that is written.

### DIFF
--- a/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
@@ -33,6 +33,7 @@ import {
   KEY_BROWSER_WEB_VITAL_NAME,
   KEY_BROWSER_WEB_VITAL_NAVIGATION_ID,
   KEY_BROWSER_WEB_VITAL_NAVIGATION_TYPE,
+  KEY_BROWSER_WEB_VITAL_NAVIGATION_URL,
   KEY_BROWSER_WEB_VITAL_RATING,
   KEY_BROWSER_WEB_VITAL_VALUE,
   KEY_EMB_WEB_VITAL_ATTRIBUTION_PREFIX,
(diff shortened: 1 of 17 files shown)


packages/web-sdk/src/instrumentations/web-vitals/WebVitalsInstrumentation/WebVitalsInstrumentation.ts
```diff
@@ -522,15 +528,9 @@ export class WebVitalsInstrumentation extends EmbraceInstrumentationBase {
...
-              // are not supported, we fall back to the attributed page.
-              [KEY_BROWSER_URL_FULL]:
-                metric.navigationURL != null &&
-                (metric.navigationType === 'soft-navigation' ||
-                  this._softNavsActive)
-                  ? metric.navigationURL
-                  : attributedPage.fullURL,
+              // The URL and the page keys must come from one source, or a record
+              // claims one URL while identifying a different page.
+              [KEY_BROWSER_URL_FULL]: attributedPage.fullURL,
```
jpmunz, line 533: wouldn't this still cause the inconsistency from the PR description since we're grabbing `attributedPage.fullURL` from the document's url but `attributePage.path` from the PageManager's current route which I believe is only updated when a soft navigation is detected? So in the example of a pushState of /dashboard with no soft navigation having `path` be "/" but `fullURL` be "/dashboard"?

PR #1466, opened 2026-08-04
feat(logs): add log.flush() and use it to drain e2e telemetry deterministically

overbalance:

`test-integration` has been red on main since #1460. Six tests fail with `expect(requests).toHaveLength(1)` receiving 2, on Next 16 Pages across all three browsers.

The SDK is fine. The tests assumed a timing coincidence:

- `BatchLogRecordProcessor` is built with no options (`initSDK.ts:253`), so it takes OTel's default `scheduledDelayMillis` of **1000ms**.
- The `requests` fixture accumulates from page load and is never reset.
- So `toHaveLength(1)` only held while the click's log landed in the *same 1s batch window* as the page-load web vitals.

Measured locally: page-load FCP recorded at `...412.9358`, the batch flushes ~1s later, and the click's `first-interaction` lands at `...414.3994`. A **1.46s** gap, past the boundary, so page-load telemetry ships as its own request and the count becomes 2.

Chromium 149 → 151, Firefox → 153 and WebKit → 26.5 shifted page-load timing just far enough to cross that line. Same class of latent fragility as the soft-navigation test in #1462: a test whose outcome silently depended on browser timing.

- New public `log.flush()` on `LogManager`, which exports logs still held in the batch buffer instead of waiting for the next scheduled export. It resolves rather than rejecting on failure, because the safe-proxy wrapper around the public API only traps synchronous throws.
- The fixture reaches it via `window.EMBRACE_SDK.log.flush()`. That global comes from #1472 below this one in the stack, so no harness change is needed here.
- New `settleAndResetRequests` fixture awaits that flush, then clears the buffer. Once flush resolves the buffer is empty as a fact, so no timing assumption remains.
- Called after `navigateAndWaitUntilReady` in the two affected tests, so each assertion counts only the traffic its own interaction caused.
- Regenerated 9 vite-7 goldens. These previously encoded the race outcome, capturing page-load telemetry that happened to share a batch with the click.

Plus the two tests that keep this honest, since nothing pinned either behaviour before:

- `logExportSchedule.test.ts` pins the 1000ms window with a fake clock: nothing exported at 999ms, exactly one record at 1001ms. It lives in its own file because @web/test-runner isolates each file in a fresh page, and sharing one with the other `initSDK` tests lets an SDK built earlier keep timers registered that the fake clock then fires.
- A new e2e test covers what `settleAndResetRequests` deliberately hides from the other tests: that real telemetry still ships separately when an interaction lands after the batch window. Waiting for the page-load flush *before* clicking makes that ordering certain rather than raced, so it is deterministic despite asserting timing-dependent behaviour.

Rejected alternatives: raising `scheduledDelayMillis` only widens the window, so the next browser bump reopens it, and there is no config knob for it today. A quiet-period wait (the first commit here) infers the flush from an absence of requests, which is a heuristic rather than a guarantee.

### Why `log.flush()` rather than the existing `SDKControl.flush()`

`SDKControl.flush()` does drain logs, but it awaits `tracerProvider.forceFlush()` as well, so it drains more than the fixture asks for. Using it here changed a request count and cost one integration test. The fixture wants logs only. It is also marked `@deprecated`, so new test infrastructure should not depend on it. It is left untouched.

### Why there is no matching `trace.flush()`

Spans cannot be flushed on demand today, and this is a wire-protocol constraint rather than a gap. Non-session-part spans carry no session identifiers: they are correlated server-side by travelling in the same export payload as the session-part span, which is why `EmbraceSessionPartBatchedSpanProcessor.forceFlush()` is a deliberate no-op (see `EmbraceUserSessionManager/README.md`). Flushing the pending buffer alone would emit spans the backend cannot attribute to a part.

Logs are self-describing and spans are not, so the asymmetry here is intentional. Making spans flushable means stamping the session and part ids onto every span, which changes what the server reads for correlation and needs its own ticket.

## Coverage

No assertions were lost. The `ttfb`/`fcp` that left `send-log.json` are asserted identically by `log.json`, which this change does not touch, and `lcp` is retained in `handle-204-with-body-logs.json`:

| Golden | ttfb | fcp | lcp |
| --- | --- | --- | --- |
| `send-log.json` before | yes | yes | |
| `send-log.json` after | | | |
| `log.json` (untouched) | yes | yes | |
| `handle-204-with-body-logs.json` | | | yes |

Note `log.json` holds these by accident rather than design: that test selects the first `/logs` request, and the comment above it says it means to take the last. Fixed separately in the stacked follow-up.

## How has this been tested?

- Reproduced the failure locally first, 5 failures matching CI's 6.
- Next 16 Pages after the fix: 31 passed, 0 failed.
- Unit suite: 1041 passed, 0 failed, including the flush, its failure path, and the batch-window boundary.
- Full integration suite: 327 passed, 72 skipped, 0 failed (up from 300; the new e2e test runs across 9 platforms and 3 browsers).
- Goldens are byte-identical across the quiet-period, `SDKControl.flush()` and `log.flush()` implementations, confirming all three produce the same request boundaries.
- Confirmed the boundary test is not vacuous: moving its first tick to 1001ms makes it fail with `expected 1 to equal +0`, so it detects the window rather than passing regardless.
- Confirmed the `Error parsing gzip request: aborted` server noise is pre-existing: it appears 8 times with the change stashed.
- `npm run lint` and `npm run check` from the repo root: clean.

## Checklist

- [ ] Documentation added
- [x] Tests added

### DIFF
--- a/packages/web-sdk/src/api-logs/api/LogAPI/LogAPI.test.ts
+++ b/packages/web-sdk/src/api-logs/api/LogAPI/LogAPI.test.ts
@@ -33,6 +33,7 @@ describe('LogAPI', () => {
       // Mock implementation of LogManager
       message: sinon.stub(),
       logException: sinon.stub(),
+      flush: sinon.stub().resolves(),
     };
     logAPI.setGlobalLogManager(logManager);
     const result = logAPI.getLogManager();
@@ -45,6 +46,7 @@ describe('LogAPI', () => {
       // Mock implementation of LogManager
       message: sinon.stub(),
       logException: sinon.stub(),
+      flush: sinon.stub().resolves(),
     };
     logAPI.setGlobalLogManager(mockLogManager);
 
--- a/packages/web-sdk/src/api-logs/manager/NoOpLogManager/NoOpLogManager.ts
+++ b/packages/web-sdk/src/api-logs/manager/NoOpLogManager/NoOpLogManager.ts
@@ -17,4 +17,8 @@ export class NoOpLogManager implements LogManager {
   ) {
     // no op
   }
+
+  public flush(): Promise<void> {
+    return Promise.resolve();
+  }
 }
--- a/packages/web-sdk/src/api-logs/manager/ProxyLogManager/ProxyLogManager.test.ts
+++ b/packages/web-sdk/src/api-logs/manager/ProxyLogManager/ProxyLogManager.test.ts
@@ -17,6 +17,7 @@ describe('ProxyLogManager', () => {
     mockDelegate = {
       message: sinon.stub(),
       logException: sinon.stub(),
+      flush: sinon.stub().resolves(),
     };
   });
 
--- a/packages/web-sdk/src/api-logs/manager/ProxyLogManager/ProxyLogManager.ts
+++ b/packages/web-sdk/src/api-logs/manager/ProxyLogManager/ProxyLogManager.ts
@@ -30,4 +30,8 @@ export class ProxyLogManager implements LogManager {
   ) {
     this.getDelegate().message(message, level, options);
   }
+
+  public flush(): Promise<void> {
+    return this.getDelegate().flush();
+  }
 }
--- a/packages/web-sdk/src/api-logs/manager/types.ts
+++ b/packages/web-sdk/src/api-logs/manager/types.ts
@@ -21,6 +21,17 @@ export interface LogManager {
   ) => void;
 
   logException: (error: unknown, options?: LogExceptionOptions) => void;
+
+  /**
+   * Exports any logs still held in the batch buffer instead of waiting for the
+   * next scheduled export. Useful before a deliberate teardown, and gives tests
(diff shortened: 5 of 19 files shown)


tests/integration/platforms/next-15-webpack-pages/src/lib/embrace.ts
jpmunz, line 1: is the extra plumbing needed given we have `flush` on the response from `initSDK`, could these call `embraceWebSdk.flush` instead?

jpmunz (commented):
alternatively could the affected tests include some additional `waitForOTelRequest` call to handle the delay?

tests/integration/platforms/next-15-webpack-pages/src/lib/embrace.ts
overbalance: that also flushes traces which increases the OTel requests count. let me design something better. I'll re-request you when I'm done

PR #1473, opened 2026-08-04
build(integration): typecheck the platform harnesses

overbalance:

The eleven integration platform harnesses had no typecheck coverage. `next build` typechecks the six Next apps, but the vite and webpack builds strip types without checking them, so a type error in those five harnesses only surfaced as a runtime test failure, or not at all.

This was not hypothetical: a broken import in a harness survived a full `npm run check` earlier in this stack, because `tests/integration/tsconfig.json` excludes `platforms/`.

- A `check` script per platform: `tsc --noEmit`, or `tsc -b --force` for the four vite apps that use project references. `--force` because a cached `tsbuildinfo` otherwise hides regressions.
- `check-platforms` in `tests/integration/package.json` runs all eleven, wired as a turbo task that `build-platforms` depends on, so the integration run gates on it.
- Fixed `webpack-5/tsconfig.json`, where `moduleResolution: "Node"` (node10) is a hard error under the repo's TypeScript 6.0.3. Changed to `"Bundler"`, matching the other ten platforms and how webpack 5 actually resolves.

Left out deliberately: adding `platforms/` to the integration tsconfig. Each app needs its own JSX, module-resolution and framework types, so one shared program does not fit; each platform's existing tsconfig is the right unit.

## How has this been tested?

- `check-platforms`: all 11 harnesses typecheck.
- `npm run lint` and `npm run check` from the repo root: clean.

## Checklist

- [ ] Documentation added
- [ ] Tests added

### DIFF
--- a/tests/integration/platforms/webpack-5/tsconfig.json
+++ b/tests/integration/platforms/webpack-5/tsconfig.json
@@ -5,7 +5,7 @@
     "module": "ESNext",
     "jsx": "react-jsx",
     "strict": true,
-    "moduleResolution": "Node",
+    "moduleResolution": "Bundler",
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "skipLibCheck": true
--- a/tests/integration/package.json
+++ b/tests/integration/package.json
@@ -5,6 +5,7 @@
   "type": "module",
   "scripts": {
     "check": "tsc",
+    "check-platforms": "for d in platforms/*/; do npm run check --prefix \"$d\" || exit 1; done",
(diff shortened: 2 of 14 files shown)


tests/integration/platforms/webpack-5/tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "module": "ESNext",
     "jsx": "react-jsx",
     "strict": true,
-    "moduleResolution": "Node",
+    "moduleResolution": "Bundler",
```
overbalance, line 8: Node is being removed from TS7. 6.0.3 has marked it as deprecated in anticipation

joaquin-diaz (approved):

PR #1477, opened 2026-08-07
test(performance): raise script duration and main thread thresholds to 200ms

overbalance:

`TOTAL_SCRIPT_DURATION_THRESHOLD_IN_MS`, `MAIN_THREAD_TIME_THRESHOLD_IN_MS` and `SCRIPT_EVAL_THRESHOLD_IN_MS` were set at 175ms and are being exceeded.

Raise those three defaults from 175 to 200. All three remain overridable via env vars through `getFromEnv`; no other thresholds change.

## How was this tested?

Config-only change. The performance suite was not run as part of this change.

## Checklist

- [x] Change is limited to threshold defaults
- [ ] Performance suite run against the new thresholds


overbalance:
Folded into #1476, which now carries both commits.

PR #1485, opened 2026-08-11
test(performance): use median of three lighthouse runs for startup metrics

overbalance:

The lighthouse Main Thread Time check fails intermittently (288ms on #1484, 341ms on a main run) with no SDK change behind it. The metric diffs two single lighthouse samples, and simulated throttling multiplies observed CPU time by 4x, so a scheduling blip on the runner reads as hundreds of ms. Main-branch values swing 56-341ms on effectively identical code.

- Run lighthouse 3 times per page (fresh browser context per run) and threshold-check the median of each metric; log per-run Main Thread Time so a noisy runner is distinguishable from a real regression
- Set the Main Thread Time and Script Evaluation thresholds to 175ms; the median makes that headroom safe again
- Fail loudly instead of skewing the median when a lighthouse run returns no result or an audit has no numeric value
- Upload the run-suffixed lighthouse HTML reports via a glob in the CI artifact step

## How has this been tested?

Full perf suite locally (passes, median diff ~104ms) and under 10 CPU-saturating processes to simulate a starved runner: per-run samples spread up to 49ms but median diffs stayed at +5.0ms / -1.6ms. CI test-performance on this PR passed in 2m19s with a 128.04ms Script Eval median.

## Checklist

- [ ] Documentation added
- [x] Tests added


joseph-embrace (approved):
This makes sense to me. If waiting for 3 LH runs becomes untenable, the other approach is to rerun on failure, i.e. the happy path is still 1 LH run but anomalous results would cause a second run.

PR #1487, opened 2026-08-11
EMBR-13729 perf(instrumentation-dom-state): cap the part-end tree walk at 50k elements

overbalance:
> Stacked on #1442; merge that first.

The part-end tree walk in `DOMStateInstrumentation` is unbounded. It runs synchronously on the `visibilitychange` / `pagehide` path, where the browser grants only a short budget, so a pathological tree spends unload time the keepalive flush needs. Measured, an unbounded walk grows linearly: ~17-22 ms at 131k nodes, worse from there.

- Cap the walk at `DOM_STATE_MAX_TRAVERSED_ELEMENTS` (50,000): far above the 1-2k elements a typical page holds, so it only guards pathology. The bail tests the frontier (visited + queued + about-to-queue is a lower bound on tree size) before pushing children, so the cost stays bounded by the ceiling even on very wide trees, and a tree of exactly the ceiling still reports.
- Past the ceiling the part-end log drops `dom_state.element_count` / `dom_state.average_depth` (absence over fabricated zeros, keeping the document box) and stamps `dom_state.traversal_limit_reached: true`, so the backend can tell "too large to measure" from "no document root". The walk returns its bail reason itself, so the flag cannot drift from the walk's outcome. A `diag.debug` breadcrumb names the ceiling on the bail.
- `dom_state.traversal_limit_reached` is a new attribute key, so it needs the backend allowlist mapping before it shows up in exported telemetry.

## How has this been tested?

Tests pin the boundary from every side: past the ceiling via width and via depth (the two bail terms), exactly at the ceiling, missing root, plus a hand-computed average-depth shift that cannot share drift with the test oracle. Benchmarks in Chromium and WebKit over realistic, wide, deep-chain and balanced trees: the bail caps the walk at ~5-6 ms no matter how large the tree; the whole part-end handler at exactly the ceiling measures ~5-11 ms, which is where the ceiling's rationale in `constants.ts` comes from. The dom-state unit suite, `npm run lint` and `npm run check` pass.

## Checklist

- [ ] Documentation added
- [x] Tests added

### DIFF
--- a/packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
@@ -225,11 +225,9 @@ export class DOMStateInstrumentation extends EmbraceInstrumentationBase {
     });
   }
 
-  // Iterative depth-first traversal that computes element count and summed
-  // depth in one pass. The root passed in sits at depth 1. Runs synchronously
-  // on the pagehide path, so it avoids per-node allocations (parallel stacks
-  // instead of entry objects) and bails at the ceiling rather than spend the
-  // unload budget on a pathological tree.
+  // The root passed in sits at depth 1. Runs synchronously on the pagehide
+  // path, so it uses parallel stacks over per-node entry objects and bails at
+  // the ceiling rather than spend the unload budget on a pathological tree.
   private _traverse(root: Element): {
     elementCount: number;
     totalDepth: number;
@@ -245,11 +243,21 @@ export class DOMStateInstrumentation extends EmbraceInstrumentationBase {
       element = elements.pop(), depth = depths.pop()
     ) {
       elementCount++;
-      if (elementCount > DOM_STATE_MAX_TRAVERSED_ELEMENTS) {
-        return null;
-      }
       totalDepth += depth;
       const { children } = element;
+      // Popped, queued and about-to-queue are disjoint, so their sum is a
+      // lower bound on the tree size: exceeding the ceiling here proves the
+      // tree is over it before the frontier is pushed, keeping the bail cost
+      // bounded by the ceiling even on wide trees.
+      if (
+        elementCount + elements.length + children.length >
+        DOM_STATE_MAX_TRAVERSED_ELEMENTS
+      ) {
+        this._diag.debug(
+          'dom-state tree walk bailed: the tree exceeds the traversal ceiling',
+        );
+        return null;
+      }
       for (let i = 0; i < children.length; i++) {
         elements.push(children[i]);
         depths.push(depth + 1);
--- a/packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
@@ -14,6 +14,7 @@ import {
   ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_TIMESTAMP,
   ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_VIEWPORT_HEIGHT,
   ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_VIEWPORT_WIDTH,
+  ATTR_DOM_STATE_TRAVERSAL_LIMIT_REACHED,
   DOM_STATE_EVENT_NAME,
   DOM_STATE_MAX_TRAVERSED_ELEMENTS,
 } from './constants.ts';
@@ -178,7 +179,9 @@ export class DOMStateInstrumentation extends EmbraceInstrumentationBase {
             // The traversal counts its root, so elementCount is at least 1.
             [ATTR_DOM_STATE_AVERAGE_DEPTH]: tree.totalDepth / tree.elementCount,
           }
-        : {}),
+        : root
+          ? { [ATTR_DOM_STATE_TRAVERSAL_LIMIT_REACHED]: true }
+          : {}),
       ...(measurement
         ? {
             [ATTR_DOM_STATE_DOCUMENT_HEIGHT]: measurement.documentHeight,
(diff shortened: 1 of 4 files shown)


packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
```diff
@@ -257,24 +263,42 @@ export class DOMStateInstrumentation extends EmbraceInstrumentationBase {
...
       elementCount++;
-      totalDepth += entry.depth;
-      const { children } = entry.element;
+      totalDepth += depth;
+      const { children } = element;
+      // Popped, queued and about-to-queue are disjoint, so their sum is a
+      // lower bound on the tree size: exceeding the ceiling here proves the
+      // tree is over it before the frontier is pushed, keeping the bail cost
+      // bounded by the ceiling even on wide trees.
+      if (
```
jpmunz, line 273: lgtm, though I do think I found the previous way this algorithm was written with the single `stack` object more readable, was there a reason for altering the structure here over putting the guard on that previous version?

jpmunz (approved):

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
overbalance: the original way increased the frame count 4x. Let me find the benchmark and send on slack

overbalance:
@jpmunz benchmark results comparing this walk against the unbounded per-node-object version, in case they help the review.

Method: both walkers inline in a throwaway test file (the first version copied verbatim from the base branch), timed over the same trees in Chromium and WebKit via web-test-runner, median of 7 samples of 5 walks each. Sanity assertions confirmed both return identical counts under the ceiling.

| Layout | Chromium first → current | WebKit first → current |
|---|---|---|
| realistic ~1.5k | 0.20 → 0.18 ms | 0.20 → 0.20 ms |
| wide flat 49k (both walk all) | 5.18 → 5.12 ms | 5.00 → 4.80 ms |
| wide flat 60k (over ceiling) | 6.04 → **0.00 ms** | 5.60 → **0.00 ms** |
| deep chain 45k (both walk all) | 4.28 → 4.86 ms | 6.00 → 4.40 ms |
| deep chain 60k (over) | 5.74 → 5.26 ms | 9.40 → 8.80 ms |
| binary 32k (both walk all) | 3.32 → 3.60 ms | 3.20 → 2.60 ms |
| binary 131k (over) | 16.66 → **5.46 ms** | 22.00 → **6.40 ms** |

Takeaways:

- The frontier bail is the real win: worst case is capped at ~5-6 ms no matter how large the tree, where the unbounded version grows linearly (17-22 ms at 131k nodes). On very wide trees the bail is effectively free, since it trips before the frontier is ever pushed.
- Under the ceiling, where both walk the full tree, raw speed is parity: WebKit slightly favors the parallel stacks everywhere, Chromium is a wash (slightly slower on two layouts). The stacks' remaining benefit there is dropping one short-lived object allocation per node on the unload path.
- The whole part-end handler (walk + document measurement + log emission) at exactly the 50k ceiling measures ~5-11 ms median, which is where the ceiling's rationale in `constants.ts` comes from.

jpmunz:
> @jpmunz benchmark results comparing this walk against the unbounded per-node-object version, in case they help the review.
> 
> Method: both walkers inline in a throwaway test file (the first version copied verbatim from the base branch), timed over the same trees in Chromium and WebKit via web-test-runner, median of 7 samples of 5 walks each. Sanity assertions confirmed both return identical counts under the ceiling.
> Layout 	Chromium first → current 	WebKit first → current
> realistic ~1.5k 	0.20 → 0.18 ms 	0.20 → 0.20 ms
> wide flat 49k (both walk all) 	5.18 → 5.12 ms 	5.00 → 4.80 ms
> wide flat 60k (over ceiling) 	6.04 → **0.00 ms** 	5.60 → **0.00 ms**
> deep chain 45k (both walk all) 	4.28 → 4.86 ms 	6.00 → 4.40 ms
> deep chain 60k (over) 	5.74 → 5.26 ms 	9.40 → 8.80 ms
> binary 32k (both walk all) 	3.32 → 3.60 ms 	3.20 → 2.60 ms
> binary 131k (over) 	16.66 → **5.46 ms** 	22.00 → **6.40 ms**
> 
> Takeaways:
> 
>     * The frontier bail is the real win: worst case is capped at ~5-6 ms no matter how large the tree, where the unbounded version grows linearly (17-22 ms at 131k nodes). On very wide trees the bail is effectively free, since it trips before the frontier is ever pushed.
> 
>     * Under the ceiling, where both walk the full tree, raw speed is parity: WebKit slightly favors the parallel stacks everywhere, Chromium is a wash (slightly slower on two layouts). The stacks' remaining benefit there is dropping one short-lived object allocation per node on the unload path.
> 
>     * The whole part-end handler (walk + document measurement + log emission) at exactly the 50k ceiling measures ~5-11 ms median, which is where the ceiling's rationale in `constants.ts` comes from.

yep applying the ceiling makes sense to me, my question from https://github.com/embrace-io/embrace-web-sdk/pull/1487#discussion_r3768743743 had been about using a single `stack` array as oppose to the two separate `elements` + `depths` arrays, it seems like that difference was more minor?

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
overbalance: The table doesn't isolate that change. The only rows where both walkers do identical work are the under-ceiling ones, and those are a wash: Chromium is slightly slower with the two arrays on deep chain and binary, WebKit slightly faster. The 4x I mentioned was allocation count rather than wall time, and I don't have a measurement showing that costs anything on this path, so it doesn't buy enough to pay for the readability.

Reverted the walk to the single `{element, depth}` stack with the guard on it. The bail expression is unchanged apart from `stack.length` in place of `elements.length`, and since the two arrays were pushed in lockstep those are equal at every iteration, so the boundary tests pass as-is.

PR #1491, opened 2026-08-13
EMBR-13582 fix(instrumentation-document-load): collect from an unbuffered navigation observer

overbalance:

`DocumentLoadInstrumentation` waited for navigation timing with a `readyState` check, a `load` listener and a `setTimeout(…, 0)`. The timeout guessed when the browser writes `loadEventEnd` instead of waiting for a signal, and `onDisable()` could not cancel it, so disabling mid-load still emitted spans. Safari reported no document load, fetch or resource spans at all.

- `onEnable` branches on `loadEventEnd`: already finished, collect on a microtask; still loading, subscribe a `navigation` PerformanceObserver with `buffered: false`.
- `_collectIfLoadEventFinished()` is the single entry point for both triggers. It re-checks `loadEventEnd`, so a zeroed entry can never produce a span with zeroed timings, and disconnects the observer once it has collected.
- `onDisable()` is now just `disconnect()`, so disabling is exact.
- The navigation entry is treated as optional. WebKit reports none at all for `about:blank`, popups and `srcdoc` iframes, and this code runs from the constructor inside `initSDK`'s try block, so an unguarded read there would fail the whole SDK init rather than just this instrumentation.
- Collection is wrapped so a throw is reported through `diag` and the observer is still disconnected. The one path that knowingly drops a page load, a navigation entry with no `fetchStart`, now warns instead of going silent.

`buffered` must stay `false`: mid-load, WebKit answers a buffered subscription with the unfinished entry and counts it as the observer's one notification, so the finalized entry never arrives.

The microtask covers an unrelated race. Under `registerGlobally: false` the tracer provider arrives from `registerInstrumentations` later in the same task as the constructor, so collecting synchronously would start spans on a no-op tracer. Under the default `registerGlobally: true` the global provider is already installed before instrumentations are constructed.

Behavior change: nothing is collected if `loadEventEnd` never becomes non-zero, where before a span ended at wall clock. That needs the load event to never fire, in which case the old listener never ran either.

## How has this been tested?

- 50 unit tests, including a fake `PerformanceObserver` modelling WebKit's buffered semantics. Both `buffered: false` and the microtask fail tests if removed.
- The `loadEventEnd` re-check and the optional-entry guard were each confirmed by deleting them and watching tests fail, so neither can be dropped silently.
- Confirmed in real Safari 26.6, not only Playwright's WebKit build.
- `npm run check` and `npm run lint` from the repo root.

Integration coverage for the mid-load path needs a `server/` page fixture and delayed-subresource endpoint, not included here.

## Checklist

- [ ] Documentation added
- [x] Tests added

### DIFF
--- a/packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
@@ -567,7 +567,10 @@ export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<Docu
 
     // buffered must stay false: mid-load, WebKit answers a buffered
     // subscription with the unfinished entry and counts it as the observer's
-    // one notification, so the finalized entry never arrives.
+    // one notification, so the finalized entry never arrives. Measured on
+    // Safari 26.6 and Playwright WebKit, 2026-08-13; the performance-timeline
+    // spec does not say whether a replay preempts a later notification, so
+    // treat neither engine's choice as guaranteed.
     this._navigationObserver = createPerformanceObserver(
       'navigation',
       () => {
--- a/packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
@@ -127,14 +127,26 @@ export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<Docu
       return;
     }
 
+    // Latched before collecting, not after: a throw part way through has
+    // already emitted spans, so retrying on a later trigger would duplicate
+    // them. Both triggers reach this method, so the catch covers both.
     this._performanceCollected = true;
-    this._collectPerformance(getPerformanceNavigationEntries());
-    this._disconnectObserver();
+
+    try {
+      this._collectPerformance(getPerformanceNavigationEntries());
+    } catch (e) {
+      this._diag.error('failed to collect the document load', e);
+    } finally {
+      this._disconnectObserver();
+    }
   }
 
-  /* loadEventEnd is written only after every load event handler returns. */
+  /* loadEventEnd is written only after every load event handler returns. WebKit
+   * reports no navigation entry at all for about:blank, popups and srcdoc
+   * iframes, so the entry is optional rather than guaranteed. */
   private _isLoadEventFinished(): boolean {
-    return performance.getEntriesByType('navigation')[0].loadEventEnd > 0;
+    const [navigationTiming] = performance.getEntriesByType('navigation');
+    return (navigationTiming?.loadEventEnd ?? 0) > 0;
   }
 
   private _disconnectObserver(): void {
@@ -157,6 +169,11 @@ export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<Docu
         entries,
       );
       if (!rootSpan) {
+        // The only point where the SDK knows for certain that the page load is
+        // being dropped, and collection is latched so nothing will retry.
+        this._diag.warn(
+          'navigation entry has no fetchStart, document load will not be collected',
+        );
         return;
       }
       context.with(trace.setSpan(context.active(), rootSpan), () => {
@@ -542,9 +559,10 @@ export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<Docu
 
   public override onEnable(): void {
     // An unbuffered subscription registered after the load event is never
-    // notified, so collect straight away instead of observing. Deferred because
-    // onEnable runs from the constructor, before the tracer provider is wired:
-    // spans started synchronously would go unrecorded.
+    // notified, so collect straight away instead of observing. Deferred one
+    // microtask because under registerGlobally: false the tracer provider
+    // arrives from registerInstrumentations later in the same task as this
+    // constructor, and spans started before it go to a no-op tracer.
     if (this._isLoadEventFinished()) {
       queueMicrotask(() => {
         if (this._isEnabled) {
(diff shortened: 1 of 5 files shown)


packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
```diff
@@ -95,23 +96,11 @@ export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<Docu
...
     if (this._config.enabled) {
       this.enable();
     }
   }
 
-  protected override init() {
-    this._diag.debug('Initializing document load instrumentation');
-    return undefined;
-  }
-
```
overbalance, line 114: this is an outlier in our repo and was negatively affecting code coverage %

packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
```diff
@@ -539,12 +517,46 @@ export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<Docu
...
-    this._waitForPageLoad();
+    this._stopObserving();
+
+    /*
+     * Collection always waits for an observer notification: onEnable runs from
+     * the constructor, before the tracer provider is wired on, so spans started
+     * synchronously here would go unrecorded.
+     *
+     * Buffering is only safe once the load has completed. Mid-flight, WebKit
+     * answers a buffered subscription with the incomplete entry and counts that
```
jpmunz: is there a reference for this behaviour? Specifically wondering if it is intentional or a bug that may eventually go away

packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
```diff
@@ -539,12 +517,46 @@ export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<Docu
...
+     * as the observer's one notification, losing the load entirely.
+     */
+    const buffered = this._hasLoadEventCompleted();
+
+    this._navigationObserver =
+      createPerformanceObserver<PerformanceNavigationTiming>(
+        'navigation',
+        () => {
+          this._stopObserving();
+          this._collectPerformance();
```
jpmunz: is it enough that the navigation timing event has been observed or do we also need to wait for `navigationTiming.loadEventEnd` to be non-zero?

overbalance: That was an invariant relying on specific browser behavior. I refactored to check for a non-zero `loadEventEnd` explicitly, so collection re-checks it on every notification and only collects once it is set.

packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
overbalance: No reference I could find, no WebKit bug and no write-up. I verified it repeatedly on Safari 26.6 and Playwright's WebKit, and it reproduces consistently.

On intentional vs bug: the performance-timeline spec is silent. It says a buffered replay is delivered in the first callback after `observe()`, but says nothing about whether an entry mutated in place re-notifies, or whether the replay preempts a later notification for the same entry. So WebKit isn't violating anything, it's choosing differently from Chromium and Gecko, which deliver both.

Which means we shouldn't rely on any engine's choice, since Chromium's second notification isn't guaranteed either. Collection re-checks `loadEventEnd` on every notification and only collects once it's non-zero, so we're correct whether or not WebKit ever changes. I've added the measurement details and that reasoning to the comment.

packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
```diff
@@ -540,11 +552,41 @@ export class DocumentLoadInstrumentation extends EmbraceInstrumentationBase<Docu
...
+    // one notification, so the finalized entry never arrives. Measured on
+    // Safari 26.6 and Playwright WebKit, 2026-08-13; the performance-timeline
+    // spec does not say whether a replay preempts a later notification, so
+    // treat neither engine's choice as guaranteed.
+    this._navigationObserver = createPerformanceObserver(
+      'navigation',
+      () => {
+        this._collectIfLoadEventFinished();
+      },
+      { buffered: false, diag: this._diag },
```
joaquin-diaz, line 579: Is there a scenario where load finishes between line 559 and the creation of the observer? Maybe it's worth defining the observer first, then check if we missed the buffer and add a flag in `_collectIfLoadEventFinished` to avoid capturing it twice?

overbalance: As I understand it, no. Synchronous  functions don't have other sync events firing in the middle of their execution.

joaquin-diaz: You're right, no need to do anything!

packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
jpmunz: ok that sounds good, thanks for the additional context!

packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/DocumentLoadInstrumentation.ts
jpmunz: would the performance observer trigger again in that case? Specifically I'm wondering about a scenario like:
* performance observer triggers on 'navigation'
* we check `loadEventEnd` and it's 0 so we don't collect
* at some point later `loadEventEnd` is updated to be non-zero
* will performance observer trigger on 'navigation' again so we get a chance to collect?

overbalance: In Chrome and Firefox the event fires twice: once with zero, once with e.g. 312. In Safari, it only fires once when it's complete **unless** you pull the buffer before `load`, then it will return 0 and not fire again. Safari is "missing" the first event compared to the others so this code is safe as long as buffered is false.

jpmunz (approved):

PR #1494, opened 2026-08-13
EMBR-13621 fix(instrumentation-server-timing): read server timings from an observer notification

overbalance:

`ServerTimingInstrumentation` read the navigation entry synchronously in `onEnable` whenever `readyState` was already `complete`. `onEnable` runs from the instrumentation's constructor, but `initSDK` only wires the logger provider onto instrumentations afterwards, via `registerInstrumentations`. Those logs therefore went to a logger that records nothing, and the once-per-page collection guard then latched so they were never retried.

That is live for an SDK initialized with `registerGlobally: false` after the page has finished loading, which is exactly what a late-initializing embedded component does.

- Read the server timings from a `navigation` PerformanceObserver notification instead of synchronously in `onEnable`, so the read lands in a later task, after the providers are wired. The subscription is buffered, which is `createPerformanceObserver`'s default, so an entry already on the timeline is replayed rather than waited for.
- Drop the `readyState` check and the `load` listener. Server timings arrive in the response headers, so they are on the navigation entry before any script runs and there is nothing to wait for beyond the entry itself.
- `onDisable()` is now just `disconnect()`, so disabling is exact.
- Default the entries to an empty array rather than guarding on length, since a `for…of` over an empty array is already a no-op.
- Warn when the entry type is unobservable. Nothing else re-triggers the read, so that case would otherwise drop server timings for the page with no diagnostic at all.
- Skip resubscribing on a re-enable once the timings have been collected.

`buffered` is deliberately left at its default here, the opposite of the `navigation` observer in `DocumentLoadInstrumentation`, which must subscribe unbuffered. The two requirements are inverted: server timings are complete on the entry from the start, so a replay carries everything and is in fact required, since the SDK usually starts after the entry was buffered. The comment says so at the subscription, so the difference between the two callers is not mistaken for an inconsistency.

## How has this been tested?

13 unit tests, covering:

- No logs are emitted while constructing, which is the regression this fixes.
- Logs are emitted whatever `readyState` is when the SDK starts, and none at all if it is disabled before the entry arrives.
- Nothing is emitted when the `serverTiming` array is empty or no navigation entry exists, which is what allows the length guard to go. Both cases now deliver an entry first, so neither passes vacuously.
- The subscription carries the buffered flag, and an unobservable entry type warns and collects nothing.

Both guards were confirmed by deleting them: forcing `buffered: false` fails 8 tests, and removing the collect-once latch fails the duplicate-collection test. Running the file against `main`'s implementation fails 6 of the 13, `does not emit while constructing` among them, so the regression is pinned rather than assumed.

Also checked in the demo (`npm run dev`), which serves a `Server-Timing` header from the dev server: the waterfall page emits exactly three `emb-server-timing` logs on load, one per header entry with the expected name, duration and description, and the soft-nav page adds none across five soft navigations that each roll the session part.

`npm run check` and `npm run lint` from the repo root.

## Checklist

- [ ] Documentation added
- [x] Tests added

### DIFF
--- a/packages/web-sdk/src/instrumentations/server-timing/ServerTimingInstrumentation/ServerTimingInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/server-timing/ServerTimingInstrumentation/ServerTimingInstrumentation.ts
@@ -1,5 +1,6 @@
 import { SeverityNumber } from '@opentelemetry/api-logs';
 import { EMB_TYPES, KEY_EMB_TYPE } from '../../../constants/index.ts';
+import { createPerformanceObserver } from '../../../utils/index.ts';
 import { EmbraceInstrumentationBase } from '../../EmbraceInstrumentationBase/index.ts';
 import {
   KEY_EMB_SERVER_TIMING_DESCRIPTION,
@@ -10,7 +11,7 @@ import {
 import type { ServerTimingInstrumentationArgs } from './types.ts';
 
 export class ServerTimingInstrumentation extends EmbraceInstrumentationBase {
-  private readonly _onLoad: () => void;
+  private _navigationObserver: PerformanceObserver | null = null;
   private _performanceCollected = false;
 
   public constructor({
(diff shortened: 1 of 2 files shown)


packages/web-sdk/src/instrumentations/server-timing/ServerTimingInstrumentation/ServerTimingInstrumentation.ts
```diff
@@ -60,13 +73,9 @@ export class ServerTimingInstrumentation extends EmbraceInstrumentationBase {
     const navEntries = performance.getEntriesByType(
       'navigation',
     ) as PerformanceNavigationTiming[];
-    const serverTiming = navEntries[0]?.serverTiming;
-
-    if (!serverTiming?.length) {
-      return;
-    }
+    const serverTimingEntries = navEntries[0]?.serverTiming ?? [];
```
overbalance, line 89: easier to see it's an array

jpmunz (approved):
lgtm depending on answer for https://github.com/embrace-io/embrace-web-sdk/pull/1491#discussion_r3779293546

PR #1498, opened 2026-08-20
feat(sdk): record SDK load and init timestamps on session parts

overbalance:

`emb.sdk_startup_duration` only measures the `initSDK` call body, so the bundle download and any delay before the app calls `initSDK` are invisible.

- Adds `emb.sdk_load_timestamp` (when the SDK's code first evaluated) and `emb.sdk_init_timestamp` (when `initSDK` was entered), both epoch millis. The gap between them is how long the page held a loaded SDK before initializing it.
- All three timings are stamped on the cold-start session part only, since they are page-scoped constants that told later parts nothing. This narrows `emb.sdk_startup_duration`, which used to land on every part.
- `utils/sdkLoadTime.ts` captures `performance.now()` at module evaluation and stores a raw origin offset, so the epoch conversion still routes through `epochMillisFromOrigin`. Works for CDN and bundled installs alike.
- `recordSDKStartupDuration` becomes `recordSDKStartupTimings`.

Note for reviewers: the two new attributes are dropped at ingestion until mapper entries exist for them.

## How has this been tested?

- `npm run check` and `npm run validate` clean. Unit tests cover the epoch shape, load-at-or-before-init ordering, and presence on the first part but not the second.
- Goldens regenerated. All three keys sit in `IGNORED_ATTRIBUTES_LIST`, which compares key presence only.

## Checklist

- [x] Documentation added
- [x] Tests added


packages/web-sdk/src/sdk/initSDK.ts
```diff
@@ -363,9 +364,11 @@ export const initSDK = (
       registry.register(sdkControl);
     }
 
-    userSessionManager.recordSDKStartupDuration(
-      perf.getNowMillis() - initSDKStart,
-    );
+    userSessionManager.recordSDKStartupTimings({
+      initDurationMillis: perf.getNowMillis() - initSDKStart,
+      loadTimestamp: perf.epochMillisFromOrigin(SDK_LOAD_ORIGIN_OFFSET_MILLIS),
```
jpmunz, line 369: using epoch timestamps here one thing that gets missed is the relative time from time origin to sdk load, e.g. the value of `SDK_LOAD_ORIGIN_OFFSET_MILLIS` directly which might be useful to include either in addition to or instead of `loadTimestamp`, similarly for the milliseconds between initSDKStart and time origin

jpmunz (approved):
change lgtm, just a question on whether we want to track those additional values

packages/web-sdk/src/sdk/initSDK.ts
overbalance: good call, I dropped them in favor of offsets only, and only on cold start parts

packages/web-sdk/src/utils/sdkLoadOriginOffset.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Milliseconds from the time origin to the moment the SDK's code first ran on
+ * the page. For a CDN install that is script evaluation; for a bundled install
+ * it is whenever the host chunk executed.
+ */
+// Evaluated at import, before any PerformanceManager or initSDK's window guard,
+// so it reads the clock directly and must not throw outside a browser (SSR).
+export const SDK_LOAD_ORIGIN_OFFSET =
+  typeof performance === 'undefined' ? 0 : performance.now();
```
jpmunz: would we need to guard against `now` being undefined as well?

overbalance: I dug in a little deeper to verify and I found that otel core reads `performance` without a guard so the SDK fails regardless. I'll remove.

PR #1500, opened 2026-08-21
feat: emit the document fetch span as a ux.resource_fetch span

joaquin-diaz:

The document fetch span (the main HTML document's own network timing) has never carried `emb.type`, so the backend's web-resource mapper (`spans_web_resource_mapper.go`, keyed on `emb.type=ux.resource_fetch`) never picks it up. Every other resource on the page (scripts, stylesheets, images, XHR/fetch...) shows up as a web resource; the main document itself doesn't, even though the span already carries equivalent timing data.

- `DocumentLoadInstrumentation.ts`: extracted the attribute-setting logic shared by resource fetch spans and the document fetch span (`emb.type`, delivery type, initiator type, render-blocking status, response status, sizes, CORS/cache/incomplete diagnostic flags) into `_addResourceAttributesToSpan`, and call it for the document fetch span too. It now sets `emb.type=ux.resource_fetch` and the same attributes as any other resource span.
- `utils.ts`: `getPerformanceNavigationEntries` now also reads `transferSize`, `initiatorType`, `responseStatus`, `deliveryType`, `renderBlockingStatus` off the navigation timing entry, since none of those are part of `PerformanceTimingNames`.

## How has this been tested?

```
Row 1:
──────
date:                   2026-08-21
app_id:                 ppduz
app_version:            2.26.0
os:                     web
os_variant:             macOS
framework:              native
sdk_version:            2.26.0
surface_name:           /
surface_label:          Embrace Web SDK React Demo
browser_name:           Chrome
browser_version:        151.0.0.0
id:                     bd2ea71b4ec0d799
render_blocking_status: non-blocking
initiator_type:         navigation
delivery_type:          cache
encoded_body_size:      468
decoded_body_size:      468
transfer_size:          0
response_status:        200
resource_type:          other
start_time:             2026-08-21 16:48:19.362
end_time:               2026-08-21 16:48:19.364
duration_ms:            2
page_load_id:           472b59c712bc3a74
page_url:               http://localhost:4847/
protocol:               http
domain:                 localhost:4847
path:                   /
```

### DIFF
--- a/packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/utils.ts
+++ b/packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/utils.ts
@@ -20,11 +20,24 @@ export type EmbracePerformanceResourceTiming = PerformanceResourceTiming & {
 };
 
 /**
+ * Adds new browser features not yet in TypeScript's DOM lib (as of Oct 2025):
+ * - notRestoredReasons: Limited availability https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming/notRestoredReasons
+ * Only the top-level document's own reasons are read; nested iframe reasons
+ * (the `children` field) aren't relevant to the document fetch span.
+ */
+export type EmbracePerformanceNavigationTiming = PerformanceNavigationTiming & {
+  deliveryType?: 'cache' | '';
+  renderBlockingStatus?: 'blocking' | 'non-blocking';
+  notRestoredReasons?: { reasons?: { reason: string }[] } | null;
+};
+
+/**
  * PerformanceTimingNames only covers navigation timing marks, so transferSize,
- * initiatorType, responseStatus, deliveryType and renderBlockingStatus - all
- * present on the underlying PerformanceNavigationTiming entry - are read
- * separately below. This lets the document fetch span carry the same
- * attributes as resource fetch spans.
+ * initiatorType, responseStatus, deliveryType, renderBlockingStatus, type and
+ * notRestoredReasons - all present on the underlying PerformanceNavigationTiming
+ * entry - are read separately below. This lets the document fetch span carry
+ * the same attributes as resource fetch spans, plus the navigation-specific
+ * ones no resource entry ever has.
  */
 export type EmbracePerformanceNavigationEntries = PerformanceEntries & {
   transferSize?: number;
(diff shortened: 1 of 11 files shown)


packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/utils.ts
```diff
@@ -9,27 +9,61 @@ import { hasKey, PerformanceTimingNames } from '@opentelemetry/sdk-trace-web';
...
 
-  return entries;
+/**
+ * PerformanceTimingNames only covers navigation timing marks, so transferSize,
+ * initiatorType, responseStatus, deliveryType and renderBlockingStatus - all
+ * present on the underlying PerformanceNavigationTiming entry - are read
+ * separately below. This lets the document fetch span carry the same
+ * attributes as resource fetch spans.
+ */
+export type EmbracePerformanceNavigationEntries = PerformanceEntries & {
```
jpmunz, line 42: could this be `EmbracePerformanceResourceTiming &`  since the navigation entry is a subclass?

packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/utils.ts
```diff
@@ -9,27 +9,61 @@ import { hasKey, PerformanceTimingNames } from '@opentelemetry/sdk-trace-web';
...
+    keys.forEach((key: PerformanceTimingNames) => {
+      if (hasKey(performanceNavigationTiming, key)) {
+        const value = performanceNavigationTiming[key];
+        if (typeof value === 'number') {
+          entries[key] = value;
+        }
+      }
+    });
+
+    entries.transferSize = performanceNavigationTiming.transferSize;
```
jpmunz, line 72: would these be temporary assuming `PerformanceTimingNames` is updated at some point to include them?

packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/utils.ts
joaquin-diaz: Honestly I wanted to remove all references to `PerformanceEntries` because it's being imported from `@opentelemetry/sdk-trace-web` and its going to be deprecated soon but I didn't want to make all those changes in this PR yet. The difference is that `PerformanceEntries` makes all keys optional and makes the instrumentation check for every one of them while in reality some of them are already widely available so we don't need to guard for it. 

I'd like to make this PR just additive and then do a clean up of PerformanceEntries and modernize the instrumentation, we forked this one and worked on top of what we currently had but it was already based on old code

packages/web-sdk/src/instrumentations/document-load/DocumentLoadInstrumentation/utils.ts
joaquin-diaz: On the contrary, we should clean the instrumentation to not depend on `@opentelemetry/sdk-trace-web` at all

jpmunz (approved):

PR #1508, opened 2026-08-26
EMBR-13803 fix(transport): drain fetch response body to release keepalive quota

overbalance:

`FetchTransport` never reads the response body. Chromium returns a request's share of the cumulative 64KiB keepalive quota only once the body is read to its end, and Blink skips its own drain when the response carries `Cache-Control: no-store`, which any gateway or CDN in front of the endpoint can set. Where it is set, exports leak the quota until it is exhausted and every later keepalive export fails silently. The transport's counters released at response headers, so it kept setting `keepalive: true` on budget the browser no longer had.

- Drain the response body once the fetch resolves, without awaiting it: the status already decides the export outcome, so a collector that stalls mid-body cannot hold up the caller.
- Release the keepalive counters when the drain settles rather than in `finally`, so they track when the browser actually returns the quota. The abort signal stays armed through the drain, so a stalled body cannot hold the budget past the request timeout.
- Read errors log at `diag.debug` and never change the export outcome. Cancelling would take Blink's client-abort path, which never reaches the release. The reader lock is released so a response handed to a later export can still be drained.
- The local collector mirrors the real ingest response, with `EMB_NO_STORE=1` as an opt-in switch for reproducing the leak. Turbo runs in strict env mode, so the flag needs a `passThroughEnv` entry to reach the collector task under `npm run dev`.
- Same defect as https://github.com/open-telemetry/opentelemetry-js/pull/7002, whose fix does not reach us: the export delegate injects this transport, not `otlp-exporter-base`'s. The approach follows https://github.com/open-telemetry/opentelemetry-js/pull/7070.

## How has this been tested?

Seven unit tests in `FetchTransport.test.ts` cover the drain on success and on a retryable status, the export settling before the body ends, locked and erroring bodies, and the budget being held until the drain settles then released when it fails. File passes 32/32, plus `npm run check` and `npm run lint` from the repo root.

Measured in headless Chromium against the local collector run with `EMB_NO_STORE=1`, so `/v2/spans` answers with `Cache-Control: no-store`. Each request posts a gzipped body of 16,529 bytes with `keepalive: true`, sequentially.

| | outcome |
| --- | --- |
| Raw `fetch`, body left unread | 3 succeed, the 4th throws `TypeError: Failed to fetch` at 49,587 B cumulative |
| Through `FetchTransport` | 20 of 20 succeed, ~330 KB cumulative, every one sent with `keepalive: true` |

The undrained arm wedges one request short of the 64KiB budget and never recovers. The drained arm passes roughly five times the budget without a failure, and a passthrough spy on `window.fetch` confirms none of the 20 silently fell back to `keepalive: false`.

## Performance Impact

Benchmarked in headless Chromium against the local collector, 1000 responses per arm, ingest-shaped reply (200, 1-byte body). Run twice, once with the default headers and once with `EMB_NO_STORE=1`.

| measurement | default | `no-store` |
| --- | --- | --- |
| 1000 drains, timed in batches | 15.2 ms total, 15.2 us per response | 11.8 ms total, 11.8 us per response |
| 1000 sequential round trips, drain awaited | 337.9 ms | 347.0 ms |
| 1000 sequential round trips, body left unread | 297.1 ms | 306.7 ms |

A drain costs ~12-15 us of main-thread time per response, so ~15 ms spread across 1000 exports. The awaited round-trip rows put an upper bound of ~0.04 ms per request on it; the shipped code does not await the drain, so nothing on the export path pays even that. `no-store` does not make the drain measurably more expensive, which is expected: the body is one byte either way, only Blink's own buffering differs.

## Checklist

- [ ] Documentation added
- [x] Tests added

### DIFF
--- a/packages/web-sdk/src/transport/FetchTransport/FetchTransport.ts
+++ b/packages/web-sdk/src/transport/FetchTransport/FetchTransport.ts
@@ -23,6 +23,51 @@ export function _resetKeepaliveTracking(): void {
   inflightKeepaliveCount = 0;
 }
 
+/**
+ * Reads the response body to its end and discards it.
+ *
+ * Chromium returns the request's share of the keepalive quota only once the
+ * body has been read to the end, and it skips the buffering consumer that would
+ * otherwise drain it when the response carries `Cache-Control: no-store`, which
+ * collectors commonly send. Cancelling is the client-abort path and measures
+ * far slower, so the body is read rather than cancelled.
+ *
+ * @see https://fetch.spec.whatwg.org/#fetch-processresponseendofbody
+ * @see https://github.com/chromium/chromium/blob/1af7c3cde84323e827f77d8066ca23da811203b8/third_party/blink/renderer/core/fetch/fetch_manager.cc#L818-L836
+ */
+async function drainResponseBody(response: Response): Promise<void> {
+  try {
+    // Empty and opaque responses have no body to read.
+    const body = response.body;
+    if (!body) {
+      return;
+    }
+
+    // Throws when the body is already locked to another reader, which happens
+    // when the same response is handed to more than one export.
+    const reader = body.getReader();
+    try {
+      // Chunks are dropped as they arrive so a large response is never buffered.
+      let chunk = await reader.read();
+      while (!chunk.done) {
+        chunk = await reader.read();
+      }
+    } finally {
+      // A held reader keeps the body locked, and a later export handed the same
+      // response could then not drain it.
+      reader.releaseLock();
+    }
+  } catch (error) {
+    // The export outcome is decided by the response status, a body that cannot
+    // be read must not change it.
+    const drainError =
+      error instanceof Error ? error : new Error(String(error));
+    diag.debug(
+      `Fetch transport failed to drain response body: ${drainError.message}`,
+    );
+  }
+}
+
 export class FetchTransport implements IExporterTransport {
   public constructor(private readonly _config: FetchRequestParameters) {}
 
@@ -103,6 +148,21 @@ export class FetchTransport implements IExporterTransport {
     }
 
     let keepalive = false;
+    let reservedBytes = 0;
+
+    const clearTimeoutIfSet = () => {
+      if (timeoutId !== undefined) {
+        clearTimeout(timeoutId);
+      }
+    };
+
+    // The quota frees only once the body drains, so the drain owns the release.
+    const releaseKeepalive = () => {
+      clearTimeoutIfSet();
+      inflightKeepaliveBytes -= reservedBytes;
+      inflightKeepaliveCount--;
+    };
+    let drainOwnsRelease = false;
 
     try {
       if (this._config.compression === 'gzip') {
@@ -126,7 +186,8 @@ export class FetchTransport implements IExporterTransport {
       keepalive = !wouldExceedBytes && !wouldExceedCount;
 
       if (keepalive) {
-        inflightKeepaliveBytes += request.byteLength;
+        reservedBytes = request.byteLength;
+        inflightKeepaliveBytes += reservedBytes;
         inflightKeepaliveCount++;
       } else {
         const reason = wouldExceedBytes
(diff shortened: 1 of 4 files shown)


packages/web-sdk/src/transport/FetchTransport/FetchTransport.ts
```diff
@@ -23,6 +23,51 @@ export function _resetKeepaliveTracking(): void {
...
+ * Reads the response body to its end and discards it.
+ *
+ * Chromium returns the request's share of the keepalive quota only once the
+ * body has been read to the end, and it skips the buffering consumer that would
+ * otherwise drain it when the response carries `Cache-Control: no-store`, which
+ * collectors commonly send. Cancelling is the client-abort path and measures
+ * far slower, so the body is read rather than cancelled.
+ *
+ * @see https://fetch.spec.whatwg.org/#fetch-processresponseendofbody
+ * @see https://github.com/chromium/chromium/blob/1af7c3cde84323e827f77d8066ca23da811203b8/third_party/blink/renderer/core/fetch/fetch_manager.cc#L818-L836
```
jpmunz, line 36: was there also an issue on chromium for this we could link to?

packages/web-sdk/src/transport/FetchTransport/FetchTransport.ts
```diff
@@ -143,6 +204,18 @@ export class FetchTransport implements IExporterTransport {
         signal,
       });
 
+      // Not awaited: the status already decides the export outcome, and a
+      // collector that stalls mid-body must not hold up the caller.
+      const drained = drainResponseBody(response);
```
jpmunz, line 209: Given how specific this issue is I think we should wrap the logic in an opt-in config available for users who run into this for their third-party export but then not needing to do the additional read by default, e.g. when sending to Embrace. We could potentially also read the response headers and check for no-store before doing the body read if that saves anything.

Along those lines since we are always doing the fetch here as a POST does that affect whether the issue occurs?

overbalance: I don't recommend reading headers because it isnt possible on an opaque request which would leave out a lot of fetches. It also embeds a chromium-only internal choice that could change over time. The person who pointed this issue out filed a chrome "bug" which google labelled "regression". I would not categorize it that way but it sounds like they'll change the logic in a future version.

Separately, the collectors should probably be updated with `no-store` to be explicit about caching even though as you stated, POST isn't cached. I could see a future where we use GET without an appID header, or a headerless POST to avoid the preflight that happens today.

packages/web-sdk/src/transport/FetchTransport/FetchTransport.ts
overbalance: replaced the link to source code with the chromium issue

packages/web-sdk/src/transport/FetchTransport/FetchTransport.ts
overbalance: @jpmunz I added a section to the body "Performance Impact" with the raw numbers for 1000 requests

jpmunz: looks good, I guess there would be one more case than the ones listed which is the response is read but the drain is not awaited, since there's no await I assume that case would be the same as "1000 sequential round trips, body left unread" ?

jpmunz (approved):

packages/web-sdk/src/transport/FetchTransport/FetchTransport.ts
overbalance: if we read the body there's no need to drain (we should skip it or remove entirely) and the quota would be released at read end. I assume _that_ number would be dependent on the body size

jpmunz: sorry I used two different terms but I was meaning the same operation, the case where we read the response but there is no await attached to that read which is what was implemented here didn't seem to be included in the performance table

PR #1514, opened 2026-09-02
EMBR-13856 feat: add fold measurement duration to DOM state instrumentation

joaquin-diaz:

The DOM state fold measurement recorded its capture moment only as a raw wall-clock timestamp (`dom_state.images_above_fold.timestamp`), which grows unbounded across a long single-page-app visit and doesn't describe "how far into the current view" the fold was captured — the same gap the SDK's other zero-time instrumentations (element timing, user timing, soft navigation, first interaction) already close.

- Add `dom_state.images_above_fold.capture_time`: the fold capture moment expressed in ms since zero time, sent alongside the existing wall-clock `dom_state.images_above_fold.timestamp`.
- Add `millisFromZeroTimeEpoch(epochMillis)` to `PerformanceManager`/`OTelPerformanceManager`: rebases an already-epoch value (e.g. `getNowMillis()`) onto zero time. This is distinct from `millisFromZeroTime(offset)`, which expects a raw time-origin-relative offset — feeding an epoch value into that method double-counts `timeOrigin`.
- Catalog `DOMStateInstrumentation` under the zero-time instrumentations section of `PerformanceManager/README.md`, and document the new method in its API table and worked example.
- Regenerate the `vite-7` integration golden files (the only platform whose goldens include `dom-state` logs) to include the new attribute.

## How has this been tested?

- Unit tests in `DOMStateInstrumentation.test.ts` cover both attributes' values and confirm the log's own timestamp still lands at part end (unaffected by the fold's earlier capture). `OTelPerformanceManager.test.ts` adds baseline, activationStart-rebase, and bfcache-rebase/clamp coverage for the new method.
- `npm run check` (typecheck + eslint) passes across all packages.
- Reran the `vite-7` Playwright suite (chromium/firefox/webkit) against the regenerated goldens — all pass.

## Checklist

- [x] Documentation added
- [x] Tests added

### DIFF
--- a/packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
@@ -10,6 +10,7 @@ import {
   ATTR_DOM_STATE_DOCUMENT_HEIGHT,
   ATTR_DOM_STATE_DOCUMENT_WIDTH,
   ATTR_DOM_STATE_ELEMENT_COUNT,
+  ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_CAPTURE_TIME,
   ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_COUNT,
   ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_TIMESTAMP,
   ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_VIEWPORT_HEIGHT,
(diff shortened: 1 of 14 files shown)


packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
```diff
@@ -146,14 +150,17 @@ export class DOMStateInstrumentation extends EmbraceInstrumentationBase {
...
         [ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_COUNT]:
           this._countImagesAboveFold(measurement),
         [ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_VIEWPORT_HEIGHT]:
           measurement.viewportHeight,
         [ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_VIEWPORT_WIDTH]:
           measurement.viewportWidth,
-        [ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_TIMESTAMP]: this.perf.getNowMillis(),
+        [ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_TIMESTAMP]: captureEpochMillis,
+        [ATTR_DOM_STATE_IMAGES_ABOVE_FOLD_CAPTURE_TIME]:
+          this.perf.millisFromZeroTimeEpoch(captureEpochMillis),
```
jpmunz, line 163: I'd be a little hesitant introducing another performance manager method since it's already a bit tough keeping in mind what the various time conversions are doing, I'd lean towards doing `captureEpochMillis - this.perf.getZeroTime()` unless we see a need for this arising in other spots

joaquin-diaz: This is the first time that we need to do millis from epoch instead of millis from offset (offset being relative to time origin) but I wouldn't be surprised if we need to do this again at some point. I think I rather keep the method so we can always isolate any time-relative math (even if it's really simple like `max(0, captureEpochMillis - this.perf.getZeroTime())` inside the Performance Manager to avoid multiple ways of doing the same or in case there are other related changes like the one we did when we added reset of soft navs

jpmunz (approved):

packages/web-sdk/src/instrumentations/dom-state/DOMStateInstrumentation/DOMStateInstrumentation.ts
jpmunz: yep that's fair, let's keep that then

